PR #1, opened 2016-10-31
Update usage to show the accurate command

kaelig:
When installing Dart Sass with `npm i dart-sass -g`, the executable is `dart-sass`. Unless I'm missing something?


nex3:
You're right, I changed this after writing the docs.

PR #5, opened 2016-10-31
Update pub global command in the documentation.

nshahan:
Should the command be `pub global activate dart-sass` instead of `pub global install dart-sass`?


nex3:
Thanks!

PR #36, opened 2016-11-08
Don't duplicate line/column info in error output.

nex3:
All span information is now presented as a stack frame, and elided if
it's redundant with the existing stack.


nex3:
This still needs updates specs.

PR #53, opened 2016-12-01
Support package: URLs in Dart

### LINKED ISSUE #19: Support package: URLs in Dart
`@import` statements should be able to resolve `package:` URLs when running in Dart mode. The CLI should probably also accept them as well for good measure. We'll want to do this by taking a [`SyncPackageResolver`](https://www.dartdocs.org/documentation/package_resolver/1.0.2/package_resolver/SyncPackageResolver-class.html) from the executable or the `render()` method.

luisvt:
fixes #19


nex3:
I like this change, but I think we should take a [`SyncPackageResolver`](https://www.dartdocs.org/documentation/package_resolver/1.0.2/package_resolver/SyncPackageResolver-class.html) rather than a raw map. A resolver is a lot easier to get for the current isolate, and it's more flexible in case the user wants to use `packages/` directories for whatever reason.

luisvt:
Using `SyncPackageResolverer` inside `visitor` will make the application not compilable to js because this depends on `dart:io`. That's why I created a separated file for handling the `PackageResolver`.

nex3:
We use configuration-specific imports to support VM-only APIs as well as JS. You can add a conditional import that imports a stubbed `SyncPackageResolver` with only the APIs we use when compiling to JS.

luisvt:
ok, I'll try to make the changes. I'll let you know

luisvt:
Changes has been made. Please review them @nex3

lib/sass.dart
```diff
@@ -15,12 +16,12 @@ import 'src/visitor/serialize.dart';
 /// If [color] is `true`, this will use terminal colors in warnings.
 ///
 /// Throws a [SassException] if conversion fails.
-String render(String path, {bool color: false}) {
+Future<String> render(String path, {bool color: false}) async {
```
nex3: This function should take the `SyncPackageResolver`, not `_PerformVisitor`. Rendering needs to be synchronous in order to be compatible with the node-sass API.

Also, there should be a command-line `--packages` flag that allows the user to manually specify a package spec.

lib/src/parse/stylesheet.dart
```diff
@@ -1403,6 +1403,12 @@ abstract class StylesheetParser extends Parser {
         break;
 
       case $0:
+        if (scanner.peekChar(1) == $backslash && scanner.peekChar(2) == $0) {
```
nex3: What's this all about?

lib/src/sync_package_resolver/interface.dart
```diff
@@ -0,0 +1,122 @@
...
+/// cross-package representation of the user's configuration, and as a means of
+/// concretely locating packages and the assets they contain.
+///
+/// Unlike [PackageResolver], all members on this are synchronous, which may
+/// require that more data be loaded up front. This is useful when primarily
+/// dealing with user-created package resolution strategies, whereas
+/// [PackageInfo] is usually preferable when the current Isolate's package
+/// resolution logic may be used.
+///
+/// This class should not be implemented by user code.
```
nex3: You don't need to copy over the documentation here or in the stub implementation.

lib/src/sync_package_resolver/node.dart
```diff
@@ -0,0 +1,125 @@
...
+  /// ensure that all URLs end with a trailing slash.
+  ///
+  /// [urlFor] should generally be used rather than looking up package URLs in
+  /// this map, to ensure that code works with a package root as well as a
+  /// package config.
+  ///
+  /// Returns `null` when using a [packageRoot] for resolution, or when no
+  /// package resolution is being used.
+  Map<String, Uri> get packageConfigMap =>
+      throw new Exception('not implemented for node');
```
nex3: These methods should throw `UnsupportedException`s, which should include the name of the class and method (for example, "SyncPackageResolver.packageConfigMap isn't supported on Node.js.").

lib/src/visitor/perform.dart
```diff
@@ -509,11 +519,17 @@ class _PerformVisitor
     });
   }
 
+  Uri _resolvePackageUri(Uri packageUri) {
+    return packageUri.scheme == 'package' && packageResolver != null
+        ? packageResolver.resolveUri(packageUri) ?? new Uri()
```
nex3: I don't think `new Uri()` is right here. A blank URL doesn't help anyone. We should throw a useful error message saying "Can't resolve "$packageUri"" if it fails to resolve, or if there's no resolver available.

node_sample/import-file/_variables.scss
```diff
@@ -0,0 +1 @@
+$color: blue;
```
nex3: What is this `node_sample` directory for?

test/import_package_uri_test.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:sass/src/exception.dart';
+import 'package:test/test.dart';
+import 'package:sass/sass.dart';
+
+main() {
+  group('import "package:" uri', () {
+    test('found', () async {
+      var result = await render('test/styles/import-package.scss');
```
nex3: Rather than using assets on the filesystem, Dart Sass tests use `scheduled_test`'s `descriptor` library to create assets in the test itself. This helps keep the test self-contained, and makes diffs a lot easier to read. See `test/cli_shared.dart` for examples.

pubspec.yaml
```diff
@@ -29,3 +30,4 @@ dev_dependencies:
   node_preamble: "^1.0.0"
   test: "^0.12.0"
   yaml: "^2.0.0"
+  bootstrap_sass: ^4.0.0-alpah.3+4
```
nex3: I don't like making assertions about other packages' assets in tests. I'd rather create a folder in the test body, then create a `.packages` file that points to it.

tool/grind.dart
```diff
@@ -16,15 +15,15 @@ import 'package:yaml/yaml.dart';
 
 /// The version of Dart Sass.
 final String _version =
-    loadYaml(new File('pubspec.yaml').readAsStringSync())['version'];
+    loadYaml(new File('pubspec.yaml').readAsStringSync())['version'] as String;
```
nex3: It'll be quicker for me to review this PR if you keep the changes focused on `package:` URLs.

lib/src/parse/stylesheet.dart
luisvt: these lines are for parsing code like next:

```scss
@media screen and (min-width:0\0) {
  ...
}
```

specifically are for parsing `0\0` inside a media query.

test/import_package_uri_test.dart
luisvt: could you show me a sample code of this?

lib/src/parse/stylesheet.dart
nex3: Please leave fixes for other issues outside of this PR. One thing at a time makes it easier to review.

lib/sass.dart
luisvt: what do you mean with `package spec`?

nex3: A `.packages` file.

luisvt: if `render` function will take `SyncPackageResolver` as parameter, how should I initialize the passed variable?

nex3:
@luisvt Is this ready for another round of review?

luisvt:
Not yet sorry

luisvt:
now I think is ready, please review it.

lib/sass.dart
```diff
@@ -15,12 +17,24 @@ import 'src/visitor/serialize.dart';
...
   return toCss(cssTree);
 }
+
+/// Loads the Sass file at [path], converts it to CSS, and returns the result in async way.
+/// It also uses the SyncPackageResolver to resolve package uri.
+///
+/// If [color] is `true`, this will use terminal colors in warnings.
+///
+/// Throws a [SassException] if conversion fails.
+Future<String> renderAsync(String path, {bool color: false}) async =>
```
nex3: I don't think I want to add this. I don't think having a default for `packageResolver` is worth an extra top-level function.

lib/src/node.dart
```diff
@@ -34,10 +36,10 @@ void main() {
 /// possible.
 ///
 /// [render]: https://github.com/sass/node-sass#options
-void _render(
-    NodeOptions options, void callback(NodeError error, NodeResult result)) {
+Future _render(NodeOptions options,
+    void callback(NodeError error, NodeResult result)) async {
```
nex3: Why are you changing the types here? Node doesn't understand what a "Future" is, so this will definitely break it.

lib/src/sync_package_resolver/index.dart
```diff
@@ -0,0 +1,3 @@
+export 'interface.dart'
```
nex3: To match the way we stub out IO, this file (`sync_package_resolver/index.dart`) should be `sync_package_resolver.dart` instead.

lib/src/sync_package_resolver/interface.dart
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+abstract class SyncPackageResolver {
+  Map<String, Uri> get packageConfigMap;
```
nex3: Only include the members that Dart Sass actually uses in practice.

lib/src/visitor/perform.dart
```diff
@@ -123,8 +128,13 @@ class _PerformVisitor
   /// invocations, and imports surrounding the current context.
   final _stack = <Frame>[];
 
+  final SyncPackageResolver packageResolver;
```
nex3: Make this private.

lib/src/visitor/perform.dart
```diff
@@ -555,11 +565,29 @@ class _PerformVisitor
     });
   }
 
+  Uri _resolvePackageUri(DynamicImport import) {
```
nex3: Document this.

lib/sass.dart
```diff
@@ -15,12 +17,24 @@ import 'src/visitor/serialize.dart';
 /// If [color] is `true`, this will use terminal colors in warnings.
 ///
 /// Throws a [SassException] if conversion fails.
-String render(String path, {bool color: false}) {
+String render(String path,
+    {bool color: false, SyncPackageResolver packageResolver}) {
```
nex3: Document the `packageResolver` parameter.

lib/src/visitor/perform.dart
```diff
@@ -555,11 +565,29 @@ class _PerformVisitor
...
+    var packageUri = import.url;
+    if (packageUri.scheme == 'package') {
+      if (packageResolver == null)
```
nex3: Nit: include `{}` even for single-statement `if`s (see [the style guide](https://www.dartlang.org/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures)).

pubspec.yaml
```diff
@@ -19,6 +19,7 @@ dependencies:
   string_scanner: ">=0.1.5 <2.0.0"
   stack_trace: ">=0.9.0 <2.0.0"
   tuple: "^1.0.0"
+  package_resolver: ^1.0.2
```
nex3: Always declare the lowest version that has all the APIs you need. In this case, that's 1.0.0.

test/dart_cli_test.dart
```diff
@@ -19,6 +21,34 @@ void main() {
     sass.stdout.expect(matches(new RegExp(r"^\d+\.\d+\.\d+")));
     sass.shouldExit(0);
   });
+
+  group("supports package imports", () {
+    test("is found", () {
+      d.file("test.scss", "a {b: 1 + 2}").create('lib');
```
nex3: Don't create files outside of the sandbox during tests. To properly test this, you'll probably need to add that `--packages` flag.

lib/src/executable.dart
```diff
@@ -47,7 +47,7 @@ main(List<String> args) async {
 
   var color = (options['color'] as bool) ?? hasTerminal;
   try {
-    var css = render(options.rest.first, color: color);
+    var css = await renderAsync(options.rest.first, color: color);
```
nex3: Defaulting to the current isolate's package resolver only makes sense here if Sass is being used as a library or a direct dependency of a package. If it's being run as a global or standalone executable—which is better supported anyway—that logic won't work.

Probably the best thing to do here is to follow Dart's behavior for auto-detecting package resolution from the Sass file's location, and allow users to override it with a `--packages` flag. For now, just remove the default behavior and implement the flag; we can do the auto-detection in a separate CL.

luisvt: ok

lib/src/node.dart
luisvt: Sorry, I forgot to revert it.

lib/src/visitor/perform.dart
```diff
@@ -555,11 +566,35 @@ class _PerformVisitor
...
+      if (_packageResolver == null) {
+        throw _exception(
+            'Can\'t resolve: "$packageUri", packageResolver is not supported by node vm.'
+            ' If you are using dart-vm please provide a `SyncPackageResolver` to the `render` function',
+            import.span);
+      }
+
+      var resolvedPackageUri = _packageResolver.resolveUri(packageUri);
+      if (resolvedPackageUri == null ||
+          !dirExists(p.dirname(p.fromUri(resolvedPackageUri)))) {
```
nex3: I'm not sure I understand the point of this check. If the package URI is resolvable, we'll already get an error when we try to load the file, right?

test/dart_cli_test.dart
```diff
@@ -19,6 +19,21 @@ void main() {
...
+//
+//      var sass = _runSass(["test.scss", "test.css"]);
+//      sass.stdout.expect(inOrder([
+//        "a {",
+//        "  b: 3;",
+//        "}",
+//      ]));
+//      sass.shouldExit(0);
+//    });
+//  });
```
nex3: Remove this commented-out code.

test/dart_script_test.dart
```diff
@@ -0,0 +1,43 @@
...
+      render(sandbox + "/test.scss", packageResolver: packageResolver);
+      fail("exception not thrown");
+    } catch (e) {
+      expect(
+          e.toString(),
+          "Error on line 1, column 9 of $sandbox/test.scss: Can't resolve: \"package:fake_package/test_aux\"\n"
+          "@import \"package:fake_package/test_aux\";\n"
+          "        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n"
+          "  $sandbox/test.scss 1:9  root stylesheet");
+    }
```
nex3: It's generally better to use the `throwsA()` matcher than to manually write a try/catch block. Then you can verify the exception's message with the `predicate()` matcher.

lib/src/visitor/perform.dart
luisvt: This check is for when the package URI is not resolvable. What should I do instead this check?

If I remove this check `resolvedPackageUri` will be null, which in turn will throw:
```
NoSuchMethodError: The getter 'scheme' was called on null.
```

Second option would be return an empty uri, which was what I have at the beginning and you told me to better add a check and throw a better error. Please see the comment you added on https://github.com/sass/dart-sass/pull/53#discussion-diff-92956954R524

Third option would be return the original `packageUri`, but that lets to previous problem on having an error message not showing good information.

Finally the line 585 in https://github.com/sass/dart-sass/pull/53#discussion-diff-96078273R585 was added because previous version of sdk returned an URI even when there was no existing file.

nex3:
Thanks for the patch!

PR #104, opened 2017-01-17
Implement homebrew (macOS) and scoop (Windows) deployments

### LINKED ISSUE #97: Distribute on Homebrew
We should distribute a snapshot via Homebrew to make it easier for Mac OS users to use Dart Sass. Ideally we'd depend on [the Dart VM's Homebrew package](https://www.dartlang.org/install/mac), which probably means that we'll have to set up our own tap.

sestegra:
FIx #97

### DIFF
--- a/package/dart-sass.sh
+++ b/package/dart-sass.sh
@@ -8,12 +8,15 @@
 # package`.
 
 function follow_links() {
-  file="$1"
+  local file="$1"
+
+  cd $(dirname "$file")
   while [ -h "$file" ]; do
     # On Mac OS, readlink -f doesn't work.
     file="$(readlink "$file")"
+    cd $(dirname "$file")
   done
-  echo "$file"
+  echo "$(pwd -P)/$(basename "$file")"
 }
 
 # Unlike $0, $BASH_SOURCE points to the absolute path of this file.
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -21,9 +21,6 @@ final String _version =
 /// The version of the current Dart executable.
 final String _dartVersion = Platform.version.split(" ").first;
 
-/// The root of the Dart SDK.
-final _sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
-
 main(List<String> args) => grind(args);
 
 @DefaultTask('Run the Dart formatter.')
@@ -212,17 +209,18 @@ Future _buildPackage(http.Client client, String os, String architecture) async {
         "${response.reasonPhrase}.";
   }
 
-  var dartExecutable = new ZipDecoder()
-      .decodeBytes(response.bodyBytes)
-      .firstWhere((file) => os == 'windows'
-          ? file.name.endsWith("/bin/dart.exe")
-          : file.name.endsWith("/bin/dart"));
+  var dartSdk = new ZipDecoder().decodeBytes(response.bodyBytes);
+  var dartExecutable = dartSdk.firstWhere((file) => os == 'windows'
+      ? file.name.endsWith("/bin/dart.exe")
+      : file.name.endsWith("/bin/dart"));
+  var dartLicense = dartSdk.firstWhere((file) => file.name.endsWith("LICENSE"));
   var executable = dartExecutable.content;
+  var license = dartLicense.content;
   var archive = new Archive()
     ..addFile(_fileFromBytes(
         "dart-sass/src/dart${os == 'windows' ? '.exe' : ''}", executable,
         executable: true))
-    ..addFile(_file("dart-sass/src/DART_LICENSE", p.join(_sdkDir, 'LICENSE')))
+    ..addFile(_fileFromBytes("dart-sass/src/DART_LICENSE", license))
     ..addFile(
         _file("dart-sass/src/sass.dart.snapshot", "build/sass.dart.snapshot"))
     ..addFile(_file("dart-sass/src/SASS_LICENSE", "LICENSE"))
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -314,10 +314,43 @@ class DartSass < Formula
     bin.install "dart-sass"
     bin.install Dir["src"]
   end
-end
-  """;
+end""";
 
   var output = "build/dart-sass.rb";
   log("Creating $output...");
   new File(output).writeAsStringSync(formulae);
+}
+
+@Task('Build Scoop manifest.')
+//@Depends(package)
+scoop_manifest() {
+  var hash_x64 = _hash('build/dart-sass-$_version-windows-x64.zip', 'sha256');
+  var hash_ia32 = _hash('build/dart-sass-$_version-windows-ia32.zip', 'sha256');
+  var manifest = """
+{
+  "homepage": "https://github.com/sass/dart-sass",
+  "version": "$_version",
+  "license": "https://raw.githubusercontent.com/sass/dart-sass/master/LICENSE",
+  "architecture": {
+    "64bit": {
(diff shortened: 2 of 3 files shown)


package/dart-sass.sh
```diff
@@ -8,12 +8,15 @@
...
+  local file="$1"
+
+  cd $(dirname "$file")
   while [ -h "$file" ]; do
     # On Mac OS, readlink -f doesn't work.
     file="$(readlink "$file")"
+    cd $(dirname "$file")
   done
-  echo "$file"
+  echo "$(pwd -P)/$(basename "$file")"
```
nex3: Can you explain in more detail what this change is doing? This function [is directly copied from the Dart SDK](https://github.com/dart-lang/sdk/blob/master/sdk/bin/dart#L6-L13), so if it's broken in some situation it seems like it should be fixed there too.

tool/grind.dart
```diff
@@ -271,3 +272,52 @@ ArchiveFile _fileFromString(String path, String contents,
...
+      break;
+    case 'sha1':
+      hasher = sha1;
+      break;
+    case 'sha256':
+      hasher = sha256;
+      break;
+  }
+  return hasher.convert(new File(file).readAsBytesSync()).toString();
+}
```
nex3: You're only ever using `sha256` here, so it's probably cleaner to just make this a `_sha256()` function.

sestegra: Done

sestegra:
In order to deploy the snapshot only, I have 2 ways:
- Download the standalone package from release page and extract only the snapshot. 
- Download the snapshot, but it's not available on release page

What do you suggest ?

Concerning the second part of your question, indeed, the easiest way is to create your own tap/bucket to control the deployment.

package/dart-sass.sh
sestegra: This script isn't deployed on Mac. /usr/local/bin/dart is directly a symlink to the binary.

Concerning this fix, Homebrew uses double relative links redirection. Without the fix, the issue is following:
```
$ dart-sass 
/usr/local/bin/dart-sass: line 21: <current pwd>/../Cellar/dart-sass/1.0.0-alpha.8/bin/src/dart: No such file or directory
/usr/local/bin/dart-sass: line 21: exec: <current pwd>/../Cellar/dart-sass/1.0.0-alpha.8/bin/src/dart: cannot execute: No such file or directory
```

nex3:
Can you include the snapshot in the package itself rather than downloading it at install-time? That's what we do for the Chocolatey package.

sestegra:
Homebrew formula (.rb file only) and scoop manifest (.json file only) aren't packages.
I can't put the snapshot with them.

package/dart-sass.sh
nex3: > This script isn't deployed on Mac. /usr/local/bin/dart is directly a symlink to the binary.

I don't understand. This script doesn't wrap the `dart` executable, it wraps `dart-sass`. And if this isn't deployed, then how does the user invoke `dart-sass`? There's no single executable to invoke, you have to pass the snapshot to the `dart` executable.

> Concerning this fix, Homebrew uses double relative links redirection. Without the fix, the issue is following:
> ```
> $ dart-sass 
> /usr/local/bin/dart-sass: line 21: <current pwd>/../Cellar/dart-sass/1.0.0-alpha.8/bin/src/dart: No such file or directory
> /usr/local/bin/dart-sass: line 21: exec: <current pwd>/../Cellar/dart-sass/1.0.0-alpha.8/bin/src/dart: cannot execute: No such file or directory
> ```

Why doesn't repeatedly `readlink`ing work? Why does `cd`ing fix it? Doesn't it change the user's working directory in a way they won't want? Why is `local file` required?

I've filed dart-lang/sdk#28747 to track the issue in the Dart SDK.

tool/grind.dart
```diff
@@ -273,3 +272,72 @@ ArchiveFile _fileFromString(String path, String contents,
...
+    url 'https://github.com/sass/dart-sass/releases/download/$_version/dart-sass-$_version-macos-x64.tar.gz'
+    sha256 '$hash_x64'
+  else
+    url 'https://github.com/sass/dart-sass/releases/download/$_version/dart-sass-$_version-macos-ia32.tar.gz'
+    sha256 '$hash_ia32'
+  end
+
+  def install
+    bin.install "dart-sass"
+    bin.install Dir["src"]
```
nex3: This looks wrong to me. Nothing in the `src` directory should be directly executed by the user, nor should it be on their path. It's just an implementation detail for the `dart-sass` executable.

sestegra: Homebrew installs the package into ```/usr/local/Cellar/dart-sass/<version>``` directory.

After installation, a symbolic link ```/usr/local/bin/dart-sass``` that's pointing on Cellar place is created.

Only ```/usr/local/bin/dart-sass``` is on the path of the end user.

```
/usr/local/bin/dart-sass -> ../Cellar/dart-sass/1.0.0-alpha.8/bin/dart-sass
```

```
$ ls -1R /usr/local/Cellar/dart-sass/
1.0.0-alpha.8

/usr/local/Cellar/dart-sass//1.0.0-alpha.8:
INSTALL_RECEIPT.json
bin

/usr/local/Cellar/dart-sass//1.0.0-alpha.8/bin:
dart-sass
src

/usr/local/Cellar/dart-sass//1.0.0-alpha.8/bin/src:
DART_LICENSE
SASS_LICENSE
dart
sass.dart.snapshot```

File: package/dart-sass.sh (hunk 1 above)
Comment (sestegra) [inline package/dart-sass.sh]: For the 1st question, sorry for the confusion, this script isn't deployed for ```dart``` package with Homebrew. Obviously, it's mandatory for ```dart-sass``` package.
```
$ ls /usr/local/bin/dart
/usr/local/bin/dart -> ../Cellar/dart/1.21.1/bin/dart
$ file /usr/local/Cellar/dart/1.21.1/bin/dart
/usr/local/Cellar/dart/1.21.1/bin/dart: Mach-O 64-bit executable x86_64
```

> Why doesn't repeatedly readlinking work? 
> Why does cding fix it?

I don't really know why ```readlink``` tool doesn't do the job.
Changing directory is working because Homebrew symlinks are relative (not absolute).

> Doesn't it change the user's working directory in a way they won't want?

The script forks shell environment when launching (because of the "shebang").
Only exported values (with ```export``` command) will be "visible" at the end of the script.
In other words, the "cd" commands in the script won't affect the user's environment: the current working directory is unchanged.

> Why is local file required?

It's a best practice to put ```local``` to reduce the scope of the variable.
Otherwise, the variable is global by default on shell script.
https://google.github.io/styleguide/shell.xml#Use_Local_Variables

**Comment by sestegra** [comment]
### Homebrew installation
https://brew.sh/ explains how to install Homebrew it self.
Then, you have to create a dedicated repository (tap).
For instance, see https://github.com/dart-lang/homebrew-dart
Assume its name is ```dart-sass/homebrew-dart-sass```.

To install dart-sass, do following:
```
$ brew tap dart-sass/dart-sass
$ brew install dart-sass
```

### Scoop installation
http://scoop.sh/ explains how to install Scoop it self.
As I put dart-sass manifest in official scoop's bucket, to install dart-sass, do following:
```
$ scoop install dart-sass
```

sestegra:
Do you need more information from my side before merging this pull request ?

nex3:
Not at the moment, I'm just waiting to find time to iterate on it.

nex3:
I've done some poking around with Homebrew, and I think I want to go a different direction. Rather than generating the entire source of the formula here, I want to keep the canonical source in our Homebrew tap and just update it from here. I do appreciate your contribution, though!

PR #118, opened 2017-02-08
Run Travis on OS X.

nex3:


nschonni:
Is there a reason to split out the `spec` and `test` runs?

nex3:
I generally like having different builds for different types of tests. Also, we'd need a way to run only tests anyway, since we don't want to run specs for each version of Node.

nschonni:
Fair enough, I was just thinking about the slowdown from waiting for the separate builds to start might outweigh that. Especially since the OSX infrastructure seems to get congested regularly

nschonni:
You could also noop the specific run parts on the Node stuff like we do for the coverage in libsass

nex3:
We're doing that somewhat via the build matrix. It may be worth just running everything in a single job on OS X though for that reason.

nschonni:
Maybe move that to a `ci-travis` script? I think my thinking was the same as yours done from a differnt angle, I just remember problems with trying to get the matrix to behave in the past.

nex3:
I'm going to close this out for now. The cycle time of OS X builds is too high to realistically make it part of our continuous integration system. We'll have to figure something else out if we want to build releases on OS X, unfortunately.

PR #126, opened 2017-02-27
Rename .analysis_options to analysis_options.yaml

HoBi:
> Breaking change: The conventional name for the analysis options file used to be .analysis_options (note the leading dot and missing .yaml suffix). We expect support for the .analysis_options name to go away in a future release, so we recommend that you rename your .analysis_options files to analysis_options.yaml.

https://www.dartlang.org/guides/language/analysis-options#the-analysis-options-file


nex3:
Sorry for the long wait time for a review! I've been on medical leave for the past couple months.

PR #133, opened 2017-05-12
Two internal/API features for Lists

srawlins:
Two internal/API features for Lists:

* Add separator character to ListSeparator, similar to BinaryOperator.
* Add trailing comma to ListExpression.

### DIFF
--- a/lib/src/value/list.dart
+++ b/lib/src/value/list.dart
@@ -57,20 +57,22 @@ class SassList extends Value {
 /// An enum of list separator types.
 class ListSeparator {
   /// A space-separated list.
-  static const space = const ListSeparator._("space");
+  static const space = const ListSeparator._("space", " ");
 
   /// A comma-separated list.
-  static const comma = const ListSeparator._("comma");
+  static const comma = const ListSeparator._("comma", ",");
 
   /// A separator that hasn't yet been determined.
   ///
   /// Singleton lists and empty lists don't have separators defiend. This means
   /// that list functions will prefer other lists' separators if possible.
-  static const undecided = const ListSeparator._("undecided");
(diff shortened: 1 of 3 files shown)


nex3:
Sorry for the long wait time for a review! I've been on medical leave for the past couple months.

I don't understand the trailing comma change here. Does it just affect the `inspect()` output? I don't think tracking whether the user wrote a trailing comma is consistent with Ruby Sass.

srawlins:
Ah sorry, the trailing comma change is just so that ListExpression exposes the notion externally, not for use in compiling to CSS.

I can table that part of this change for now. It's part of a bigger effort I'd like to engage in, where I'd like the AST to be a more compete representation of the source. We can chat offline.

nex3:
> I can table that part of this change for now. It's part of a bigger effort I'd like to engage in, where I'd like the AST to be a more compete representation of the source. We can chat offline.

Ah, I see. Yeah, let's VC about that.

srawlins:
OK reduced changes to just adding the separator character.

lib/src/value/list.dart
```diff
@@ -57,20 +57,22 @@ class SassList extends Value {
...
   ///
   /// Singleton lists and empty lists don't have separators defiend. This means
   /// that list functions will prefer other lists' separators if possible.
-  static const undecided = const ListSeparator._("undecided");
+  static const undecided = const ListSeparator._("undecided", null);
 
   final String _name;
 
-  const ListSeparator._(this._name);
+  final String separator;
```
nex3: Document this. In particular, mention that it's `null` sometimes since that's not necessarily obvious from the API.

srawlins: Done.

nex3:
Thanks!

Issue #8, opened 2016-11-01
JS API: Support formatting options


nex3:
We need to support the node-sass API's options for formatting its output:

* [x] [`indentType`](https://github.com/sass/node-sass#indenttype--v300) controls whether tabs or spaces are emitted.
* [x] [`indentWidth`](https://github.com/sass/node-sass#indentwidth--v300) controls how many tabs or spaces are emitted per indentation.
* [x] [`linefeed`](https://github.com/sass/node-sass#linefeed--v300), which controls what line break sequence to use in the resulting CSS.
* [ ] [`outputStyle`](https://github.com/sass/node-sass#outputstyle), which controls the style of CSS emitted. Note that this will only support the value `expanded` for the time being, and unlike node-sass, it will default to `expanded`.

xzyfer:
I expect Node Sass 4 will also default to expanded, and SassC will eventually follow.

srawlins:
@nex3, when you say:

> Note that this will only support the value expanded for the time being, and unlike node-sass, it will default to expanded.

Do you mean that in order to close this issue (and for 1.0.0), all you want is that `--outputStyle` is available as a flag, but it only supports the value `expanded`?

nex3:
That's right.

hanakin:
not sure why this is closed and I might be missing something but indentType and indentWidth are not currently supported or at least not documented on how to use anywhere?

nex3:
This issue was tracking support for these options in the legacy JS API, which are documented [here](https://sass-lang.com/documentation/js-api/interfaces/LegacySharedOptions). We did not choose to implement these in the new JS API because we want Sass to be opinionated about its formatting rather than highly customizable.

hanakin:
opinionated is one thing but I would suggest allowing for indentType should be included in CLI as tabs are widely used

nex3:
The answer is no.

Issue #7, opened 2016-11-01
JS API: Support input options


nex3:
We need to support the node-sass API's input options:

* [x] [`file`](https://github.com/sass/node-sass#file), which supplies a source file path.
* [x] [`data`](https://github.com/sass/node-sass#data), which supplies a source file as a string.
* [ ] [`includePaths`](https://github.com/sass/node-sass#includepaths), which indicates paths to search for imported files.
* [x] [`indentedSyntax`](https://github.com/sass/node-sass#indentedsyntax), which indicates whether the input is indented or not.

PR #136, opened 2017-05-15
Add flags: --indent-type and --indent-width

srawlins:
This is for https://github.com/sass/dart-sass/issues/8, 2 of the 4 tasks. There is a unit test, but here's some example usage as well:

```none
$ cat x.scss 
p {
  $color: red;
  a {
    color: $color;
  }
}
$ osdart bin/sass.dart x.scss
p a {
  color: red;
}
$ osdart bin/sass.dart x.scss --indent-width 3
p a {
   color: red;
}
$ osdart bin/sass.dart x.scss --indent-type tab
p a {
                color: red;
}
$ osdart bin/sass.dart x.scss --indent-type tab --indent-width 1
p a {
        color: red;
}
$ node build/npm/sass.js x.scss 
p a {
  color: red;
}
$ node build/npm/sass.js x.scss --indent-width 4
p a {
    color: red;
}
$ node build/npm/sass.js x.scss --indent-width 1 --indent-type tab
p a {
        color: red;
}
```


srawlins:
I see. I like it. Should I just:

* kill my changes in `lib/src/executable.dart`?
* move the valid argument checks into `render()`?
* and kill my test?

nex3:
I think this probably shouldn't be exposed via the Dart `render()` method either. You'd just add them to the node `render()`, defined in `lib/src/node.dart`.

srawlins:
OK I've removed `--indent-type` and `--indent-width` from the CLI.

Unfortunately, to remove `indentType` and `indentWidth` from `render()`, I'd have to fork `render()`, as Node gets to `toCss` via that `render` method.

nex3:
Hmm, that is tricky. Maybe add `lib/src/render` that has the union of the two APIs?

srawlins:
OK, I've got a nice impl going, but let me know:

* `lib/sass.dart` no longer implements `render()`, it just exports `lib/src/render.dart`.
* `lib/src/render.dart` re-exports either `render/vm.dart` or `render/node.dart`, with interface `render/interface.dart`.
* `lib/src/render/vm.dart` disallows indentWidth and indentType.
* To avoid duplication (and even further duplication once #137 is merged), both implementations import `shared.dart`, which chooses how to parse (which will make [`indentedSyntax`](https://github.com/sass/dart-sass/issues/7) very easy to add), builds a SassTree, evaluates to a CssTree, then serializes to CSS.

There are two things I don't like about this solution: the name of the "private" method `renderSourceToCss()`, called by `render()` and soon to be called by `renderSource()`. I'm open to suggsetions. Also, it's not actually private. The only way to make it private would be to make it a `part` of `render/node.dart` AND `render/vm.dart`. I see no parts in this repo, so I'm hesitant to add the first. Also, does the spec even allow that?

lib/src/render.dart
```diff
@@ -0,0 +1,7 @@
+// Copyright 2017 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+export 'render/interface.dart'
+    if (dart.library.io) 'render/vm.dart'
+    if (node) 'render/node.dart';
```
nex3: This isn't necessary. The API exposed by `lib/sass.dart` is only relevant to Dart. The API that's seen by JS is produced by the exports assigned in `lib/src/node.dart`. That may change in the future as JS interop gets more fully-featured, but for now it doesn't make sense to export any Node APIs through `lib/sass.dart`.

srawlins: Done.

srawlins:
Yes a much simpler solution. Thanks! Done.

lib/src/render.dart
```diff
@@ -0,0 +1,50 @@
...
+        "either 'space' or 'tab'.");
+    exitCode = 1;
+    return null;
+  }
+  if (indentWidth < 0 || indentWidth > 10) {
+    stderr.writeln("Error: $indentWidth is an invalid indent width; must be "
+        "between 0 and 10, inclusive.");
+    exitCode = 1;
+    return null;
+  }
```
nex3: These errors should just be thrown, they shouldn't force the process to exit.

lib/src/node/options.dart
```diff
@@ -7,4 +7,6 @@ import 'package:js/js.dart';
 @JS()
 class NodeOptions {
   external String get file;
+  external String get indentType;
+  external int get indentWidth;
```
nex3: It looks like node-sass supports String indent widths as well, and just parses them into ints. That seems ridiculous to me, but we should probably match their behavior.

lib/src/render.dart
```diff
@@ -0,0 +1,50 @@
...
+/// Otherwise, they aren't supported. It takes a [SyncPackageResolver][] from
+/// the `package_resolver` package.
+///
+/// [SyncPackageResolver]: https://www.dartdocs.org/documentation/package_resolver/latest/package_resolver/SyncPackageResolver-class.html
+///
+/// Finally throws a [SassException] if conversion fails.
+String render(String path,
+    {bool color: false,
+    SyncPackageResolver packageResolver,
+    String indentType: 'space',
```
nex3: For pure-Dart representations, let's make this `bool useSpaces` and keep the translation from JS-isms isolated in `lib/src/node.dart`.

lib/src/visitor/serialize.dart
```diff
@@ -802,8 +822,9 @@ class _SerializeCssVisitor
   /// Writes indentation based on [_indentation].
   void _writeIndentation() {
     for (var i = 0; i < _indentation; i++) {
-      _buffer.writeCharCode($space);
-      _buffer.writeCharCode($space);
+      for (var j = 0; j < _indentWidth; j++) {
+        _buffer.writeCharCode(_indentCharacter);
```
nex3: It looks like this will use two tabs for indentation if `indentType: 'tab'` is passed with the default width. Is that what node-sass does?

srawlins: Yes, I believe it is. https://github.com/sass/node-sass/blob/c048687bf5df536b298154a72939438ea7be0b58/lib/index.js#L110..L137

lib/src/node/options.dart
srawlins: Done.

lib/src/render.dart
srawlins: Done.

lib/src/render.dart
srawlins: Done.

srawlins:
OK I think this is ready again.

nex3:
Looks good, thanks!

PR #137, opened 2017-05-15
Read Sass from stdin

srawlins:
Addresses one task of #7 .

This includes a breaking change to `render()`: the old required argument is now an optional named argument, `path`.

In action:

```none
$ osdart bin/sass.dart 
$color: red;
a { color: $color; }^D
a {
  color: red;
}
$ node build/npm/sass.js 
$color: red;
a { color: $color; }^D
a {
  color: red;
}
```


lib/sass.dart
```diff
@@ -22,13 +22,27 @@ import 'src/visitor/serialize.dart';
...
-  var contents = readFile(path);
-  var url = p.toUri(path);
-  var sassTree = p.extension(path) == '.sass'
-      ? new Stylesheet.parseSass(contents, url: url, color: color)
-      : new Stylesheet.parseScss(contents, url: url, color: color);
+String render(
+    {String path,
+    String data,
+    bool color: false,
+    SyncPackageResolver packageResolver}) {
```
nex3: I don't think I want to change the signature here. The existing `render()` signature is clean and useful. I'd much rather add a separate function that takes `data`.

lib/src/io/node.dart
```diff
@@ -48,6 +58,16 @@ class Stderr {
   void flush() {}
 }
 
+class Stdin {
```
nex3: What's the benefit of defining this class, instead of just using `_Stdin` in `readStdin()`?

lib/src/io/node.dart
```diff
@@ -80,6 +100,19 @@ _readFile(String path, [String encoding]) {
   }
 }
 
+String readStdin() async {
```
nex3: This needs to be defined in `io/interface.dart` as well.

Also, it should be `Future<String>`.

lib/src/io/node.dart
```diff
@@ -80,6 +100,19 @@ _readFile(String path, [String encoding]) {
...
+  var completer = new Completer<Null>();
+  var contents = new StringBuffer();
+  _stdin.on('data', allowInterop((buf) {
+    contents.write(UTF8.decode(buf));
+  }));
+  _stdin.on('end', allowInterop(() {
+    completer.complete();
+  }));
+  await completer.future;
```
nex3: Rather than doing this dance, just make it a `Completer<String>` and return `completer.future`.

lib/src/io/node.dart
```diff
@@ -80,6 +100,19 @@ _readFile(String path, [String encoding]) {
   }
 }
 
+String readStdin() async {
+  var completer = new Completer<Null>();
+  var contents = new StringBuffer();
+  _stdin.on('data', allowInterop((buf) {
+    contents.write(UTF8.decode(buf));
+  }));
```
nex3: We should handle any errors that node emits here.

lib/src/io/vm.dart
```diff
@@ -39,6 +39,13 @@ String readFile(String path) {
   }
 }
 
+String readStdin() {
+  var inputBuffer = new StringBuffer();
+  var line;
+  while ((line = io.stdin.readLineSync()) != null) inputBuffer.writeln(line);
+  return inputBuffer.toString();
```
nex3: It may be faster to use [`StringAccumulatorSink`](https://www.dartdocs.org/documentation/convert/2.0.1/convert/StringAccumulatorSink-class.html) here and consume the data asynchronously.

lib/src/io/node.dart
srawlins: Sorry, just copy-pasting. Removed.

lib/src/io/node.dart
srawlins: Done.

lib/src/io/node.dart
srawlins: Done.

lib/src/io/node.dart
srawlins: Done. (\*facepalm* I swear I know how to program 😝  )

lib/src/io/vm.dart
srawlins: I love it. Done.

srawlins:
Changed to read from stdin on `-`; lemme know what you think!

lib/sass.dart
```diff
@@ -33,3 +33,23 @@ String render(String path,
       evaluate(sassTree, color: color, packageResolver: packageResolver);
   return toCss(cssTree);
 }
+
+/// Loads [source] as Sass, converts it to CSS, and returns the result.
```
nex3: "as Sass" makes it a little ambiguous whether this is parsed as the indented syntax or as SCSS by default. I think SCSS is the right answer, but it should be clearer in the docs. It would also be good to add an option to configure that.

lib/sass.dart
```diff
@@ -33,3 +33,23 @@ String render(String path,
...
+/// If [packageResolver] is provided, it's used to resolve `package:` imports.
+/// Otherwise, they aren't supported. It takes a [SyncPackageResolver][] from
+/// the `package_resolver` package.
+///
+/// [SyncPackageResolver]: https://www.dartdocs.org/documentation/package_resolver/latest/package_resolver/SyncPackageResolver-class.html
+///
+/// Finally throws a [SassException] if conversion fails.
+String renderSource(String source,
+    {bool color: false, SyncPackageResolver packageResolver}) {
+  var url = new UriData.fromString(source);
```
nex3: I suspect using a `data:` URL here will produce some pretty gnarly errors. We should allow the user to pass in their own URL if they want, but if they don't I think we should just forward the `null` value to `parseScss()`. If it chokes, that's a bug.

lib/src/io/vm.dart
```diff
@@ -39,6 +41,14 @@ String readFile(String path) {
   }
 }
 
+Future<String> readStdin() async {
+  var completer = new Completer<String>();
+  var inputBuffer = new StringAccumulatorSink();
+  io.stdin.transform(new Utf8Decoder()).listen(inputBuffer.add,
+      onDone: () => completer.complete(inputBuffer.string));
+  return completer.future;
+}
```
nex3: I think `Encoding.decodeStream()` does this.

And we should probably use `SystemEncoding` here as well.

lib/src/io/node.dart
```diff
@@ -80,6 +90,22 @@ _readFile(String path, [String encoding]) {
   }
 }
 
+Future<String> readStdin() async {
+  var completer = new Completer<String>();
+  var contents = new StringBuffer();
+  _stdin.on('data', allowInterop((buf) {
+    contents.write(UTF8.decode(buf));
```
nex3: This runs the risk of splittting a character across chunk boundaries. You can use `Converter.startChunkedConversion()` to track state across chunks, probably with `StringConversionSink.withCallback()` to get the end result.

Also this should probably be `SystemEncoding`.

lib/src/io/node.dart
```diff
@@ -80,6 +90,22 @@ _readFile(String path, [String encoding]) {
...
+  }));
+  _stdin.on('end', allowInterop(() {
+    completer.complete(contents.toString());
+  }));
+  _stdin.on('error', allowInterop((e) {
+    stderr.writeln('Failed to read from stdin');
+    stderr.writeln(e);
```
nex3: I think this should be forwarded to `completer`.

test/cli_shared.dart
```diff
@@ -35,6 +35,18 @@ void sharedTests(ScheduledProcess runSass(List arguments)) {
...
+    var sass = runSass(["-"]);
+    sass.writeLine("a {b: 1 + 2}");
+    sass.closeStdin();
+    sass.stdout.expect(inOrder([
+      "a {",
+      "  b: 3;",
+      "}",
+    ]));
+    sass.shouldExit(0);
+  });
```
nex3: Test what an error message looks like from stdin.

lib/sass.dart
srawlins: Done. I think the option will be addressed wth the indentedSyntax issue, https://github.com/sass/dart-sass/issues/7.

lib/sass.dart
srawlins: Done.

lib/src/io/vm.dart
srawlins: Done.

lib/src/io/node.dart
srawlins: Done. Unfortunately I found no [notion of "system encoding" in Node](https://www.google.com/search?q=node+system+encoding). The closest thing to a default is that Buffer [defaults to 'utf8' all over the place](https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings).

lib/src/io/node.dart
srawlins: Done.

PR #153, opened 2017-06-05
Support --linefeed in Dart and Node

srawlins:
Addresses part of #8 

If someone passes in something other than `cr`, `crlf`, `lf`, `lfcr`, it defaults of `lf`, [like node-sass](https://github.com/sass/node-sass/blob/c048687bf5df536b298154a72939438ea7be0b58/lib/index.js#L146..L155).

I'm happy to move `parseLineFeed` and `enum LineFeed` to... other files, open to suggestions.

I'm also happy to add tests, but `emitsInOrder` or whatever Stream, is not letting me see the line endings...

But here's a cmdline test, with the newlines marked with `cat -e`:

```
$ dart bin/sass.dart --linefeed cr bb.scss | cat -e
a {^M  b: 7;^M}$
$ dart bin/sass.dart --linefeed crlf bb.scss | cat -e
a {^M$
  b: 7;^M$
}$
$ dart bin/sass.dart --linefeed lf bb.scss | cat -e
a {$
  b: 7;$
}$
```

### DIFF
--- a/lib/src/node.dart
+++ b/lib/src/node.dart
@@ -11,6 +11,7 @@ import 'node/exports.dart';
 import 'node/options.dart';
 import 'node/result.dart';
 import 'render.dart';
+import 'visitor/serialize.dart';
 
 /// The entrypoint for Node.js.
 ///
--- a/lib/src/node.dart
+++ b/lib/src/node.dart
@@ -41,11 +41,11 @@ void _render(
     var indentWidth = indentWidthValue is int
         ? indentWidthValue
         : int.parse(indentWidthValue.toString());
-    var linefeed = parseLineFeed(options.linefeed);
+    var lineFeed = _parseLineFeed(options.linefeed);
     var result = newNodeResult(render(options.file,
         useSpaces: options.indentType == 'space',
         indentWidth: indentWidth,
-        linefeed: linefeed));
+        lineFeed: lineFeed));
     callback(null, result);
   } on SassException catch (error) {
     // TODO: populate the error more thoroughly if possible.
@@ -65,13 +65,31 @@ NodeResult _renderSync(NodeOptions options) {
     var indentWidth = indentWidthValue is int
         ? indentWidthValue
         : int.parse(indentWidthValue.toString());
-    var linefeed = parseLineFeed(options.linefeed);
+    var lineFeed = _parseLineFeed(options.linefeed);
     return newNodeResult(render(options.file,
         useSpaces: options.indentType == 'space',
         indentWidth: indentWidth,
-        linefeed: linefeed));
+        lineFeed: lineFeed));
   } on SassException catch (error) {
     // TODO: populate the error more thoroughly if possible.
     throw new NodeError(message: error.message);
+  }
+}
+
+/// Parses a String representation of a [LineFeed] into a value of the enum-like
+/// class.
+LineFeed _parseLineFeed(String str) {
+  switch (str) {
+    case 'cr':
+      return LineFeed.cr;
+      break;
+    case 'crlf':
+      return LineFeed.crlf;
+      break;
+    case 'lfcr':
+      return LineFeed.lfcr;
+      break;
+    default:
+      return LineFeed.lf;
   }
 }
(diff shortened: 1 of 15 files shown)


nex3:
Like with `indentType` and `indentWidth`, we're only supporting the `linefeed` parameter for compatibility with node-sass, so I'd prefer not to expose it as part of the Dart API or the CLI.

srawlins:
Done! Removed `linefeed` from Dart.

nex3:
Please also remove it from the command-line interface.

srawlins:
Done.

lib/src/node.dart
```diff
@@ -41,8 +41,11 @@ void _render(
     var indentWidth = indentWidthValue is int
         ? indentWidthValue
         : int.parse(indentWidthValue.toString());
+    var linefeed = parseLineFeed(options.linefeed);
```
nex3: We should consistently capitalize `lineFeed`, at least in places where we it doesn't need to be identical to the JS API. I think capitalizing "feed" is more correct.

lib/src/render.dart
```diff
@@ -27,5 +27,23 @@ String render(String path,
       : new Stylesheet.parseScss(contents, url: url, color: color);
   var cssTree =
       evaluate(sassTree, color: color, packageResolver: packageResolver);
-  return toCss(cssTree, useSpaces: useSpaces, indentWidth: indentWidth);
+  return toCss(cssTree,
+      useSpaces: useSpaces, indentWidth: indentWidth, linefeed: linefeed);
+}
+
+LineFeed parseLineFeed(String str) {
```
nex3: It seems like this should be a private method in node.dart.

Also, please document it.

lib/src/render.dart
```diff
@@ -27,5 +27,23 @@ String render(String path,
...
+    case 'cr':
+      return LineFeed.CR;
+      break;
+    case 'crlf':
+      return LineFeed.CRLF;
+      break;
+    case 'lfcr':
+      return LineFeed.LFCR;
+      break;
+    case 'lf':
```
nex3: This line probably isn't necessary.

lib/src/render.dart
```diff
@@ -17,7 +16,8 @@ String render(String path,
     {bool color: false,
     SyncPackageResolver packageResolver,
     bool useSpaces: true,
-    int indentWidth: 2}) {
+    int indentWidth: 2,
+    LineFeed linefeed: LineFeed.LF}) {
```
nex3: Generally we avoid providing default argument values for non-boolean arguments, since it makes it impossible to explicitly say "use the default value". You can just use `??` in the method body instead.

Same goes for `indentWidth`, now that I think of it.

lib/src/visitor/serialize.dart
```diff
@@ -19,6 +19,15 @@ import 'interface/css.dart';
 import 'interface/selector.dart';
 import 'interface/value.dart';
 
+enum LineFeed { CR, CRLF, LF, LFCR }
```
nex3: Official Dart enums suck :stuck_out_tongue_closed_eyes:. If you use a custom class, you can just add the text as a field.

Also, I believe these should be all lowercase.

lib/src/visitor/serialize.dart
```diff
@@ -94,16 +105,20 @@ class _SerializeCssVisitor
   /// The number of spaces or tabs to be used for indentation.
   final int _indentWidth;
 
+  final String _linefeed;
```
nex3: Please document this.

srawlins: Done.

lib/src/visitor/serialize.dart
srawlins: They do :frowning_face: Done.

lib/src/render.dart
srawlins: Done.

lib/src/render.dart
srawlins: Done.

lib/src/render.dart
srawlins: Done. and Done for `indentWidth`.

lib/src/node.dart
srawlins: Done.

nex3:
Thanks Sam!

PR #222, opened 2018-01-27
Make function in package script more portable

jathak:
The script in the standalone package currently fails on OSes like Ubuntu where /bin/sh does not support functions declared with the function keyword.

This removes the function keyword from `follow_links`.


nex3:
Thanks!

PR #228, opened 2018-02-03
Fixes bug in visitColor for compressed mode

jathak:
If `value` is not a named color, `name.length` will error.
This checks that `name` is not null before checking the length.


nex3:
Good catch, thanks!

PR #237, opened 2018-02-06
Fixed bug, that parsing sass with selector lists ignored all selectors but the last.

AlexanderJohr:
Fixed bug, that parsing sass with selector lists ignored all selectors but the last. Created a test case which verifies, that sass with selector lists is parsed correctly.

I also used dartfmt to format the files I edited. If this is a problem (there are a lot of changes) please let me know, and I make the pull request again, without using dartfmt.

The error was in this snipped in `lib/src/parse/stylesheet.dart`

``` dart
case $ff:
  if (indented) break loop;
```

When .sass syntax is used the parser searches for the selector by simply parsing until the end of the line. In the scss version this is different, as the selector is parsed until the scanner reaches a curly brace.

For sass syntax, searching for the end of the line isn't enough. The next line also needs to be indented further than the current.

``` dart
case $ff:
  if (indented && peekIndentationOfNextLine() > currentIndentation)
    break loop;
```

The only function which peeks the next line for Indentation was `peekIndentation`, but it also advanced the scanner forward. So I wrote my own `peekIndentationOfNextLine` which only does a peek.
This should probably be refactored, as it is very similar to `peekIndentation`.

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -147,9 +147,11 @@ abstract class StylesheetParser extends Parser {
       var flag = identifier();
       if (flag == 'default') {
         guarded = true;
-      } else if (flag == 'global') {
+      }
+      else if (flag == 'global') {
         global = true;
-      } else {
+      }
+      else {
         scanner.error("Invalid flag name.",
             position: flagStart, length: scanner.position - flagStart);
       }
@@ -283,7 +285,8 @@ abstract class StylesheetParser extends Parser {
       return nameBuffer
         ..write(midBuffer)
         ..writeCharCode($colon);
-    } else if (indented && _lookingAtInterpolatedIdentifier()) {
+    }
+    else if (indented && _lookingAtInterpolatedIdentifier()) {
       // In the indented syntax, `foo:bar` is always considered a selector
       // rather than a property.
       return nameBuffer..write(midBuffer);
@@ -308,7 +311,8 @@ abstract class StylesheetParser extends Parser {
         // properties nested beneath them, so we force an error. This will be
         // caught below and cause the text to be reparsed as a selector.
         if (couldBeSelector) scanner.expectChar($semicolon);
-      } else if (!atEndOfStatement()) {
+      }
+      else if (!atEndOfStatement()) {
         // Force an exception if there isn't a valid end-of-property character
         // but don't consume that character. This will also cause the text to be
         // reparsed.
@@ -329,7 +333,7 @@ abstract class StylesheetParser extends Parser {
     }
 
     var children =
-        lookingAtChildren() ? this.children(_declarationChild) : null;
+    lookingAtChildren() ? this.children(_declarationChild) : null;
     if (children == null) expectStatementSeparator();
 
     return new Declaration(name, scanner.spanFrom(start),
@@ -355,7 +359,7 @@ abstract class StylesheetParser extends Parser {
 
     var value = _declarationExpression();
     var children =
-        lookingAtChildren() ? this.children(_declarationChild) : null;
+    lookingAtChildren() ? this.children(_declarationChild) : null;
     if (children == null) expectStatementSeparator();
     return new Declaration(name, scanner.spanFrom(start),
         value: value, children: children);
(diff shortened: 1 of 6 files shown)


AlexanderJohr:
I ran `pub run grinder`. Certainly made it better. Thanks for the tip!

test/dart_api/function_test.dart
```diff
@@ -130,4 +129,27 @@ main() {
...
+  test("can compile sass with selector list", () {
+    expect(compileString('''
+a,
+b
+  c: d''', indented: true), equals('''
+a,
+b {
+  c: d;
+}'''));
+  });
```
nex3: Language test cases should go in [the sass-spec repo](https://github.com/sass/sass-spec). If you make a pull request there, you can link it here by adding "See sass/sass-spec#number" to the PR description. This will cause Travis to run specs against your PR rather than the master branch of sass-spec.

test/dart_api/function_test.dart
```diff
@@ -130,4 +129,27 @@ main() {
...
+  test("can compile scss with selector list", () {
+    expect(compileString('''
+a,
+b{
+  c: d;
+}''', indented: false), equals('''
+a,
+b {
+  c: d;
+}'''));
```
nex3: I think this case is already covered by sass-spec.

lib/src/parse/stylesheet.dart
```diff
@@ -2269,7 +2269,9 @@ abstract class StylesheetParser extends Parser {
         case $cr:
         case $lf:
         case $ff:
-          if (indented) break loop;
+          if (indented && peekIndentationOfNextLine() > currentIndentation)
```
nex3: I don't think this logic is right. The indented syntax has very specific logic for multiline selectors, which isn't just "fold together everything on adjacent lines with equal or lower indent". The next line should be consumed as a selector if

1. the current line ended with a comma (possibly followed by whitespace and comments), and
2. the next line has the same level of indentation.

This is going to be kind of tricky to implement here, since we consume selectors as `_almostAnyValue()` rather than just going line-by-line like Ruby's parser does.

AlexanderJohr:
When I run the tests I get an error which has nothing to do with the test itself I guess:
Failed to load "C:/sassTest/dartSass\test\node_api\value\boolean_test.dart": Unsupported operation: Illegal character in path: sassTest\dartSass
Any Idea if this is a bug and where I should file it?

lib/src/parse/stylesheet.dart
AlexanderJohr: Yes, I see this logic doesn't seem right. But I thought `_almostAnyValue` should do the same for sass and scss.

After `_almostAnyValue` is called in _styleRule it assigns the value directly to the selector and then continues with the children. For sass and scss the parsed selector are different.

For scss
```dart
    var selector = _almostAnyValue(); // selector: a,
    var children = this.children(_statement);
```
For sass
```dart
    var selector = _almostAnyValue(); // selector: a,\nb
    var children = this.children(_statement);
```

So I think it should be corrected in `_almostAnyValue`

nex3:
Can you provide more details? What command did you invoke? Did it print a stack trace?

AlexanderJohr:
I just right clicked on the test folder in webstorm and clicked on _run 'tests in dart-sass'_.

Stacktrace is:

```
Hint: 2 hint(s) suppressed in package:collection.
Compiled 6,660,059 characters Dart to 1,321,340 characters JavaScript in 13.9 seconds
Dart file (C:\TEMP\dart_test_0198dc7d-1028-11e8-9941-0028f88ad6c5\runInBrowser.dart) compiled to JavaScript: /C:/TEMP/dart_test_0198dc70-1028-11e8-9941-0028f88ad6c5/test_0198dc7c-1028-11e8-9941-0028f88ad6c5/color_test.dart.node_test.dart.js
dart:core                              _Uri.toFilePath
package:path/path.dart 96:21           Object.current
package:path/src/context.dart 62:57    Context.relative
package:path/src/context.dart 451:3    Context.relative
package:path/src/context.dart 1053:18  Context.prettyUri
package:path/path.dart 452:26          Frame.library

Failed to load "C:/sassTest/dartSass\test\node_api\value\color_test.dart": Unsupported operation: Illegal character in path: sassTest\dartSass
```

lib/src/parse/stylesheet.dart
nex3: I think the best way to handle this is to have `_almostAnyValue()` break after *any* newline in Sass mode, and override `styleRule()` in `SassParser` to check for trailing commas and consume additional lines as necessary.

nex3:
Thanks for looking into this! I ended up fixing it on my own to make sure it got into the next beta release, but I appreciate your help tracking down the root cause!

AlexanderJohr:
You're welcome, I appreciate it! Even though I just decided to create a Linux workstation to make the error messages go away and dive deeper. Well then, I'm prepared for the next opportunity to contribute!

PR #295, opened 2018-04-14
Add interactive mode

### LINKED ISSUE #265: Add interactive mode
Like with Ruby Sass's `--interactive`, it would be nice to provide an interactive REPL-style interface for playing around with SassScript.

jathak:
Closes #265.

This adds an interactive REPL that should match the behavior of the one from Ruby Sass as far as I can tell (aside from the error messages being different).

When run with `--interactive` or `-i`, the executable will repeatedly take in a line of SassScript and attempt to parse it as an `Expression` or a `VariableDeclaration`. It will then evaluate the expression in a shared environment and display the result. For declarations, it will also store the result in the environment to be referenced in future evaluations.

To support this, I added constructors `Expression.parse` and `VariableDeclaration.parse` and a function `evaluateExpression` (+ corresponding async version) as suggested.

As for the REPL itself, I had previously written [cli_repl](https://pub.dartlang.org/packages/cli_repl) for another project when I couldn't find any existing package for line editing in Dart. It should allow for simple line editing and history in terminals that support it, though not to the same extent as readline/libedit. I'm using it here, but if there's a better solution, let me know.

I've also looked at supporting `@function` rules in this. The Ruby REPL didn't support them, but would support for them be worth adding here?


jathak:
Sorry for the delay on this; I was caught up in some other work.

I think I've addressed all your comments.

For error logging, it prints the carets underneath the bad input and then prints the message. When color is enabled, both the carets and the input text itself will be highlighted.

lib/src/executable.dart
```diff
@@ -212,33 +172,44 @@ String _writeSourceMap(ExecutableOptions options, SingleMapping sourceMap) {
...
+    } on SassException catch (error, stackTrace) {
+      var highlighted = error.span.highlight();
+      var arrows = highlighted.split('\n').last.trimRight();
+      var buffer = new StringBuffer();
+      if (options.color) {
+        int start = arrows.length - arrows.trimLeft().length;
+        buffer.write("\u001b[1F"); // move to start of input line
+        buffer.write("\u001b[${start + 3}C"); // move to start of error
+        buffer.write("\u001b[31m"); // set color to red
+        buffer.write(line.substring(start, arrows.length)); // write bad input
```
nex3: It looks like this line crashes when the syntax error points after the final character of the string. For example, the input `foo(` will cause it to crash.

lib/src/executable.dart
```diff
@@ -212,33 +172,44 @@ String _writeSourceMap(ExecutableOptions options, SingleMapping sourceMap) {
...
+        buffer.write("\n"); // move to start of output line
+      }
+      buffer.write("   "); // align with start of input
```
nex3: I'd write this as `" " * repl.prompt.length` so if we change the prompt later on it stays correctly aligned.

lib/src/executable_options.dart
```diff
@@ -100,6 +104,24 @@ class ExecutableOptions {
...
+        _options.wasParsed('indented') ||
+        _options.wasParsed('load-path') ||
+        _options.wasParsed('style') ||
+        _options.wasParsed('quiet') ||
+        _options.wasParsed('help') ||
+        _options.wasParsed('source-map') ||
+        _options.wasParsed('source-map-urls') ||
+        _options.wasParsed('embed-sources') ||
+        _options.wasParsed('embed-source-map')) {
+      throw new UsageException("Option not supported with --interactive.");
```
nex3: Include the name of the offending option(s).

lib/src/executable_options.dart
```diff
@@ -100,6 +104,24 @@ class ExecutableOptions {
...
   bool get version => _options['version'] as bool;
 
+  /// Whether to run an interactive shell.
+  bool get interactive {
+    if (!(_options['interactive'] as bool)) return false;
+    if (_options.wasParsed('stdin') ||
+        _options.wasParsed('indented') ||
+        _options.wasParsed('load-path') ||
+        _options.wasParsed('style') ||
+        _options.wasParsed('quiet') ||
```
nex3: I think this makes sense to support--we could produce warnings in interactive mode.

lib/src/executable_options.dart
```diff
@@ -100,6 +104,24 @@ class ExecutableOptions {
...
+        _options.wasParsed('help') ||
```
nex3: `--help` takes precedence over all other options anyway, so there's no need to explicitly disallow it here.

lib/src/visitor/async_evaluate.dart
```diff
@@ -66,12 +66,28 @@ Future<EvaluateResult> evaluateAsync(Stylesheet stylesheet,
...
+Future<Value> evaluateExpressionAsync(Expression expression,
+    {Map<String, Value> variables,
+    Iterable<AsyncCallable> functions,
+    Logger logger}) {
+  var evaluator = new _EvaluateVisitor(
+      functions: functions, logger: logger, sourceMap: false);
+  for (var name in variables.keys) {
+    evaluator._environment
+        .setVariable(name, variables[name], null, global: true);
+  }
```
nex3: Nit, but I'd pass in `variables` to `new EvaluateVisitor()` rather than modifying a private variable. I find it's usually cleaner to avoid accessing another class's privates when possible even if they are visible.

test/cli_shared.dart
```diff
@@ -657,6 +675,53 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
...
+            "          ^^^^^^^^",
+            '\u001b[0mError: Undefined operation "1 + blue".',
+            r">> 1 + $x + 3",
+            "\u001b[1F\u001b[7C\u001b[31m\$x",
+            r"       ^^",
+            "\u001b[0mError: Undefined variable."
+          ]));
+      expect(sass.stdout, emitsDone);
+      await sass.shouldExit(0);
+    });
```
nex3: Test what happens when the user presses enter with a blank line. Right now it crashes due to the parse error problem I described above, but the correct answer is probably to just ignore the empty line.

jathak:
I think I've addressed all your comments. The one thing I'm not sure about is how to handle a line that both emits a warning and errors

example

The normal error logging assumes the input is the line above the current, so extra output causes issues.

Right now, I don't think there's a way to tell whether a warning was emitted or not. Would it make sense to add something to the logger to allow that?

nex3:
> Right now, I don't think there's a way to tell whether a warning was emitted or not. Would it make sense to add something to the logger to allow that?

Ah, really good catch! I don't necessarily want to expand the official `Logger` API, since it's public, but you could create a package-private subclass that tracks whether a warning was emitted. Maybe add `TrackingLogger` in `lib/src/logger/tracking.dart` which wraps an existing logger and adds `emittedWarning` and `emittedDebug` fields.

jathak:
Okay, I did that and then changed it to output a simpler error message when the logger emits something.

Issue #321, opened 2018-05-15
Improve documentation for extending Importer/AsyncImporter.


nshahan:
I am trying to implement an `Importer` and I see the documentation that details the correct order to attempt to find files in the implementation of `canonicalize()`.

1. File name as provided if it has an extension.
2. File name with `.sass` extension added.
3. File name with `.scss` extension added.

https://github.com/sass/dart-sass/blob/454603d160d839bbc8a4bc2687d92e2d8b6c056d/lib/src/importer/async.dart#L35-L40

In practice, it seems like an Importer also needs to look for files that might be a partial but the leading underscore was omitted. Is there a documented correct order to attempt finding all of the previously listed versions of the file as a partial as well?

chriseppstein:
That appears to be what the filesystem importer does.

https://github.com/sass/dart-sass/blob/master/lib/src/importer/filesystem.dart#L19-L22
https://github.com/sass/dart-sass/blob/master/lib/src/importer/utils.dart#L12

If you're looking on disk for a file, you probably want to inherit from FilesystemImporter instead of Importer.

And yes, the API documentation is incomplete here. Would you be able to submit a PR for that?

nshahan:
> And yes, the API documentation is incomplete here. Would you be able to submit a PR for that?

Sure!  One question though; I'm not familiar with the last attempt that treats the path as a directory and tries to read a file there named index. Is that a documented feature of sass `@import` that an Importer should provide as well?

https://github.com/sass/dart-sass/blob/454603d160d839bbc8a4bc2687d92e2d8b6c056d/lib/src/importer/utils.dart#L35-L40

chriseppstein:
Yes, that was recently added to the Sass Spec, the docs should match. Good catch  💯 

https://github.com/sass/sass-spec/pull/1206

PR #325, opened 2018-05-21
Add details to documentation for implementing an importer

nshahan:
Fixes https://github.com/sass/dart-sass/issues/321.

### DIFF
--- a/lib/src/importer/async.dart
+++ b/lib/src/importer/async.dart
@@ -35,9 +35,21 @@ abstract class AsyncImporter {
   /// * If the [url] ends in `.sass` or `.scss`, the importer should look for
   ///   a stylesheet with that exact URL and return `null` if it's not found.
   ///
-  /// * Otherwise, the importer should look for a stylesheet at `"$url.sass"` or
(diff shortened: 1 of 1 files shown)


lib/src/importer/async.dart
```diff
@@ -35,9 +35,21 @@ abstract class AsyncImporter {
...
+  ///   * `"_$url.sass"`
+  ///   * `"$url.sass"`
+  ///   * `"_$url.scss"`
+  ///   * `"$url.scss"`
+  ///
+  /// * Finally, the importer should check for a directory default index.
+  ///   * `"$url<File System Delimiter>_index.sass"`
+  ///   * `"$url<File System Delimiter>index.sass"`
+  ///   * `"$url<File System Delimiter>_index.scss"`
+  ///   * `"$url<File System Delimiter>index.scss"`
```
nex3: I think we should use `/` rather than `<File System Delimiter>` here, since we're dealing with URLs rather than paths. As a bonus, it's also easier to read!

nshahan: Sounds good. Done.

PR #338, opened 2018-05-30
Fix an Importer-calling bug

### LINKED ISSUE #334: Passing packageResolver arg broke in v1.4.0
The sass_builder package is currently using the `packageResolver` arg in `compileString()` to support Dart style package imports. Something broke in v1.4.0. The files fail to be found in the temporary directory that the packageResolver is using. 
```
Error: Can't find stylesheet to import.
@import 'size';
        ^^^^^^
  /private/var/folders/my/nydltsbd5hndsjbt7l850ds4009rgp/T/scratch_spaceenDxCe/packages/angular_components/material_list/_mixins.scss 5:9    @import
  /private/var/folders/my/nydltsbd5hndsjbt7l850ds4009rgp/T/scratch_spaceenDxCe/packages/angular_components/app_layout/_mixins.scss 5:9       @import
  /private/var/folders/my/nydltsbd5hndsjbt7l850ds4009rgp/T/scratch_spaceenDxCe/packages/angular_components_example/gallery/gallery.scss 1:9  root stylesheet

```

If I pin the dependency back to v1.3.2 everything works fine. 

I'm going to be replacing this with an Importer instead which seems to work fine in v1.4.0 but it seems like there might be a bug or some detail I am unintentionally depending on changed.

nex3:
We were resolving URLs relative to the canonical URL rather than the
original URL, which broke importers for which those were different,
like the package importer.

Closes #334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.4.1
+
+* Fix a bug where an importer would be passed an incorrectly-resolved URL when
+  handling a relative import.
+
 ## 1.4.0
 
 * Improve the error message for invalid semicolons in the indented syntax.
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -116,7 +116,7 @@ class AsyncImportCache {
       [AsyncImporter baseImporter, Uri baseUrl]) async {
     var tuple = await canonicalize(url, baseImporter, baseUrl);
     if (tuple == null) return null;
-    var stylesheet = await importCanonical(tuple.item1, tuple.item2);
+    var stylesheet = await importCanonical(tuple.item1, tuple.item2, url);
     return new Tuple2(tuple.item1, stylesheet);
   }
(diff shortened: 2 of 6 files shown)


nshahan:
I tried out this branch with the current version of sass_builder and the problem still exists. It looks like the new tests you added are catching the failure now though.

nex3:
It works for me when I test it locally. Do you have a stand-alone reproduction?

nshahan:
I just made a small repro in the branch: https://github.com/dart-league/sass_builder/tree/sass-break-repro

In the example directory there is a web package you can build with `pub run build_runner build`. It should work. If you change the sass dependency override to 1.4.0 or the dev branch there should be a failure when you build.

https://github.com/dart-league/sass_builder/blob/b88394d42cee2c4ab8b83f629127abf380ae728f/example/pubspec.yaml#L14

nshahan:
I've confirmed it works for me with your latest commit.

PR #339, opened 2018-05-31
Error out on ambiguous imports

### LINKED ISSUE #335: FilesystemImporter attempts to resolve 'file.scss' as a partial first
If you go out of your way and write the full name of the import with an extension the implementation of `FilesystemImporter` will try to load the file as a partial first. It relies on `_tryPath()` when the import has a file extension.
https://github.com/sass/dart-sass/blob/fba35fe0dadf452ced65a610234cb998b7145353/lib/src/importer/utils.dart#L15

`_tryPath()` tries the file as a partial before trying the import as written. 
https://github.com/sass/dart-sass/blob/fba35fe0dadf452ced65a610234cb998b7145353/lib/src/importer/utils.dart#L31-L32

Is this the correct behavior?  If you have a partial file `_foo.scss` and a regular file `foo.scss` at the same location you can't ever load `foo.scss`.

nex3:
Closes #335

See sass/sass-spec#1255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 * Fix a bug where an importer would be passed an incorrectly-resolved URL when
   handling a relative import.
 
+* Throw an error when an import is ambiguous due to a partial and a non-partial
+  with the same name, or multiple files with different extensions. This matches
+  the standard Sass behavior.
+
 ### Command-Line Interface
 
 * Add an `--interactive` flag that supports interactively running Sass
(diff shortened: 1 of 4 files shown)


nex3:
@nshahan Can you take a look at the wording of the documentation?

nshahan:
Looks good, the only thing I see is that it kind of implies that `foo.sass` and `foo.scss` can coexist as long as you use the file extension when you import them. Is the intended?

nex3:
That's accurate. I'll add some documentation and explicit tests for that.

nshahan:
Also, I believe in the current implementation if you create `_foo.scss` and `foo.scss` you can never import `foo.scss` but you can import `_foo.scss` if you use the partial prefix in your import.

PR #356, opened 2018-06-15
Automate benchmarking and update benchmarks

### LINKED ISSUE #293: Update perf stats for latest version
Would be great if you could redo the perf-stats with latest versions, since I think we/I made some substantial improvement with @extend the last year. If you have time to do so :)

nex3:
Closes #293

### DIFF
--- a/tool/grind/utils.dart
+++ b/tool/grind/utils.dart
@@ -2,11 +2,13 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import 'dart:async';
 import 'dart:io';
 
 import 'package:archive/archive.dart';
 import 'package:dart2_constant/convert.dart' as convert;
 import 'package:grinder/grinder.dart';
+import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart';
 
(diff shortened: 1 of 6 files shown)


tool/grind/utils.dart
```diff
@@ -62,3 +64,29 @@ ArchiveFile fileFromString(String path, String contents,
...
+  if (p.url.extension(name) == ".git") name = p.url.withoutExtension(name);
+
+  var path = p.join("build", name);
+
+  if (new Directory(p.join(path, '.git')).existsSync()) {
+    log("Updating $url");
+    await runAsync("git",
+        arguments: ["fetch", "origin"], workingDirectory: path);
+    await runAsync("git",
+        arguments: ["reset", "--hard", "origin/master"],
```
jmesserly: This can be a bit destructive on your local changes, if you ran it by accident. Not only to the working tree but also to the current branch pointer. I wonder if it'd be worth a `git checkout origin/master` just to be safe, so it'd error out if you had local changes.

Since it's only used for other sass impl repos it's probably pretty safe even so, but just wondering if the extra level of `reset --hard` is necessary.

nex3: `checkout` doesn't update the local `master` branch's pointer, which I worry could hypothetically mess up some infrastructure. These repositories are all in the `build/` directory anyway, which is generally expected to be destructively modified by grinder tasks, so I don't feel too bad about doing that here as well.

PR #361, opened 2018-06-18
Add an explicit compatibility policy

nex3:
[skip ci]


nex3:
@jmesserly It looks like Chris may not be able to review this. Can you give me your thoughts?

PR #383, opened 2018-06-25
Run the Dart CLI tests from app snapshots on CI

nex3:
This substantially speeds up the test time, and once we re-enable Dart
2 support is should work around dart-lang/sdk#33257.

### DIFF
--- a/test/cli/dart_test.dart
+++ b/test/cli/dart_test.dart
@@ -50,7 +50,16 @@ Future<TestProcess> runSass(Iterable<String> arguments) {
       (path) => new File(path).existsSync(),
       orElse: () => p.absolute("bin/sass.dart"));
 
+  var args = ["--checked"];
+
+  // Work around dart-lang/sdk#33622.
+  if (Platform.isWindows) args.add("--packages=${p.absolute('.packages')}");
+
   return TestProcess.start(
-      Platform.executable, ["--checked", executable]..addAll(arguments),
-      workingDirectory: d.sandbox, description: "sass");
+      Platform.executable,
+      args
+        ..add(executable)
+        ..addAll(arguments),
+      workingDirectory: d.sandbox,
+      description: "sass");
 }
--- a/test/cli/dart_test.dart
+++ b/test/cli/dart_test.dart
@@ -12,9 +12,19 @@ import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
 import 'package:test_process/test_process.dart';
 
+import '../io.dart';
 import 'shared.dart';
 
+/// Paths where snapshots of the Sass binary might exist, in order of
+/// preference.
+final _snapshotPaths = [
+  p.absolute("build/sass.dart.app.snapshot"),
+  p.absolute("build/sass.dart.snapshot")
+];
+
 void main() {
+  setUpAll(ensureSnapshotUpToDate);
+
   sharedTests(runSass);
 
   test("--version prints the Sass version", () async {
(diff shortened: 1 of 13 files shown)


test/io.dart
```diff
@@ -0,0 +1,45 @@
...
+    entriesToCheck.add(new File("pubspec.yaml"));
+  } else {
+    entriesToCheck.add(new File("pubspec.lock"));
+  }
+
+  for (var entry in entriesToCheck) {
+    if (entry is File) {
+      var entryLastModified = entry.lastModifiedSync();
+      if (lastModified.isBefore(entryLastModified)) {
+        throw "${entry.path} was modified after ${p.prettyUri(p.toUri(path))} "
```
jmesserly: does this cause a stack trace to get printed? if so, might be nicer to print an error message and exit? (or does package:test handle that?)

Or alternatively: maybe it should just call the command to build the snapshot for you?

test/cli/dart_test.dart
```diff
@@ -24,8 +34,32 @@ void main() {
...
+      ensureUpToDate(path, "pub run grinder app-snapshot");
+      return;
+    }
+  }
+}
+
+Future<TestProcess> runSass(Iterable<String> arguments) {
+  var executable = _snapshotPaths.firstWhere(
+      (path) => new File(path).existsSync(),
+      orElse: () => p.absolute("bin/sass.dart"));
```
jmesserly: It seems like once you build a snapshot, you're basically stuck with snapshots? (Unless you delete the build output.) Should it fall back to bin/sass.dart if the snapshots are out of date? Or if snapshots are faster, perhaps it should always require building snapshots?

tool/grind/standalone.dart
```diff
@@ -21,20 +22,30 @@ snapshot() {
...
+/// Compiles Sass to an application snapshot.
+///
+/// If [release] is `true`, this compiles in checked mode. Otherwise, it
+/// compiles in unchecked mode.
+void _appSnapshot({@required bool release}) {
+  var args = [
+    '--snapshot=build/sass.dart.app.snapshot',
+    '--snapshot-kind=app-jit'
+  ];
+  if (!release) args..add('--checked');
```
jmesserly: TODO to remove `--checked` once on Dart 2?

test/cli/dart_test.dart
nex3: > It seems like once you build a snapshot, you're basically stuck with snapshots? (Unless you delete the build output.)

Yeah, that's right.

> Should it fall back to bin/sass.dart if the snapshots are out of date?

This would require checking if the snapshot is up-to-date for every test run, as opposed to once at the beginning of the file. I guess we could get around that with global state, but cross-test state is super risky.

> Or if snapshots are faster, perhaps it should always require building snapshots?

Snapshots are faster if you're running a whole bunch of CLI tests at once, but if you're debugging a single test it's a lot faster to just run from source. Pub has supported both modes for a long time, and I find myself using source for local debugging most of the time and using snapshots on CI and when I need to run a lot of tests at once as a sanity check.

test/io.dart
nex3: > does this cause a stack trace to get printed? if so, might be nicer to print an error message and exit? (or does package:test handle that?)

There's not really a way to fail a test without producing a stack trace--`test` generally tries to provide as much information as possible. The stack trace is very short, though.

tool/grind/standalone.dart
nex3: Done.

PR #391, opened 2018-06-29
Add a --stop-on-error flag

### LINKED ISSUE #264: Support CLI watcher and mtime-aware updater
Dart Sass should support watching and updating flags similar to Ruby Sass:

* [x] Support `source:target` syntax for compiling multiple files.
* [x] Support compiling all non-partial stylesheets in a directory with `source:target`.
* [x] Add an `--update` flag that tells Sass to recompile only those files whose imports have been modified since the target was generated.
* [x] Add a `--watch` flag to update files like `--update`, then continue to watch the sources and anything they import for changes and update the targets as necessary.
* [ ] Add a `--poll` flag to tell `--watch` to use a polling watcher rather than the native OS's watcher.
* [ ] Add a `--stop-on-error` flag to tell Sass to exit as soon as an error is encountered rather than waiting for all compilations to finish.

nex3:
Closes #264

### DIFF
--- a/lib/src/executable/watch.dart
+++ b/lib/src/executable/watch.dart
@@ -41,7 +41,11 @@ Future watch(ExecutableOptions options, StylesheetGraph graph) async {
     var destination = options.sourcesToDestinations[source];
     graph.addCanonical(new FilesystemImporter('.'),
         p.toUri(p.canonicalize(source)), p.toUri(source));
-    await watcher.compile(source, destination, ifModified: true);
+    var success = await watcher.compile(source, destination, ifModified: true);
+    if (!success && options.stopOnError) {
+      dirWatcher.events.listen(null).cancel();
+      return;
+    }
   }
 
   print("Sass is watching for changes. Press Ctrl-C to stop.\n");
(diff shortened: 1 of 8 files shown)


lib/src/executable/watch.dart
```diff
@@ -176,29 +194,38 @@ class _Watcher {
...
+  Future<bool> _compileIfEntrypoint(Uri url) async {
+    if (url.scheme != 'file') return true;
 
     var source = p.fromUri(url);
     var destination = _destinationFor(source);
-    if (destination == null) return;
+    if (destination == null) return true;
 
-    await compile(source, destination);
+    return await compile(source, destination);
```
jmesserly: `return compile(source, destination);`?  (slightly faster/less chars)

(same below)

nex3: I actually strongly prefer `return await`, because it makes the function much more resilient against refactors. If you just write `return`, it's easy to miss that the function you're returning is asynchronous when you come back later, which can lead to dangling futures.

jmesserly: can you explain more what you mean by this? what kind of refactor are you envisioning?

nex3: Suppose I wanted to add error handling, for example. If I had just written `return compile(...)`, it would be easy to accidentally write:

```dart
try {
  return compile(source, destination);
} catch (error) {
  // ...
}
```

But this code is buggy; it won't actually catch exceptions thrown asynchronously by `compile()`. `return await` avoids that issue.

jmesserly: interesting. Seems suboptimal to add `await` everywhere because we might hypothetically add a try+catch to handle async errors around a return. But, up to you.

nex3: I generally try to write code in a way that requires future edits to re-load as little knowledge as possible into their heads about what all the code is doing. I think that's worth a few extra tokens. As for runtime performance, I don't think this makes a measurable difference, and if it ever became a serious problem it would be easy to optimize away at the compiler level.

PR #404, opened 2018-07-10
Run Yarn after updating the Bazel rule's package.json

nex3:
See bazelbuild/rules_sass#36


nex3:
cc @jelbourn

nex3:
We need to check in the updated lockfile to the Sass Bazel repo when we update the Sass version, so this invokes `yarn` to generate that lockfile.

PR #416, opened 2018-07-20
Remove --devel as it's no longer --devel

### LINKED ISSUE #414: Installing dart sass with homebrew. Error: No devel block is defined for sass/sass/sass
I noticed while installing dart-sass with homebrew using the [README.md](https://github.com/sass/dart-sass/blob/master/README.md#from-homebrew-os-x) instructions, it couldn't find the devel block. I assume the `--devel` is used for non stable code, maybe this was meant to be removed and forgotten about. This is the command I followed:
```
brew install --devel sass/sass/sass
```

Removing `--devel` worked fine for me, I am happy to update the documentation however you may intend for it to be installed with `--devel`.

alex-page:
This fixes #414

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ If you use [the Homebrew package manager](https://brew.sh/) for Mac OS X, you
 can install Dart Sass by running
 
 ```sh
-brew install --devel sass/sass/sass
+brew install sass/sass/sass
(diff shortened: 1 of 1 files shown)


nex3:
Thanks!

PR #419, opened 2018-07-21
Improvements in locating the sass binary

adrium:
I don't like adding paths to my `$PATH`. To install sass system-wide, I installed it in `/usr/local/lib/sass` and symlinked the binary to `/usr/local/bin`. The package script does not work with that approach. Also, the path canonicalization logic is simpler with realpath. As readlink's manpage states:

> Note the `realpath` command is the preferred command to use for canonicalization.

I hereby sign the Google Individual Contributor License Agreement.

### DIFF
--- a/package/dart-sass.sh
+++ b/package/dart-sass.sh
@@ -7,19 +7,12 @@
 # executable and a snapshot of Sass. It can be created with `pub run grinder
 # package`.
 
-follow_links() {
-  file="$1"
-  while [ -h "$file" ]; do
-    # On Mac OS, readlink -f doesn't work.
-    file="$(readlink "$file")"
-  done
-  echo "$file"
-}
+path="$(which "$0")"
+path="$(realpath "$path")"
+path="$(dirname "$path")"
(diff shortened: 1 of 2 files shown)


nex3:
I'm sorry, but I'm not allowed to accept your contribution unless you officially sign the CLA with a Google account. I can say that the reason we don't use `realpath` is that it doesn't exist on Mac OS, though.

adrium:
> not allowed to accept your contribution

No problem, it's small anyway. But maybe you get an idea of the issue.

> the reason we don't use realpath is that it doesn't exist on Mac OS

Oh... That's unfortunate. Is the `which` command available? Maybe you can include that?

nex3:
`which` does exist, but I don't know that it provides much benefit since we want the script to run independent of the user's `PATH`.

Testing this locally, I'm unable to reproduce the issue. If I create a symlink to the script, executing that symlink works fine. If you can rpovide a consistent reproduction, feel free to file an issue about it.

PR #427, opened 2018-07-25
Normalize escapes in identifiers

### LINKED ISSUE #425: Properly handle escapes in identifiers
See https://github.com/sass/sass/issues/1542

nex3:
Closes #425
See sass/sass#1542
See sass/sass-spec#1272

### DIFF
--- a/lib/src/parse/parser.dart
+++ b/lib/src/parse/parser.dart
@@ -126,7 +126,7 @@ abstract class Parser {
     } else if (isNameStart(first)) {
       text.writeCharCode(scanner.readChar());
     } else if (first == $backslash) {
-      text.write(escape());
+      text.write(escape(identifierStart: true));
     } else {
       scanner.error("Expected identifier.");
     }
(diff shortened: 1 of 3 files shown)


lib/src/parse/parser.dart
```diff
@@ -364,34 +364,48 @@ abstract class Parser {
...
+    if (identifierStart ? isNameStart(value) : isName(value)) {
+      return new String.fromCharCode(value);
+    } else if (value < 0x1F || value == 0x7F) {
+      var buffer = new StringBuffer()..writeCharCode($backslash);
+      if (value > 0xF) buffer.writeCharCode(hexCharFor(value >> 4));
+      buffer.writeCharCode(hexCharFor(value & 0xF));
+      buffer.writeCharCode($space);
+      return buffer.toString();
+    } else {
+      var buffer = new StringBuffer()
```
jmesserly: could these lines be replaced by:

    return new String.fromCharCodes([$backslash, value]);

nex3: Good thinking. Done.

PR #430, opened 2018-07-31
Add an error() function to Parser

nex3:
This simplifies the process of throwing an error associated with an
existing span.


nex3:
> Love that `@alwaysThrows`.

Thanks for making it real!

PR #434, opened 2018-08-03
Pass --no-preview-dart-2 to the Chocolatey script

### LINKED ISSUE #433: Can't load Kernal binary
Installed Sass yesterday via Chocolatey (0.10.11) and anytime I use anything like --version or --help or anything, it gives me the following error:

Can't load Kernel binary: File size is too small to a valid kernel file.
Error while initializing Kernel isolate

nex3:
I missed this in c462b82.

Closes #433

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.10.3
+
+### Command-Line Interface
+
+* Run the Chocolatey script with the correct arguments so it doesn't crash.
+
 ## 1.10.2
 
 * No user-visible changes.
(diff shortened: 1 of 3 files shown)


nshahan:
This is the process for installing and running Sass on a windows machine? Is this coupled with a version of the SDK that supports the --no-preview-dart-2 flag?

nex3:
I believe all SDK versions that Sass's pubspec allows support that flag. When we distribute via Chocolatey, though, we build with the latest SDK (in `--no-preview-dart-2` mode because it's faster) and then ship a package that depends on that exact SDK version, since snapshots aren't compatible across VM versions.

PR #438, opened 2018-08-09
Support Dart 2 stable

nex3:


appveyor.yml
```diff
@@ -1,7 +1,7 @@
 build: off
 
 branches:
-  only: [master, /^features\..*]
+  only: [master, "/^feature\\..*/", "/^\\d+\\.\\d+\\.(\\d+([+-].*)?|x)$/"]
```
srawlins: A comment would be nice explaining the second string.

nex3: Done.

PR #448, opened 2018-08-14
Remove the source map comment from the JS blob we release

nex3:
See bazelbuild/rules_sass#44


nex3:
cc @alexeagle

alexeagle:
LGTM but it doesn't resolve that issue, see comment there

PR #470, opened 2018-09-07
Properly adjust source maps when prepending encoding information

### LINKED ISSUE #469: Chained source map breaks if @charset not removed
I'm working on integrating Dart Sass to my build tool and wanted to report this strange experience.

**Build workflow goes: Sass --> PostCSS (Autoprefixer) --> clean-css**

I'm gluing source maps using https://www.npmjs.com/package/merge-source-map (which works fantastically and apparently also being used by Vue)

If I don't remove `@charset "UTF-8";\n` from the very top of the Sass-compiled CSS output before feeding the CSS into PostCSS, the resulting source map will be off by 4-5 lines for some reason...

> (Browser points to the right file of the CSS rule but wrong lines)

```ts
let sassResult = await this.runSassAsync(sassOptions);
let css: string = sassResult.css.toString('utf8');
let charsetHeader = '@charset "UTF-8";\n';
if (css.startsWith(charsetHeader)) {
    css = css.substring(charsetHeader.length);
}
```

But when I remove that line, the resulting source map will be accurate!

> [I suspect this is due to source map not being aware of the lines being added to the final CSS?](https://github.com/sass/dart-sass/blob/3b6730369bdff9bb7859411c82e4b21f194d2514/lib/src/visitor/serialize.dart#L73) I'm not 100% sure though.

nex3:
Closes #469

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.13.3
+
+* Properly generate source maps for stylesheets that emit `@charset`
+  declarations.
+
 ## 1.13.2
 
 * Properly parse `:nth-child()` and `:nth-last-child()` selectors with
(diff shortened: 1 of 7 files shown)


ryanelian:
Thank you for fixing this!

Should probably also double-check other places (if exists) where CSS is being changed without updating the source map.

PR #476, opened 2018-09-19
[WIP] Set up CI with Azure Pipelines

nex3:


kaylangan:
I'm a Program Manager on Azure Pipelines. Let me know if you have any questions or suggestions. I'm happy to help!

nex3:
Thanks Kayla! I had a few questions:

* When a build is triggered by a pull request, is it possible to access the text of that pull request? This is one of our longstanding issues with Travis.

* Is there a way to store encrypted credentials in the configuration file itself, similar to Travis's `secure:` feature?

I also had [one suggestion](https://visualstudio.uservoice.com/forums/330519-visual-studio-team-services/suggestions/34253173-support-ansi-escape-codes-for-color-in-build-outpu) that would make the system a lot more usable for us.

nschonni:
>     * Do you need to access the text of the pull request from the build itself?  Several variables related to pull requests [are available](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=vsts#system-variables) during builds, but we don't yet have one for the PR text.  Would it meet your needs if we added one named `System.PullRequest.Description`?  In the meantime, could your build query our REST API to get the PR description?  I could help with that.

The PR description is actually used currently to drive what would be the `matrix` in Pipelines. EX: The description in https://github.com/sass/sass-spec/pull/1286 tells the runner what platforms that this should be passing for

> One thought (you may already be aware): if your tests are already outputting JUnit results, you can publish them to Azure Pipelines for reports/trends/etc. by adding the [Publish Test Results](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=vsts&tabs=yaml) task to your YAML ([example test results](https://dev.azure.com/vscode/VSCode/_build/results?buildId=8062&view=ms.vss-test-web.test-result-details)).

Forgot to trace down the UV or DC for these things since it looks like a server-side change rather than a vsts-task-lib, but [TAP](https://en.wikipedia.org/wiki/Test_Anything_Protocol) support would be useful for the general non-.Net community rather than using a second reporters to try and translate to JUnit

nex3:
Closing this out because I don't think it's the road we're going to take for now. It remains a future possibility, but I'm pretty happy with Travis and there are better places to spend energy than changing that.

PR #481, opened 2018-09-21
Prettify URIs

### LINKED ISSUE #464: [Polish] Relative Paths should be consistent
Issue Type: Polish

Let's say you type in this command into a command prompt: `sass --watch ./scss/main.scss ./css/main.min.css`, the compiler then prints the first path `./scss/main.scss` as this: `scss\main.scss`, and leaves the second path completely how it is, here's a screenshot, look at the bottom:
image

`Compiled scss\main.scss to ./css/main.min.css`

This should really be polished, either:
- use backslashes OR forward slashes
- include the WHOLE path (including the `./`) or leaving small "prefixes" out
 
Dart Sass 1.13.0 64-bit
Windows 10 1803, 17134.228

beastdestroyer:
This issue closes #464.
The changes I have made include prettifying the destination variable with this: `p.prettyUri(destination)` at line 82. In the relevant issue, I raised two questions:
 
> - use backslashes OR forward slashes
> - include the WHOLE path (including the ./) or leaving small "prefixes" out

- To answer the first prompt. In order not to confuse anyone who is used to current system, I have not made any changes. In Windows it will still print `\` and other OSes will print `/`.
- To answer the second prompt, I have left the small prefixes out. This is to ensure that nested directories with lots of `../../` and other crowded file prefixes won't clog up the screen. One disadvantage is for people who have many of the same filenames, not knowing which file has been compiled, however that should really be a problem for more disorganised systems (no saltiness).
@nex3

### DIFF
--- a/lib/src/executable/compile_stylesheet.dart
+++ b/lib/src/executable/compile_stylesheet.dart
@@ -79,6 +79,7 @@ Future compileStylesheet(ExecutableOptions options, StylesheetGraph graph,
   if (options.color) buffer.write('\u001b[32m');
 
   var sourceName = source == null ? 'stdin' : p.prettyUri(p.toUri(source));
+  destination = p.prettyUri(destination);
   buffer.write('Compiled $sourceName to $destination.');
   if (options.color) buffer.write('\u001b[0m');
   print(buffer);
--- a/lib/src/executable/compile_stylesheet.dart
+++ b/lib/src/executable/compile_stylesheet.dart
@@ -79,7 +79,7 @@ Future compileStylesheet(ExecutableOptions options, StylesheetGraph graph,
   if (options.color) buffer.write('\u001b[32m');
 
   var sourceName = source == null ? 'stdin' : p.prettyUri(p.toUri(source));
-  var destinationName = p.prettyUri(destination);
+  var destinationName = p.prettyUri(p.toUri(destination));
   buffer.write('Compiled $sourceName to $destinationName.');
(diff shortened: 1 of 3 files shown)


beastdestroyer:
I have signed it

beastdestroyer:
@nex3

lib/src/executable/compile_stylesheet.dart
```diff
@@ -79,6 +79,7 @@ Future compileStylesheet(ExecutableOptions options, StylesheetGraph graph,
   if (options.color) buffer.write('\u001b[32m');
 
   var sourceName = source == null ? 'stdin' : p.prettyUri(p.toUri(source));
+  destination = p.prettyUri(destination);
```
nex3: Rather than modifying this variable, can you either create a new `destinationVariable` or include the call to `prettyUri()` inline in the `buffer.write()` call?

beastdestroyer: Would something like `var prettyDestination = p.prettyUri(destination);`
and then replacing `$destination` with `$prettyDestination` work?

nex3: I would prefer `destinationName`, since that matches `sourceName`, but yes.

beastdestroyer: Thanks, will do!

beastdestroyer: Done!

PR #483, opened 2018-09-25
Fix a parsing bug involving nested content blocks

nex3:
See sass/sass-spec#1285


nex3:
I'm holding off on landing this pending discussion of whether categories should be able to work like folders (see https://github.com/dart-lang/dartdoc/pull/1761#issuecomment-424125587).

PR #488, opened 2018-10-03
Fix an import-resolution bug

nex3:
When a stylesheet is imported, the parsed stylesheet object is cached
based on its canonical URL. However, the stylesheet.span.sourceUrl was
based on the text of the import that was used to load that stylesheet.
The idea was to make the source URL in stack traces look nicer, but it
meant that relative URLs could be resolved based on the old importer's
URL before being sent to the new importer, which caused bugs.

Now stylesheet.span.sourceUrl is always the canonical URL of the
stylesheet, and thus safe to cache. We then use the import cache to
convert the canonical URL to a human-friendly URL at the point at
which we generate stack traces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
 
 * Don't crash when using `@debug` in a stylesheet passed on standard input.
 
+### Dart API
+
+* `AsyncImporter.canonicalize()` and `Importer.canonicalize()` must now return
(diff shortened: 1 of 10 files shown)


nex3:
@jmesserly I (believe I) fixed a bunch of tests, and I deprecated relative canonical URLs as per our in-person discussion. PTAL

PR #503, opened 2018-10-17
Add a ParenthesizedExpression class

nex3:
This allows us to accurately track the source spans for parenthesized
expressions, which in turn allows us to print accurate error
indications.

Adding a new class for this more accurately represents the structure
of the expression, but it also involves an extra allocation during
parsing and an extra level of nesting during evaluation which could
have a small but real performance impact.

We could alternatively add a package-internal setter for
Expression.span, and update the source span for parenthesized
expressions after they're initially parsed. However, this has its own
downsides: it adds complexity and mutability to the object model; and
many expression classes currently use lazily-generated spans, so
making them settable would require adding extra slots on those
classes.

I decided to go with the extra class because it only adds overhead
when parentheses are actually used in practice, as opposed to adding
overhead to every list/color/etc. The runtime overhead is also likely
to be mitigated if at any point we add a constant-folding step.

See sass/sass-spec#1293

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -1772,7 +1772,7 @@ relase. For details, see http://bit.ly/moz-document.
 
       if (!scanner.scanChar($comma)) {
         scanner.expectChar($rparen);
-        return first;
+        return new ParenthesizedExpression(first, scanner.spanFrom(start));
       }
       whitespace();
 
--- a/test/cli/shared.dart
+++ b/test/cli/shared.dart
@@ -253,6 +253,22 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
     await sass.shouldExit(65);
   });
 
+  // Regression test for an issue mentioned in sass/linter#15
+  test("gracefully reports errors for binary operations with parentheized ""operands", () async {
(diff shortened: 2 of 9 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -1772,7 +1772,7 @@ relase. For details, see http://bit.ly/moz-document.
 
       if (!scanner.scanChar($comma)) {
         scanner.expectChar($rparen);
-        return first;
+        return new ParenthesizedExpression(first, scanner.spanFrom(start));
```
srawlins: It is nice that this is an exceedingly simple implementation. Like you say in the description, it only adds complexity when parens are used, and this is the only line in source code that instantiates new Parenthesized Expression nodes.

test/cli/shared.dart
```diff
@@ -253,6 +253,22 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
     await sass.shouldExit(65);
   });
 
+  // Regression test for an issue mentioned in sass/linter#15
+  test("gracefully reports errors for binary operations with parentheized ""operands", () async {
```
srawlins: `""operands` typo?

srawlins:
Thanks again!

PR #534, opened 2018-12-02
Fix typo in async_evaluate.dart

egml:
Line [602](https://github.com/sass/dart-sass/blob/master/lib/src/visitor/async_evaluate.dart#L602): "compound selectors may longer be extended" -> "compound selectors may no longer be extended"


nex3:
Closed by #576.

PR #541, opened 2018-12-10
Don't canonicalize file case in source maps

### LINKED ISSUE #540: File names are converted to lowercase in source map
Say my Main.scss file has this:
`@import "TestFile";`

Then I run: 
`sass --source-map Main.scss Main.css`

Then the generated source map will have:
`sources: ["main.scss","testfile.scss"]`

This will cause file not found in case sensitive web server, http://xxx.xxx.xxx/mian.scss will not get the content from http://xxx.xxx.xxx/Mian.scss
May I know if this is by design? And if there is any walkaround to make the generated source list match the original file name case?

Thanks!

nex3:
Closes #540

### DIFF
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -407,7 +407,7 @@ class ExecutableOptions {
   Uri sourceMapUrl(Uri url, String destination) {
     if (url.scheme.isNotEmpty && url.scheme != 'file') return url;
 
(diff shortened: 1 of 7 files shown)


nex3:
Adding tests first, without implementation, to verify that they fail on Windows and thus that the fix actually does something.

PR #551, opened 2018-12-17
Match Node Sass's this.includePaths behavior for importers

### LINKED ISSUE #549: `includePaths` should joined by platform-specific path separator
I found that parsed `includePaths` are separated by colon regardless of platform. It may cause errors on Windows.

https://github.com/sass/dart-sass/blob/94fd7e6e503a977e60bb8c7b894ac9fd3d597a45/lib/src/node.dart#L256

nex3:
On Windows, paths are separated with a semicolon rather than a colon.

Closes #549

### DIFF
--- a/lib/src/io/vm.dart
+++ b/lib/src/io/vm.dart
@@ -82,8 +82,7 @@ DateTime modificationTime(String path) {
   return stat.modified;
 }
 
-String getEnvironmentVariable(String name) =>
-    io.Platform.environment['SASS_PATH'];
+String getEnvironmentVariable(String name) => io.Platform.environment[name];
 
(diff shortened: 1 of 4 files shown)


lib/src/io/vm.dart
```diff
@@ -82,8 +82,7 @@ DateTime modificationTime(String path) {
   return stat.modified;
 }
 
-String getEnvironmentVariable(String name) =>
-    io.Platform.environment['SASS_PATH'];
```
nshahan: Nothing was relying on this always being 'SASS_PATH'?

nex3: No, we were always passing it in explicitly as well (which is why it worked in the Node implementation).

PR #581, opened 2019-01-29
Allow individual directories to be passed on the command-line

### LINKED ISSUE #543: Support passing individual directories to the CLI
A year ago I developed a website with ruby-sass. Now with dart-sass I'm trying to get the website running again. So I installed dart-sass, added it to the PATH (on Windows). I am using the MinGW shell on Windows, added the folder under "C:/Program Files (x86)/Sass". However, now I'm getting a confusing error:

```
sass.bat --watch ./website/public/css
--watch is not allowed when printing to stdout.
```

What does that mean - the entire point of the "watch" command is to hot-reload and print errors to stdout. Why can't I just watch a directory on Windows? I can get my files to compile, if I add `> /dev/null` at the end, however, then dart-sass immediately exits (but compiles the files, at least). However, that's not hot-reloading, so... what's the point of the watch command, then? I'm not sure if this is a bug related to Windows - the error also appears if I run sass via cmd.

All I want is for sass to start (like ruby-sass did) and watch my files, so that it prints errors as it hot-reloads the files - however, how am I supposed to see the error output without using stdout? And how do I get hot-reload to work at all under Windows?

nex3:
Closes #543

### DIFF
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -218,21 +218,22 @@ class ExecutableOptions {
     var stdin = _options['stdin'] as bool;
     if (_options.rest.isEmpty && !stdin) _fail("Compile Sass to CSS.");
 
+    var directories = Set<String>();
     var colonArgs = false;
     var positionalArgs = false;
     for (var argument in _options.rest) {
       if (argument.isEmpty) _fail('Invalid argument "".');
 
-      // If the colon appears at position 1, treat it as a Windows drive
-      // letter.
-      if (!argument.contains(":") ||
-          (_isWindowsPath(argument, 0) &&
+      if (argument.contains(":") &&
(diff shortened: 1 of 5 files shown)


lib/src/executable/options.dart
```diff
@@ -251,6 +252,8 @@ class ExecutableOptions {
             {null: _options.rest.isEmpty ? null : _options.rest.first});
       } else if (_options.rest.length > 2) {
         _fail("Only two positional args may be passed.");
+      } else if (directories.isNotEmpty) {
+        _fail('Directory "${directories.first}" may not be a positional arg.');
```
nshahan: Does this mean it is valid to envoke `sass input.scss output.css` but not `sass input_dir output_dir`? If so maybe the error message would be more clear if you say something that suggests the `input_dir:output_dir` syntax.

PR #592, opened 2019-02-12
Add support for parsing @use

nex3:
See https://github.com/sass/sass-spec/pull/1341

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -535,8 +535,7 @@ abstract class StylesheetParser extends Parser {
         return supportsRule(start);
       case "use":
         _isUseAllowed = wasUseAllowed;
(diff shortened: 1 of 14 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -494,6 +533,11 @@ abstract class StylesheetParser extends Parser {
         return _disallowedAtRule(start);
       case "supports":
         return supportsRule(start);
+      case "use":
+        _isUseAllowed = wasUseAllowed;
+        if (!root || !_isUseAllowed) _disallowedAtRule(start);
+        if (!_parseUse) _disallowedAtRule(start);
```
nshahan: Might be worth having a more specific error message here since this is a new feature and users might bump into it when trying to migrate from using `@import`.

PR #593, opened 2019-02-13
Merge branch 'feature.use' into master

nex3:
This disables @use parsing, but otherwise matches the feature branch
exactly. This will allow us to ensure the @use implementation doesn't
drift away from the release branch.

See https://github.com/sass/sass-spec/pull/1342

### DIFF
--- a/tool/travis/sass-spec-ref.sh
+++ b/tool/travis/sass-spec-ref.sh
@@ -7,7 +7,7 @@
 # run. If we're running specs for a pull request which refers to a sass-spec
 # pull request, we'll run against the latter rather than sass-spec master.
 
-default=feature.use
+default=master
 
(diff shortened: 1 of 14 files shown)


tool/travis/sass-spec-ref.sh
```diff
@@ -7,7 +7,7 @@
 # run. If we're running specs for a pull request which refers to a sass-spec
 # pull request, we'll run against the latter rather than sass-spec master.
 
-default=master
+default=feature.use
```
nshahan: Should this still be master on this branch?

nex3: Yes! Good catch!

nshahan: Then, next question I just though of... Why was travis green when running the feature.use spec when it should be failing?

PR #596, opened 2019-02-15
Fix bug in RecursiveStatementVisitor

jathak:
The value of an `AtRule` can be null, so it should not be visited in that case.

Ran across this issue when I attempted to run the module migrator on a stylesheet containing `@font-face` (which has children, but no value).


nex3:
Thanks!

PR #608, opened 2019-02-27
Adding pedantic

domesticmouse:
Adding pedantic along with a bunch of minor cleanups. PTAL and let me know what you think. Thanks!


nex3:
Thanks for the contribution! However, I'm not really into adding a bunch of lints that don't clearly improve the code, especially when some of them (`deprecated_member_use_from_same_package` for example) are pure negative value.

PR #614, opened 2019-03-04
https-ify sass-lang.com urls

asottile:
See https://github.com/sass/sass-site/issues/217

Committed via https://github.com/asottile/all-repos


nex3:
Thanks!

PR #620, opened 2019-03-11
Emit CSS for used modules

nex3:
See https://github.com/sass/sass-spec/pull/1357

### DIFF
--- a/lib/src/async_environment.dart
+++ b/lib/src/async_environment.dart
@@ -29,6 +29,10 @@ class AsyncEnvironment {
   /// This is `null` if there are no namespaceless modules.
   Set<AsyncModule> _globalModules;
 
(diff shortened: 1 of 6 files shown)


lib/src/async_environment.dart
```diff
@@ -122,6 +126,7 @@ class AsyncEnvironment {
   AsyncEnvironment({bool sourceMap = false})
       : _modules = {},
         _globalModules = null,
+        _allModules = [],
```
nshahan: Why the mix of representation between `_globalModules = null` and `_allModules = []`? Is this just an optimistic optimization in hopes that global modules will be rare?

nex3: Exactly.

PR #634, opened 2019-03-27
Include CSS for modules used by imported files

nex3:
https://github.com/sass/sass-spec/pull/1367

### DIFF
--- a/lib/src/ast/css/at_rule.dart
+++ b/lib/src/ast/css/at_rule.dart
@@ -2,6 +2,7 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import '../../visitor/interface/css.dart';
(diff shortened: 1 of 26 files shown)


nex3:
We have two separate ASTs: one for Sass source, and one for plain CSS. When we load a `@use` through an `@import`, we first convert it to CSS (we can't evaluate it as Sass directly because each module's Sass source should only be executed once, and it may be imported multiple times). Then we execute that CSS more or less as though it were Sass (we can't inject it into the stylesheet as-is because the `@import` may be nested in other rules). Hence the duplication. I'll expand the comment in `_EvaluateVisitor` to explain this better.

Unfortunately, this duplication will outlast `@import`, because it's also necessary for implementing the new `load-css()` mixin. It's necessary as long as there's any way to inject CSS from one file into a nested context, which users desire.

PR #637, opened 2019-04-03
Only claim "package:" URLs are unsupported on Node

### LINKED ISSUE #630: If import URL is wrong but even if package urls are supported, dart-sass complains: "package:" URLs aren't supported on this platform.
I get the Error `"package:" URLs aren't supported on this platform`.

The error is resolved when the path is correct.

I ran into this issue when I renamed the package and hence the import urls needed to be updated.

dart-sass tries to import and if it didn't work and it is a package URL, it doesn't check if it was just a wrong url. It just assumes, that it is because package aren't isn't supported on that platform, even when they are.

It can be found in \visitor\evaluate.dart and \visitor\async_evaluate.dart:

https://github.com/sass/dart-sass/blob/bc5e3c4f8f220cc4f9aefa79d3d6354ffdaeffb2/lib/src/visitor/async_evaluate.dart#L854

https://github.com/sass/dart-sass/blob/bc5e3c4f8f220cc4f9aefa79d3d6354ffdaeffb2/lib/src/visitor/evaluate.dart#L854

```dart
// [...]
    try {
      if (_nodeImporter != null) {
        var stylesheet = _importLikeNode(url);
        if (stylesheet != null) return Tuple2(null, stylesheet);
      } else {
        var tuple = _importCache.import(
            Uri.parse(url), _importer, _stylesheet.span?.sourceUrl);
        if (tuple != null) return tuple;
      }
 
// AlexanderJohr: If the program reached here, the import didn't work. Could be because of wrong url or because package URLs aren't supported, but it just assumes, that it is because of package URLs not supported on the platform.
      if (url.startsWith('package:')) {
        // Special-case this error message, since it's tripped people up in the
        // past.
        throw "\"package:\" URLs aren't supported on this platform.";
      } else {
        throw "Can't find stylesheet to import.";
      }
    } on SassException catch (error) {
// [...]
```

But for now I'm not sure how to fix it myself.

nex3:
Closes #630

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.17.5
+
+* Don't claim that "package:" URLs aren't supported when they actually are.
+
 ## 1.17.4
 
 * Consistently parse U+000C FORM FEED, U+000D CARRIAGE RETURN, and sequences of
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -27,6 +27,7 @@ import '../extend/extender.dart';
 import '../importer.dart';
 import '../importer/node.dart';
 import '../importer/utils.dart';
+import '../io.dart';
 import '../logger.dart';
 import '../parse/keyframe_selector.dart';
 import '../syntax.dart';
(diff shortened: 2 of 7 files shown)


jathak:
Looks like `lib/src/visitor/evaluate.dart` needs to be updated

nex3:
> Looks like `lib/src/visitor/evaluate.dart` needs to be updated

Done.

PR #638, opened 2019-04-03
Fix importing files relative to "package:" imports

### LINKED ISSUE #631: Relative paths do not work within partials imported via package:* import
In trying to create a standalone compile script that utilizes `packageResolver` - it seems that relative paths are broken once the compiler enters a file that was imported via a `package:` import.

```scss
// package_two/sass/app.scss

@import 'package:package_one/sass/api';
```

```scss
// package_one/sass/_api.scss
.foo {
  color: rebeccapurple;
}

// Throws "Can't find stylesheet to import." when _api.scss is imported by package_two
// > https://github.com/sass/dart-sass/blob/a0e63ac704c6c98c8a9123872e3a8c1f312bb2ed/lib/src/visitor/evaluate.dart#L859
@import '../some/relative/path/_some_partial.scss'; 
```

This can be reproduced by cloning / pulling [this branch of the Dart `bootstrap` package](https://github.com/alademann/bootstrap_for_pub/tree/package-resolver-bug-case) locally, running `pub get`, and then running:

```
dart ./tool/compile_css.dart
```

from the root of the package.

The `packageResolver` can be found within the `compile_css.dart` script where the call to `sass.compile()` is made.

@nex3 

cc/ @greglittlefield-wf

nex3:
The PackageImporter wasn't accepting paths relative to its
canonicalized outputs as inputs.

Closes #631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,14 @@
+## 1.17.5
+
+* Fix importing files relative to `package:`-imported files.
+
+### Dart API
+
+* Explicitly require that importers' `canonicalize()` methods be able to take
+  paths relative to their outputs as valid inputs. This isn't considered a
+  breaking change because the importer infrastructure already required this in
+  practice.
+
 ## 1.17.4
 
 * Consistently parse U+000C FORM FEED, U+000D CARRIAGE RETURN, and sequences of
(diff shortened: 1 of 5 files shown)


aaronlademann-wf:
Thank you @nex3 !!!!!!!!!

PR #641, opened 2019-04-03
Don't recursively list directories in realCasePath()

### LINKED ISSUE #636: Sass requesting access to my AddressBook on mac after installing from npm?
```
nvm use lts/dubnium
npm install -g sass
sass --version # 1.17.3 compiled with dart2js 2.2.0
sass foo.scss bar.css
```

Asks to access my AddressBook? When I deny it, it says:
```
Error reading Library/Application Support/AddressBook: operation not permitted.
```

nex3:
We only need to list the path's immediate parent directory in order to
find its real case.

Closes #636

### DIFF
--- a/lib/src/io/node.dart
+++ b/lib/src/io/node.dart
@@ -191,14 +191,23 @@ void ensureDir(String path) {
   });
 }
 
-Iterable<String> listDir(String path) {
-  Iterable<String> list(String parent) =>
-      _fs.readdirSync(parent).expand((child) {
-        var path = p.join(parent, child as String);
-        return dirExists(path) ? listDir(path) : [path];
-      });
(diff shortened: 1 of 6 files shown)


lib/src/io/node.dart
```diff
@@ -191,14 +191,23 @@ void ensureDir(String path) {
...
+    if (!recursive) {
+      return _fs
+          .readdirSync(parent)
+          .map((child) => p.join(parent, child as String))
+          .where((path) => !dirExists(path));
+    } else {
+      Iterable<String> list(String parent) =>
+          _fs.readdirSync(parent).expand((child) {
+            var path = p.join(parent, child as String);
+            return dirExists(path) ? list(path) : [path];
```
srawlins: nice local recursive function

PR #654, opened 2019-04-23
Make sass compiler compatible with bazel persistent worker.

fenghaolw:
This is required before we can turn on persistent worker support in [sass_binary bazel rule](https://github.com/bazelbuild/rules_sass).

Tested that `dartanalyzer lib test` and `pub run test -x node` all pass.

### DIFF
--- a/lib/src/executable.dart
+++ b/lib/src/executable.dart
@@ -3,8 +3,10 @@
 // https://opensource.org/licenses/MIT.
 
 import 'dart:async';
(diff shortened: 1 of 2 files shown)


nex3:
We talked about this offline, but I want to avoid building in Bazel knowledge, dependencies, and maintenance overhead into the core Sass distribution. This is something that should be added as a wrapper around Sass, like all other build system integrations.

PR #658, opened 2019-04-29
Add support for attribute selector modifiers

### LINKED ISSUE #656: Case-insensitive attribute selectors are not parsed
Based on this standard: https://www.w3.org/TR/selectors-4/#attribute-case
```
p[title='test' i] {
	color: red;
}
```
In dart-sass it causes parse error:
```
Error: expected "]".
  ╷
1 │ p[value="test" i]{
  │                ^
  ╵
  sass\test.scss 1:16  root stylesheet
```
It seems it doesn't yet support this standard.  But in Ruby Sass it works as expected and just copies attribute selector as is. Example codepen: https://codepen.io/SagePtr/pen/qwgwQw
In node-sass (based on libsass) it also works as intended.

 ```
$ sass --version
1.19.0 compiled with dart2js 2.2.0
```

nex3:
Closes #656
See sass/sass-spec#1386


jathak:
It seems like Travis is failing because it's using an old version of `.travis.yml` that says to run `tool/travis/test.sh`.

nex3:
@jathak Fixed; PTAL

PR #659, opened 2019-04-30
Generate a CSS file describing an error when one occurs

### LINKED ISSUE #657: Include error messages in compiled output
Hi,

if I install Ruby-Sass, I can see all the errors on browser but with the other sass versions ( sass --version: 1.19.0) I can see all errors on Terminal but nothing on browser and this is not so useful. Thats way I have installed the Ruby-Sass, which is not recommended.

Has anyone idea How to show sass compilation errors in the browser with the Sass-version 1.19.0.

I am using Mac-Majove and working on Wordpress project. 
Sass version is "Ruby Sass 3.7.4"

thanks 
JQ

nex3:
Closes #657


nshahan:
I'm concerned about what this will do to the downstream workflows that use the sass bazel rules. Where before no file was output and there was a build breakage will there now be an output file so the build will continue?

nex3:
I believe Bazel (and hopefully other build tools) considers the executable's status code the canonical source of information about whether it succeeded or failed, not which files it produced.

PR #685, opened 2019-05-17
Merge master into feature.use

nex3:


lib/src/parse/stylesheet.dart
```diff
@@ -28,7 +28,7 @@ import 'parser.dart';
 /// `feature.use` branch. It allows us to avoid having separate development
 /// tracks as much as possible without shipping `@use` support until we're
 /// ready.
-const _parseUse = true;
+const _parseUse = false;
```
jathak: This should stay true on `feature.use`, right?

tool/travis/sass-spec-ref.sh
```diff
@@ -7,7 +7,7 @@
 # run. If we're running specs for a pull request which refers to a sass-spec
 # pull request, we'll run against the latter rather than sass-spec master.
 
-default=feature.use
+default=master
```
jathak: Same here

PR #687, opened 2019-05-17
Resolve imports using custom imports before load paths

### LINKED ISSUE #681: Importers should take precedence over load paths in the JS API
To match the behavior of Node Sass, importers should take precedence over load paths.

nex3:
Closes #681

### DIFF
--- a/lib/src/importer/node/implementation.dart
+++ b/lib/src/importer/node/implementation.dart
@@ -34,10 +34,10 @@ import '../utils.dart';
 /// * The order of import precedence is as follows:
 ///
 ///   1. Filesystem imports relative to the base file.
-///   2. Filesystem imports relative to the working directory.
-///   3. Filesystem imports relative to an `includePaths` path.
-///   3. Filesystem imports relative to a `SASS_PATH` path.
-///   4. Custom importer imports.
+///   2. Custom importer imports.
+///   3. Filesystem imports relative to the working directory.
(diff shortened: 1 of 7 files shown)


lib/src/importer/node/implementation.dart
```diff
@@ -81,6 +81,11 @@ class NodeImporter {
       if (value != null) return _handleImportResult(url, previous, value);
     }
 
+    if (parsed.scheme == '' || parsed.scheme == 'file') {
+      var result = _resolveLoadPath(p.fromUri(parsed), previous);
+      if (result != null) return result;
+    }
```
jelbourn: Dedupe with the same block in `loadAsync`?

lib/src/importer/node/implementation.dart
```diff
@@ -104,23 +109,35 @@ class NodeImporter {
       if (value != null) return _handleImportResult(url, previous, value);
     }
 
+    if (parsed.scheme == '' || parsed.scheme == 'file') {
+      var result = _resolveLoadPath(p.fromUri(parsed), previous);
+      if (result != null) return result;
```
jelbourn: Could this just `return _resolveLoadPath(p.fromUri(parsed), previous);` since the function would return `null` anyway if it goes passed this block?

nex3: It definitely could!

PR #693, opened 2019-05-23
Factor out a function to load modules

nex3:
This will allow us to re-use logic for @forward. It also fixes some
usability issues where incorrect or duplicated spans were being used
for @use errors.

See sass/sass-spec#1392

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -175,7 +175,11 @@ class _EvaluateVisitor
   /// imports, it contains the URL passed to the `@import`.
   final _includedFiles = Set<String>();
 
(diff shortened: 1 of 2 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -356,14 +361,40 @@ class _EvaluateVisitor
     return expression.accept(this);
   }
 
+  /// Loads the module at [url] and passes it to [callback].
+  ///
+  /// The [stackFrame] is used as the name of the stack frame in which the new
+  /// module is executed.
+  Future<void> _loadModule(Uri url, String stackFrame, AstNode nodeForSpan,
```
nshahan: Add `nodeForSpan` to dartdoc.

PR #709, opened 2019-05-31
Enable stricter analysis options, especially around typing

nex3:


analysis_options.yaml
```diff
@@ -1,5 +1,17 @@
...
+    strict-raw-types: true
   errors:
     missing_js_lib_annotation: ignore
+    deprecated_member_use_from_same_package: ignore
+
+    # These are necessary for matching the JS API.
+    avoid_types_as_parameter_names: ignore
+
+    # This has tons of false positives for StreamSubscription.close().
+    unawaited_futures: ignore
```
srawlins: Nice; it's great that you can include pedantic, and then ignore specific lints.

lib/src/node/render_options.dart
```diff
@@ -13,7 +13,7 @@ class RenderOptions {
   external String get data;
   external dynamic get importer;
   external dynamic get functions;
-  external List get includePaths; // contains Strings
+  external List<Object /* String */ > get includePaths;
```
srawlins: Hmm, why can't it be typed `List<String>`?

lib/src/visitor/async_evaluate.dart
```diff
@@ -42,7 +42,8 @@ import 'interface/modifiable_css.dart';
 import 'interface/statement.dart';
 
 /// A function that takes a callback with no arguments.
-typedef _ScopeCallback = Future Function(Future Function() callback);
+typedef _ScopeCallback = Future<void> Function(
```
srawlins: I really like these fixes; nice for readers to see that something is  `Future<void>`.

lib/src/visitor/async_evaluate.dart
```diff
@@ -1580,6 +1581,7 @@ class _EvaluateVisitor
 
     _verifyArguments(positional.length, named, IfExpression.declaration, node);
 
+    // ignore: prefer_is_empty
```
srawlins: I think this qualifies as a bug for `prefer_is_empty`. Have you filed  one?

lib/src/node/render_options.dart
nex3: Because this comes from JS, its runtime is `JSArray`, which doesn't have a reified generic type. Without something like https://github.com/dart-lang/sdk/issues/30554, we have no way of causing it to produce a properly reified type.

lib/src/visitor/async_evaluate.dart
nex3: It seems valid to me... `positional.isNotEmpty` would be be better style than `positional.length > 0` almost all the time, except for here where it's contrasting with other length checks.

Issue #554, opened 2018-12-26
Support the new module system


nex3:
See https://github.com/sass/sass/issues/1094

ryanelian:
Hi. I have a question about `@use` and `@forward`: is custom `importer` use supported with these?

nex3:
Yes, `@use` and `@forward` use the same importer infrastructure as `@import`. As with all module loads, two files with the same URL will only be loaded once.

PR #716, opened 2019-06-06
Add built-in modules for core Sass functions

nex3:
Partially addresses #554

See https://github.com/sass/sass-spec/pull/1400


nex3:
It might be a good idea to look at the color file; that module has the most differences between the module-system version and the global-function version, which means the most changes from the old structure.

nshahan:
Thanks I'll give a closer look now.

nshahan:
To confirm, should I expect that the functions in this chart that do not appear in the global module should be removed in this change?
https://github.com/sass/sass/blob/master/accepted/module-system.md#built-in-modules-1

For example, should `red`, `green`, `blue` be removed from the global module and only available in color?

nex3:
No, everything that was available globally is still available globally, for backwards compatibility. The functions that are listed as global in that chart are ones that are intended to *always* remain in the global namespace.

nex3:
Thanks Nick!

PR #721, opened 2019-06-12
Add module-variables() and module-functions()

nex3:
Partially addresses #554

See sass/sass-spec#1403

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -302,6 +302,19 @@ class _EvaluateVisitor
         });
       }),
 
+      BuiltInCallable("module-functions", r"$module", (arguments) {
+        var namespace = arguments[0].assertString("module");
+        var module = _environment.modules[namespace.text];
+        if (module == null) {
+          throw 'There is no module with namespace "${namespace.text}".';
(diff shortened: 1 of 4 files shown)


nshahan:
Does this travis run need attention?

nex3:
It's failing because of the issue you pointed out in the specs. I've restarted it, so it should work now.

lib/src/visitor/async_evaluate.dart
```diff
@@ -282,6 +282,32 @@ class _EvaluateVisitor
...
+      BuiltInCallable("module-variables", r"$module", (arguments) {
+        var namespace = arguments[0].assertString("module");
+        var module = _environment.modules[namespace.text];
+        if (module == null) {
+          throw 'There is no module with namespace "${namespace.text}".';
+        }
+
+        return SassMap({
+          for (var entry in module.variables.entries)
+            SassString(entry.key): entry.value
```
nshahan: Any danger of exposing something you don't want to by returning this `Value` through the map without reboxing it in some way?

nex3: No, `Value`s are intentionally only ever normal Sass values.

PR #724, opened 2019-06-18
Merge feature.use into master

nex3:
See https://github.com/sass/sass-spec/pull/1411


nex3:
This isn't actually ready for review yet; I want to wait until https://github.com/sass/dart-sass/pull/722 has landed as well.

nex3:
Now it's ready!

lib/src/visitor/async_evaluate.dart
```diff
@@ -282,14 +427,52 @@ class _EvaluateVisitor
         callback);
   }
 
+  /// Loads the module at [url] and passes it to [callback].
+  ///
+  /// The [stackFrame] and [nodeForSpan] are used for the name and location of
+  /// the stack frame in which the new module is executed.
```
jathak: Maybe clarify that the stack frame is why this passes `Module` to a callback instead of just returning `Future<Module>`? This confused me before I realized that

PR #727, opened 2019-06-19
Require dependencies in the preamble rather than through JS-interop

### LINKED ISSUE #726: Webpack cannot statically analyse dart-sass `self.require()`
The JS-compiled dart-sass output starts with a pattern resembling:

```javascript
var self = Object.create(global);
self.require = require;
```

When bundling through webpack, it is unable to statically analyse the `require` due to the re-assignment, resulting in a bundle time warning and runtime failure. It introduces a `webpackEmptyContext` which will prevent any `require`s.

I've prepared a more thorough write up on [StackOverflow](https://stackoverflow.com/questions/56653326/webpacking-sass-dart-sass-uses-require-dynamically-hard-code-dependencies) and in [this gist](https://gist.github.com/tchetwin/eeeffdfbbf973b59f05a3ecba3ab253a). I'm in the process of answering the SO post with a current workaround. Filing this Issue in response to [this tweet](https://twitter.com/SassCSS/status/1141138173738668032).

Thanks!

nex3:
Closes #726

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
 * Add a top-level `warn()` function for custom functions and importers to print
   warning messages.
 
+### JavaScript API
+
+* Avoid re-assigning the `require()` function to make the code statically
+  analyzable by Webpack.
+
 ## 1.20.3
 
 * No user-visible changes.
--- a/lib/src/io/node.dart
+++ b/lib/src/io/node.dart
@@ -87,10 +87,8 @@ class Stderr {
   void flush() {}
 }
 
-@JS("require")
-external _FS _require(String name);
-
-final _fs = _require("fs");
+@JS("fs")
+external _FS get _fs;
 
 @JS("process")
 external _Process get _process;
--- a/lib/src/node/chokidar.dart
+++ b/lib/src/node/chokidar.dart
@@ -4,9 +4,6 @@
 
 import 'package:js/js.dart';
 
-@JS("require")
-external Chokidar _require(String name);
-
 @JS()
 class Chokidar {
   external ChokidarWatcher watch(String path, ChokidarOptions options);
@@ -30,4 +27,5 @@ class ChokidarWatcher {
 /// The Chokidar module.
 ///
 /// See [the docs on npm](https://www.npmjs.com/package/chokidar).
-final chokidar = _require("chokidar");
+@JS("chokidar")
+external Chokidar get chokidar;
--- a/tool/grind/npm.dart
+++ b/tool/grind/npm.dart
@@ -50,7 +50,11 @@ void _js({@required bool release}) {
         text.replaceFirst(RegExp(r"\n*//# sourceMappingURL=[^\n]+\n*$"), "\n");
   }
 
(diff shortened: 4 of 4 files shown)


nex3:
@tchetwin Does this fix your Webpack issues?

tchetwin:
It very well might! Would need to see JS generated output, I'll take a look tomorrow, thanks!

nex3:
@tchetwin Here's [the compiled JS file](https://github.com/sass/dart-sass/files/3307908/sass.dart.js.gz) (gzipped to allow me to upload it as a GitHub attachment).

tchetwin:
First, thank you so much for these changes, in their current form they will satisfy *our use-case* without additional webpack configuration! 🎉 

With some tweaks I think *all* use cases could be satisfied, and without any webpack warning messages.

---

The relevant parts of sass.dart.js now look like (comments my own):

```javascript
// This is still going on, which makes webpack nervous and causes a static analysability error
var self = Object.create(global);
self.require = require;

// These now use un-aliased `require`, webpack can get behind these!
self.fs = require("fs");
self.chokidar = require("chokidar");

// "readline" is still using `self.require`, usage that relies on this dependency will presumably fail post-webpack
u($,"MX","Ho",function(){return self.require("readline")})
```
These lines result in the following when webpacked:

1. `webpackEmptyContext` still present due to the `self.require = require` assignment
    * Emits a warning message when packaging tooo
2. "readline" is not identified and wired up

Applying the same technique in this PR again here should fix 2:

https://github.com/sass/dart-sass/blob/425305725b87f1f45f1cb81cc2ee1073ddfcf3b1/lib/src/node/chokidar.dart#L33

Perhaps removing this would fix 1 as well:

https://github.com/sass/dart-sass/blob/425305725b87f1f45f1cb81cc2ee1073ddfcf3b1/lib/src/node/chokidar.dart#L7-L8

nex3:
It's harder to make `readline` work, since we don't control that code, but I guess we can post-process the JS...

nex3:
@tchetwin Can you verify that [the latest generated JS](https://github.com/sass/dart-sass/files/3311604/sass.dart.js.gz) doesn't produce those Webpack warnings?

jathak:
Would there be any way to restructure [this code](https://github.com/jathak/cli_repl/blob/master/lib/src/repl_adapter/node.dart) to avoid the dynamic require?

nex3:
Good question, but I don't think so... as a library, `cli_repl` doesn't have any control over how it's built to JS, so it can't inject a static require. The principled way to fix this would be to build in support for JS modules to Dart's JS interop, but work on that seems to have stagnated.

tchetwin:
Sorry for the delay. It looks great!

There's no `webpackEmptyContext` and webpack correctly finds all three `require`s:

```javascript
self.fs = __webpack_require__(/*! fs */ \"fs\");\nself.chokidar = __webpack_require__(/*! chokidar */ \"./node_modules/chokidar/index.js\");\nself.readline = __webpack_require__(/*! readline */ \"readline\");
```

PR #728, opened 2019-06-21
Add support for "@use with"

nex3:
See sass/sass-spec#1412

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -449,8 +449,7 @@ class _EvaluateVisitor
     if (builtInModule != null) {
       if (configuration != null || _configuration != null) {
         throw _exception(
-            "This variable was not declared with !default in the @used module.",
-            configuration.values.first.configurationSpan);
+            "Built-in modules can't be configured.", nodeForSpan.span);
       }
 
       callback(builtInModule);
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -37,6 +37,8 @@ import '../module/built_in.dart';
 import '../parse/keyframe_selector.dart';
 import '../syntax.dart';
 import '../util/fixed_length_list_builder.dart';
+import '../util/limited_map_view.dart';
+import '../util/unprefixed_map_view.dart';
 import '../utils.dart';
 import '../value.dart';
 import '../warn.dart';
@@ -238,6 +240,10 @@ class _EvaluateVisitor
   /// module.
   Extender _extender;
 
+  /// A map from variable names to the values that override their `!default`
+  /// definitions in this module.
+  Map<String, _ConfiguredValue> _configuration;
+
   _EvaluateVisitor(
       {AsyncImportCache importCache,
       NodeImporter nodeImporter,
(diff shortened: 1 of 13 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -429,12 +435,24 @@ class _EvaluateVisitor
...
+  ///
   /// The [stackFrame] and [nodeForSpan] are used for the name and location of
   /// the stack frame in which the new module is executed.
   Future<void> _loadModule(Uri url, String stackFrame, AstNode nodeForSpan,
-      void callback(Module module)) async {
+      void callback(Module module),
+      {Map<String, _ConfiguredValue> configuration}) async {
     var builtInModule = _builtInModules[url];
     if (builtInModule != null) {
+      if (configuration != null || _configuration != null) {
```
nshahan: What exactly is this guarding against? Could the error message be more specific?

lib/src/visitor/async_evaluate.dart
```diff
@@ -467,62 +486,90 @@ class _EvaluateVisitor
...
 
-        await visitStylesheet(stylesheet);
-        css = _outOfOrderImports == null
-            ? _root
-            : CssStylesheet(_addOutOfOrderImports(), stylesheet.span);
+    var alreadyLoaded = _modules[url];
+    if (alreadyLoaded != null) {
+      if (configuration != null || _configuration != null) {
+        throw _exception(
+            "This module was already loaded, so it can't be configured using "
```
nshahan: Can this message include the the name of the module?

lib/src/visitor/async_evaluate.dart
```diff
@@ -467,62 +486,90 @@ class _EvaluateVisitor
...
+      _styleRule = oldStyleRule;
+      _mediaQueries = oldMediaQueries;
+      _declarationName = oldDeclarationName;
+      _inUnknownAtRule = oldInUnknownAtRule;
+      _atRootExcludingStyleRule = oldAtRootExcludingStyleRule;
+      _inKeyframes = oldInKeyframes;
+
+      if (configuration != null && _configuration.isNotEmpty) {
+        throw _exception(
+            "This variable was not declared with !default in the @used module.",
```
nshahan: Can the message include the name of the variable and module?

lib/src/visitor/async_evaluate.dart
nex3: This is guarding against someone passing in configuration to a built-in module, which never expose variables. I used the same error message as elsewhere because I want users to think of built-in modules as essentially the same kind of thing as user-defined modules, but if you think it would be substantially clearer I could change this to something like "Built-in modules can't be configured."

nshahan: Yeah I think that error with the updated message would be more clear and prevent someone from thinking they could possibly configure some built-in module if they just had the right variable in their config.

nex3: Done.

nex3:
Tests are failing due to an analyzer deprecation that's fixed in #735. I'll have to merge that into feature.use and then merge feature.use into this to fix the issue, but you should be able to review in the meantime.

PR #741, opened 2019-06-27
Add load-css()

nex3:
See sass/sass-spec#1417

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -380,7 +380,9 @@ class _EvaluateVisitor
       })
     ];
 
-    var metaModule = BuiltInModule("meta", [...meta.global, ...metaFunctions]);
+    var metaModule =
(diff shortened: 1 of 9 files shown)


nex3:
The tests are failing because of a new analysis error Dart 2.4.0 is producing, but this should be fit to review anyway.

lib/src/visitor/async_evaluate.dart
```diff
@@ -1323,30 +1387,41 @@ class _EvaluateVisitor
...
+            }
+          });
+          return null;
         });
         return null;
       });
-      return null;
-    });
+    } else {
+      throw UnsupportedError("Unknown callable type $mixin.");
```
nshahan: Did you mean to print the `.runtimeType` here?

nex3: I was just planning on relying on the object's toString(). I thought `Object.runtimeType` was in the language team's crosshairs.

Issue #113, opened 2017-02-03
Make the JS version faster


nex3:
The JS-compiled version of Dart Sass hovers around 2.5x and 3x slower than when running on the Dart VM. Some slowdown is inevitable, but this seems extreme. We should put dedicated effort into profiling the JS output and optimizing hot spots.

ranquild:
Common thing with dart2js performance issues is dynamic dispatch. For example, in `dynamic s = func(); s.method();`. call of `method` will be compiled to something like `J.getInterceptor(s).method$0();`, where `getInterceptor` - is sort-of runtime prototype wrapper that allows to call non-existing methods on JS primitives. Types like `List` or `String` are not wrapped in compiled dart code, they are `Array` and `String` in JS. So, when you have dynamic variable and call method/property like `first`, interceptor returns correct prototype for object by doing types checks for `String`, `Array`, `int` and so on. 

What that does mean is that usage of `dynamic` ruins compiled-to-js performance and should be avoided at all costs. Easy workaround is to use `Object` for variables that could contain objects of different types and doing `is` checks and calling method on variable of specific types. Using `dynamic` in VM does not have those performance issues and that's why there is so huge performance gap. 

P.S. Working in a company with > 500k lines of dart code.

nex3:
See [this test file](https://github.com/alex-page/sass-a11ycolor/blob/5c719eb75e7c2462e40c90a0e9b3cd9b6518761d/test/test1.scss) as a test case where JS is vastly slower than the Dart VM.

nex3:
@ogonkov Can you share the stylesheets that are running so much slower?

ogonkov:
I'm afraid no, i will try to reproduce the issue, by excluding sass files, but don't know when

nex3:
[This comment](https://github.com/sass/dart-sass/issues/206#issue-283828041) also includes a zip file that's reported to have poor JS performance.

nex3:
Dart Sass 1.13.4 should substantially improve performance for many of the worst cases. I've verified that it speeds up @alex-page's a11ycolor example from about 11s to about 3s on my computer, which puts it within the expected 2-3x slower than the Dart VM. If anyone has any more examples where the JS speed is 4x or more slower than LibSass or the Dart VM, please include them here.

I'm keeping this issue open because I still think there are likely to be code changes we can make that will further improve the JS speed.

alex-page:
[...]

PR #760, opened 2019-07-12
Improve the JS performance for logic-heavy stylesheets

nex3:
See #113

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -1962,15 +1962,19 @@ class _EvaluateVisitor
       {bool trackSpans}) async {
     trackSpans ??= _sourceMap;
 
-    var positional = (await mapAsync(arguments.positional,
-            (Expression expression) => expression.accept(this)))
-        .toList();
+    var positional = [
+      for (var expression in arguments.positional) await expression.accept(this)
+    ];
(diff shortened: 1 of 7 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -1962,15 +1962,19 @@ class _EvaluateVisitor
...
     var named = await normalizedMapMapAsync<String, Expression, Value>(
         arguments.named,
         value: (_, expression) => expression.accept(this));
 
-    var positionalNodes =
-        trackSpans ? arguments.positional.map(_expressionNode).toList() : null;
+    var positionalNodes = trackSpans
+        ? [
+            for (var expression in arguments.positional)
+              _expressionNode(expression)
```
srawlins: Good that this new syntax is performant!

tool/grind/npm.dart
```diff
@@ -40,7 +40,7 @@ void _js({@required bool release}) {
     // * We expect our test coverage to ensure that nothing throws subtypes of
     //   Error.
     // * We thoroughly test edge cases in user input.
-    if (release) ...["-O4", "--fast-startup"]
+    if (release) ...["-O4", "--no-minify", "--fast-startup"]
```
srawlins: Should we comment on why we pass `no-minify`?

PR #761, opened 2019-07-14
Fix typo in silent comment error message

mgreter:
See sass/sass-spec#1432


nex3:
Thanks for the fix! Can you also update the newly-failing spec?

mgreter:
There you go https://github.com/sass/sass-spec/pull/1432

nex3:
Thanks! For future reference, when making changes across both Dart Sass and sass-spec, please link the CLs as described in step 4 of https://github.com/sass/sass/blob/master/CONTRIBUTING.md. I've done it for you for this PR though!

PR #765, opened 2019-07-17
Merge origin/master into feature.use

nex3:
See sass/sass-spec#1436


nex3:
@mgreter Can you post a comment here that contains only "I consent."?

mgreter:
I consent.

nex3:
Thanks!

PR #778, opened 2019-07-22
Release 1.22.8

nex3:
This doesn't contain any code changes relative to 1.22.7, but the JS
release will pick up mbullington/node_preamble.dart#16.


jathak:
It looks like this is failing on node v8.x since `url.pathToFileURL` wasn't added until v10.x (the same thing is happening for sass/migrator#70)

nex3:
Good catch, @jathak. I've limited the Node version range.

PR #779, opened 2019-07-24
Ensure bundling compatibility

AviVahl:
- declare "browser" field. webpack and other bundlers pick up this field and avoid bundling these deps.
- applied CI (min node version) changes from other release PR.
- ~~upgrade node_preamble to 1.4.5 so that sass.dart.js bundle is Worker-compatible as well. checked locally. removal of the `self.require = require` is still needed.~~

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -21,5 +21,10 @@
   },
   "main": "sass.dart.js",
   "bin": "sass.js",
-  "keywords": ["style", "scss", "sass", "preprocessor", "css"]
+  "keywords": ["style", "scss", "sass", "preprocessor", "css"],
+  "browser": {
+    "fs": false,
+    "chokidar": false,
+    "readline": false
+  }
 }
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -37,7 +37,7 @@ dev_dependencies:
   grinder: "^0.8.0"
   http: ">=0.11.0 <0.13.0"
   js: "^0.6.0"
-  node_preamble: "^1.1.0"
+  node_preamble: "^1.4.5"
   pub_semver: "^1.0.0"
   source_span: "^1.5.2"
   stream_channel: ">=1.0.0 <3.0.0"
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -37,7 +37,7 @@ dev_dependencies:
   grinder: "^0.8.0"
   http: ">=0.11.0 <0.13.0"
   js: "^0.6.0"
-  node_preamble: "^1.4.5"
+  node_preamble: "^1.1.0"
   pub_semver: "^1.0.0"
   source_span: "^1.5.2"
   stream_channel: ">=1.0.0 <3.0.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ jobs:
   - &node-tests
     name: Node tests | Dart stable | Node stable
     language: node_js
-    node_js: 10
+    node_js: stable
     install: pub run grinder before-test
     script: tool/travis/task/node_tests.sh
   - <<: *node-tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,7 @@ jobs:
   - &node-tests
     name: Node tests | Dart stable | Node stable
     language: node_js
-    node_js: stable
+    node_js: 10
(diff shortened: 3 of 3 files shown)


AviVahl:
I signed it!

AviVahl:
hey @nex3, any chance you can review this? :)

package/package.json
```diff
@@ -14,12 +14,17 @@
...
+    "node": ">=10.0.0"
   },
   "dependencies": {
     "chokidar": ">=2.0.0 <4.0.0"
   },
   "main": "sass.dart.js",
   "bin": "sass.js",
-  "keywords": ["style", "scss", "sass", "preprocessor", "css"]
+  "keywords": ["style", "scss", "sass", "preprocessor", "css"],
+  "browser": {
```
nex3: Can you explain what this field is doing? It doesn't seem to match npm's [documentation for this field](https://docs.npmjs.com/files/package.json#browser).

pubspec.yaml
```diff
@@ -37,7 +37,7 @@ dev_dependencies:
   grinder: "^0.8.0"
   http: ">=0.11.0 <0.13.0"
   js: "^0.6.0"
-  node_preamble: "^1.1.0"
+  node_preamble: "^1.4.5"
```
nex3: I would prefer to leave this as-is. We aren't relying on any *API features* of more recent preamble versions, we're just taking advantage of bug fixes.

package/package.json
AviVahl: https://github.com/defunctzombie/package-browser-field-spec
basically, it's a field used by bundlers to remap module requests inside a package (when bundling it for the web).

many projects use it, such as axios, react, etc.

pubspec.yaml
AviVahl: reverting, although I could really use a `sass.dart.js` that has 1.4.5+, as it fixes several issues I've encountered with 1.1.0.

AviVahl:
@nex3 thx for the quick review.

package/package.json
nex3: Can you walk me through the use-case here? Dart Sass doesn't support running in the browser, so what benefit does this provide, exactly?

pubspec.yaml
nex3: We'll download the latest version regardless of whether the version range forces it to be downloaded, unless doing so would cause a version conflict.

package/package.json
AviVahl: I wanted to run render/renderSync inside a Web Worker, in order to compile `.sass/scss` code into `.css`.

With these changes I can bundle it and run it in the browser, so I'm not sure what you mean by "doesn't support".

nex3: Dart Sass is built with the assumption that built-in Node.js modules will always be available. If they're not, it's liable to crash in unexpected ways if you invoke it wrong. 

If the goal here is to invoke Dart Sass from within the browser, I want to do it right: by addressing #25 and not adding the default filesystem importer in a web context, and documenting that as an option.

AviVahl: I understand wanting to provide a full built-in support for running `sass` with different hosts (other than node, i.e. `deno` or other runtimes).

> Dart Sass is built with the assumption that built-in Node.js modules will always be available. If they're not, it's liable to crash in unexpected ways if you invoke it wrong.

Several node APIs (`url`, `Buffer`, `process` etc.) are polyfill-able and are being automatically polyfilled by bundlers out there. The "browser" field enables me to to ignore other ones (`fs`). As long as sass' process doesn't try to use that `fs`, nothing breaks.

My guess was that `chokidar` and `readline` are used for by the CLI, and so are not relevant to a browser-based flow.

In addition, I found that the existing `data` and `importer` options allow me to "connect" my custom (browser-compatible) fs and resolver to sass' process. This enables me to bypass the need for sass to require an fs provider and achieve rendering of `.scss/sass` files in the browser.

imho, this PR is a good first step towards resolving issue #25.

AviVahl:
rebased to current `master`.

package/package.json
AviVahl: Further investigation into @import handling revealed the gaps in terms of custom hosts. I now see what you mean.

AviVahl:
Alright I'm closing until a more complete solution is achieved.

PR #784, opened 2019-07-29
Allow @use in the REPL

nex3:


lib/src/ast/sass/statement/use_rule.dart
```diff
@@ -31,6 +33,14 @@ class UseRule implements Statement {
       {Map<String, Tuple2<Expression, FileSpan>> configuration})
       : configuration = Map.unmodifiable(configuration ?? const {});
 
+  /// Parses a variable declaration from [contents].
```
jathak: This should say "use rule" instead of "variable declaration"

lib/src/executable/repl.dart
```diff
@@ -74,3 +82,12 @@ void _logError(SassException error, StackTrace stackTrace, String line,
...
 }
+
+// /// Returns whether [span] refers to subset of portion of the repl input [line]
+// /// that's indicated by the span's column number.
+// ///
+// /// Always returns `false` for multi-line spans since they can never be
+// /// highlighted inline.
+// bool _lineContainsSpan(String line, SourceSpan span) =>
+//     span.start.line == span.end.line &&
+//     line.substring(span.start.column, span.end.column) == indexOf(span.text) == span.start.column;
```
jathak: Remove this

lib/src/ast/sass/statement/use_rule.dart
nex3: Done.

lib/src/executable/repl.dart
nex3: Oops 🤪. Done.

PR #794, opened 2019-08-09
Build the Mac OS deployment on Mac OS

nex3:
This allows it to build a native executable, rather than a script
snapshot. I don't know why we weren't doing this already.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -221,6 +221,7 @@ jobs:
     if: *deploy-if
     env: *github-env
     script: skip
+    os: osx
     deploy:
       provider: script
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -221,6 +221,7 @@ jobs:
     if: *deploy-if
     env: *github-env
     script: skip
+    os: osx
```
nshahan: That is all it takes? \o/

nex3: I think so... it's hard to test, though, so we'll find out next time I cut a release!

PR #803, opened 2019-08-20
Change variable syntax to namespace.$name

nex3:
See sass/sass#2731
See sass/sass-spec#1459


lib/src/parse/stylesheet.dart
```diff
@@ -604,44 +683,51 @@ abstract class StylesheetParser extends Parser {
   }
 
   /// Consumes an at-rule allowed within a function.
```
jathak: This comment should be updated to reflect the new behavior of this function.

PR #842, opened 2019-10-08
Typo fixes

XhmikosR:


XhmikosR:
Do I need to update any other files so that the checksums match?

If so, could someone else do this so that I don't set up dart just for this? Thanks!

PR #849, opened 2019-10-16
Fix #417 preserve location of trailing loud comments

### LINKED ISSUE #417: Inline comments should be kept inline
This is a "repeat" of the ticket made at sass/sass#235 which presented the problem of code like:
```
.myselector {
  display: none; /* javascript will fix this */
}
```
compiling to:
```
.myselector {
  display: none;
  /* javascript will fix this */
}
```
That ticket was moved to ruby-sass and closed as on ice since it was deemed to be specific to Ruby Sass's deprecated implementation. However, I'm seeing it happen for me despite using Dart Sass. I use `--style expanded` for what it's worth.

My apologies if this is somehow being caused by interference with an old and probably only partially working ruby (when I run `sass --help` I am seeing references to Dart, e.g. `    --version                  Print the version of Dart Sass.`, though don't know if there's still possible room for interference, or how I might resolve that). Thanks for any possible work on this, or letting me know if there is simply some flag or style setting whose documentation I missed.

nickbehrens:
Fix #417 by preserving the location of trailing loud comments.

Loud comment AST nodes now have the notion of whether they're "trailing".  During parsing, we look for loud comments at the end of lines and if found, tag them as trailing.  This information is then plumbed through and made available during serialization.  

During serialization, where we would previously unconditionally insert newlines between nodes, we now check to see if the next node is a trailing loud comment.  If so, we instead write out a space followed by the trailing loud comment.

Ran the following locally, all tests passing:
$ pub run grinder && dartanalyzer lib test && pub run test -x node

See https://github.com/sass/sass-spec/pull/1485

### DIFF
--- a/lib/src/visitor/serialize.dart
+++ b/lib/src/visitor/serialize.dart
@@ -210,7 +210,7 @@ class _SerializeVisitor
 
     if (!node.isChildless) {
       _writeOptionalSpace();
-      _visitChildren(node.children, node);
+      _visitChildren(node);
     }
   }
 
@@ -228,7 +228,7 @@ class _SerializeVisitor
     });
 
     _writeOptionalSpace();
-    _visitChildren(node.children, node);
+    _visitChildren(node);
   }
 
   void visitCssImport(CssImport node) {
@@ -275,7 +275,7 @@ class _SerializeVisitor
         () =>
             _writeBetween(node.selector.value, _commaSeparator, _buffer.write));
     _writeOptionalSpace();
-    _visitChildren(node.children, node);
+    _visitChildren(node);
   }
 
   void _visitMediaQuery(CssMediaQuery query) {
@@ -300,7 +300,7 @@ class _SerializeVisitor
 
     _for(node.selector, () => node.selector.value.accept(this));
     _writeOptionalSpace();
-    _visitChildren(node.children, node);
+    _visitChildren(node);
   }
 
   void visitCssSupportsRule(CssSupportsRule node) {
@@ -317,7 +317,7 @@ class _SerializeVisitor
     });
 
     _writeOptionalSpace();
-    _visitChildren(node.children, node);
+    _visitChildren(node);
   }
 
   void visitCssDeclaration(CssDeclaration node) {
@@ -1288,19 +1288,20 @@ class _SerializeVisitor
   void _write(CssValue<String> value) =>
       _for(value, () => _buffer.write(value.value));
 
-  /// Emits [children] in a block.
-  void _visitChildren(List<CssNode> children, CssParentNode parent) {
+  /// Emits [parent.children] in a block.
+  void _visitChildren(CssParentNode parent) {
     _buffer.writeCharCode($lbrace);
 
-    CssNode? prePrevious_;
-    CssNode previous = parent;
-    for (var child in children) {
+    CssNode? prePrevious;
+    CssNode? previous;
+    for (var child in parent.children) {
       if (_isInvisible(child)) continue;
-      if (previous != parent && _requiresSemicolon(previous)) {
+
+      if (previous != null && _requiresSemicolon(previous)) {
         _buffer.writeCharCode($semicolon);
       }
 
-      if (_isTrailingComment(child, previous)) {
+      if (_isTrailingComment(child, previous ?? parent)) {
         _writeOptionalSpace();
         _withoutIndendation(() => child.accept(this));
       } else {
@@ -1310,17 +1311,16 @@ class _SerializeVisitor
         });
       }
 
-      prePrevious_ = previous;
+      prePrevious = previous;
       previous = child;
     }
 
-    if (previous != parent) {
+    if (previous != null) {
       if (_requiresSemicolon(previous) && !_isCompressed) {
         _buffer.writeCharCode($semicolon);
       }
 
-      if (prePrevious_ == parent &&
-          _isTrailingComment(previous, prePrevious_!)) {
+      if (prePrevious == null && _isTrailingComment(previous, parent)) {
         _writeOptionalSpace();
       } else {
         _writeLineFeed();
@@ -1332,7 +1332,7 @@ class _SerializeVisitor
   }
 
   /// Whether [node] requires a semicolon to be written after it.
-  bool _requiresSemicolon(CssNode? node) =>
+  bool _requiresSemicolon(CssNode node) =>
       node is CssParentNode ? node.isChildless : node is! CssComment;
 
   /// Whether [node] represents a trailing comment when it appears after
@@ -1349,8 +1349,8 @@ class _SerializeVisitor
     var previousSpan = previous.span;
     if (previous is CssParentNode && previous.children.contains(node)) {
       // Walk back from just before the current node starts looking for the
-      // parent's left brace (to open the child block).  This is safer than
-      // a simple forward search of the previousSpan.text as that might contain
+      // parent's left brace (to open the child block). This is safer than a
+      // simple forward search of the previousSpan.text as that might contain
       // other left braces.
       var searchFrom = node.span.start.offset - previousSpan.start.offset - 1;
       if (searchFrom < 0) {
--- a/lib/src/visitor/serialize.dart
+++ b/lib/src/visitor/serialize.dart
@@ -1319,7 +1319,8 @@ class _SerializeVisitor
         _buffer.writeCharCode($semicolon);
       }
 
-      if (prePrevious_ == parent && _isTrailingComment(previous, prePrevious_)) {
+      if (prePrevious_ == parent &&
+          _isTrailingComment(previous, prePrevious_!)) {
         _writeOptionalSpace();
       } else {
         _writeLineFeed();
--- a/lib/src/visitor/serialize.dart
+++ b/lib/src/visitor/serialize.dart
@@ -21,6 +21,7 @@ import '../util/character.dart';
 import '../util/no_source_map_buffer.dart';
 import '../util/number.dart';
 import '../util/source_map_buffer.dart';
+import '../util/span.dart';
 import '../value.dart';
 import 'interface/css.dart';
 import 'interface/selector.dart';
@@ -1346,29 +1347,20 @@ class _SerializeVisitor
     if (_isCompressed) return false;
     if (node is! CssComment) return false;
 
-    var previousSpan = previous.span;
-    if (previous is CssParentNode && previous.children.contains(node)) {
-      // Walk back from just before the current node starts looking for the
-      // parent's left brace (to open the child block). This is safer than a
-      // simple forward search of the previousSpan.text as that might contain
-      // other left braces.
-      var searchFrom = node.span.start.offset - previousSpan.start.offset - 1;
-      if (searchFrom < 0) {
-        // This can happen when the loud comment is the first statement in a
-        // mixin that's later included. In that case, node.span.start.offset
-        // (representing the loud comment in the mixin) will be less than
-        // previousSpan.start.offset (representing the statement where the
-        // mixin is later included) ==> the loud comment cannot possibly be
-        // trailing.
-        return false;
-      }
-
-      var endOffset = previousSpan.text.lastIndexOf("{", searchFrom);
-      endOffset = math.max(0, endOffset);
-      previousSpan = previousSpan.file.span(
-          previousSpan.start.offset, previousSpan.start.offset + endOffset);
+    if (!previous.span.contains(node.span)) {
+      return node.span.start.line == previous.span.end.line;
     }
-    return node.span.start.line == previousSpan.end.line;
+
+    // Walk back from just before the current node starts looking for the
+    // parent's left brace (to open the child block). This is safer than a
+    // simple forward search of the previous.span.text as that might contain
+    // other left braces.
+    var searchFrom = node.span.start.offset - previous.span.start.offset - 1;
+    var endOffset = previous.span.text.lastIndexOf("{", searchFrom);
+    endOffset = math.max(0, endOffset);
+    var span = previous.span.file.span(
+        previous.span.start.offset, previous.span.start.offset + endOffset);
+    return node.span.start.line == span.end.line;
   }
 
   /// Writes a line feed, unless this emitting compressed CSS.
--- a/lib/src/visitor/serialize.dart
+++ b/lib/src/visitor/serialize.dart
@@ -153,14 +153,16 @@ class _SerializeVisitor
 
   void visitCssStylesheet(CssStylesheet node) {
     CssNode? previous;
-    for (var i = 0; i < node.children.length; i++) {
-      var child = node.children[i];
+    for (var child in node.children) {
       if (_isInvisible(child)) continue;
-
       if (previous != null) {
         if (_requiresSemicolon(previous)) _buffer.writeCharCode($semicolon);
-        _writeLineFeed();
-        if (previous.isGroupEnd) _writeLineFeed();
+        if (_isTrailingComment(child, previous)) {
+          _writeOptionalSpace();
+        } else {
+          _writeLineFeed();
+          if (previous.isGroupEnd) _writeLineFeed();
+        }
       }
       previous = child;
 
@@ -208,7 +210,7 @@ class _SerializeVisitor
 
     if (!node.isChildless) {
       _writeOptionalSpace();
-      _visitChildren(node.children);
+      _visitChildren(node.children, node);
     }
   }
 
@@ -226,7 +228,7 @@ class _SerializeVisitor
     });
 
     _writeOptionalSpace();
-    _visitChildren(node.children);
+    _visitChildren(node.children, node);
   }
 
   void visitCssImport(CssImport node) {
@@ -273,7 +275,7 @@ class _SerializeVisitor
         () =>
             _writeBetween(node.selector.value, _commaSeparator, _buffer.write));
     _writeOptionalSpace();
-    _visitChildren(node.children);
+    _visitChildren(node.children, node);
   }
 
   void _visitMediaQuery(CssMediaQuery query) {
@@ -298,7 +300,7 @@ class _SerializeVisitor
 
     _for(node.selector, () => node.selector.value.accept(this));
     _writeOptionalSpace();
-    _visitChildren(node.children);
+    _visitChildren(node.children, node);
   }
 
   void visitCssSupportsRule(CssSupportsRule node) {
@@ -315,7 +317,7 @@ class _SerializeVisitor
     });
 
     _writeOptionalSpace();
-    _visitChildren(node.children);
+    _visitChildren(node.children, node);
   }
 
   void visitCssDeclaration(CssDeclaration node) {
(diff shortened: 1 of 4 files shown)


lib/src/visitor/serialize.dart
```diff
@@ -1085,23 +1090,41 @@ class _SerializeVisitor
...
+      _writeOptionalSpace();
+      indentNextChild = false;
+    }
+
     CssNode previous;
-    _indent(() {
-      for (var i = 0; i < children.length; i++) {
-        var child = children[i];
-        if (_isInvisible(child)) continue;
+    for (var i = 0; i < children.length; i++) {
```
nex3:
```suggestion
    for (var child in children) {
```

I don't know why this was using an int-style for loop in the first place.

lib/src/visitor/serialize.dart
```diff
@@ -1085,23 +1090,41 @@ class _SerializeVisitor
...
+          indentNextChild = false;
         }
-        previous = child;
+        if (previous.isGroupEnd) _writeLineFeed();
+      }
+      previous = child;
 
-        child.accept(this);
+      if (indentNextChild) {
+        _indentation++;
```
nex3: I'm not sure I understand this indentation logic. It looks like if `indentNextChild` is false, it will print the child at one level of indentation lower than it would otherwise, but don't we want *zero* indentation for trailing comments?

I wonder if it makes more sense to completely special-case these comments, and write something more like:

```dart
if (_isTrailingComment(child)) {
  _writeOptionalSpace();
  _withoutIndentation(() => child.accept(this));
  continue;
}
```

For the post-bracket comments you'd want to do instead something like:

```dart
if (children.isNotEmpty && _isTrailingComment(children.first)) {
  _writeOptionalSpace();
  _withoutIndentation(() => children.first.accept(this));
  children = children.skip(1);
}
```

test/output_test.dart
```diff
@@ -59,4 +59,34 @@ void main() {
       });
     });
   });
+
+  // Regression test for sass/dart-sass#417.
+  group("preserve trailing loud comments", () {
+    // No need for "in Sass" cases as it's not possible to have
+    // trailing loud comments in the Sass syntax.
+    group("in SCSS", () {
```
nex3: I'd just fold this into the previous group name to save on some indentation.

test/output_test.dart
```diff
@@ -59,4 +59,34 @@ void main() {
       });
     });
   });
+
+  // Regression test for sass/dart-sass#417.
```
nex3: I wouldn't call this a regression test, since it's really more of a feature.

test/output_test.dart
```diff
@@ -59,4 +59,34 @@ void main() {
...
+    group("in SCSS", () {
+      test("after open block", () {
+        expect(compileString("""
+selector { /* please don't move me */
+  name: value;
+}"""), equals("""
+selector { /* please don't move me */
+  name: value;
+}"""));
+      });
```
nex3: Style nit: add empty lines between each test case.

test/output_test.dart
```diff
@@ -59,4 +59,34 @@ void main() {
...
+      test("after declaration", () {
+        expect(compileString("""
+selector {
+  name: value; /* please don't move me */
```
nex3: I'd test this both as the last declaration in a block and as a middle declaration. Same with at-rules.

test/output_test.dart
```diff
@@ -59,4 +59,34 @@ void main() {
       });
     });
   });
+
+  // Regression test for sass/dart-sass#417.
+  group("preserve trailing loud comments", () {
+    // No need for "in Sass" cases as it's not possible to have
+    // trailing loud comments in the Sass syntax.
+    group("in SCSS", () {
+      test("after open block", () {
```
nex3: Also test what happens if the comment is the only child of the block.

lib/src/visitor/serialize.dart
nickbehrens: I tried to incorporate your feedback here. PTAL.

test/output_test.dart
nickbehrens: done

test/output_test.dart
nickbehrens: yep, copy-paste from above, removed "regression"

test/output_test.dart
nickbehrens: done

test/output_test.dart
nickbehrens: done

test/output_test.dart
nickbehrens: done

nickbehrens:
> Apologies for not realizing this in advance, but I think you can actually do this without touching the parser at all. A comment is trailing iff it starts on the same line as the previous (or parent) statement, right? So you can just compare their `span.start.line` values in the serializer to tell if the comment should be treated as "trailing".

No worries.  I've reverted the parser changes and instead am now using the span line numbers to look for trailing comments.

lib/src/parse/parser.dart
```diff
@@ -79,11 +79,17 @@ class Parser {
...
-  void whitespaceWithoutComments() {
+  bool whitespaceWithoutComments() {
+    bool sawAnyNewlines = false;
     while (!scanner.isDone && isWhitespace(scanner.peekChar())) {
-      scanner.readChar();
+      if (isNewline(scanner.readChar())) {
+        sawAnyNewlines = true;
+      }
     }
+    return sawAnyNewlines;
```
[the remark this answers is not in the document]
nickbehrens: thanks for catching, reverting

lib/src/visitor/serialize.dart
```diff
@@ -1078,30 +1080,47 @@ class _SerializeVisitor
       _for(value, () => _buffer.write(value.value));
 
   /// Emits [children] in a block.
-  void _visitChildren(List<CssNode> children) {
+  void _visitChildren(Iterable<CssNode> children, CssNode parent) {
```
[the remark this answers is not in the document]
nickbehrens: ISTR there being a compiler error that wouldn't go away until I switched away from List.  But I'm not able to repro by reverting locally so I'll drop that change.

test/output_test.dart
```diff
@@ -59,4 +59,73 @@ void main() {
...
+      expect(compileString("""
+selector {
+  name: value;
+} /* please don't move me */"""), equals("""
+selector {
+  name: value;
+} /* please don't move me */"""));
+    });
+
+    test("only content in block", () {
```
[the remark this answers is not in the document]
nickbehrens: adding test case for this

lib/src/visitor/serialize.dart
```diff
@@ -1115,6 +1134,22 @@ class _SerializeVisitor
...
+  /// Whether [node] represents a trailing comment when it appears after
+  /// [previous] in a sequence of nodes.  If there are no nodes before [node],
+  /// then [previous] should be the parent node.
+  bool _isTrailingComment(CssNode node, CssNode previous) {
+    if (node is CssComment) {
+      // While the last condition (comment start line == previous end line)
+      // seems like it should be sufficient, it doesn't handle the case where
+      // "previous" is a parent node and the trailing comment appears right
+      // after a block open lbrace.
+      return node.span.start.line == previous.span.start.line
```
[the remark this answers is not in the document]
nickbehrens: adding test case for this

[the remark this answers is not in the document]
nickbehrens: I updated the PR to do as you suggested but I had trouble figuring out how to implement the beforeChildren getter in CssParentNode (see the [TODO in CssStylesheet](https://github.com/sass/dart-sass/compare/master...nickbehrens:fix-417#diff-493a0350d1b87914576cbfa5c665bb64)).

Putting the getter impl just in CssParentNode didn't seem to work for classes that merely implement (vs extend) CssParentNode (like ModifiableCssParentNode).

test/output_test.dart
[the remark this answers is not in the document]
nickbehrens: FYI the trailing comment's location is preserved but a new newline is added before the closing brace (after the trailing comment).  I added a comment about where this happens in the serializer but wasn't sure it was worth the effort to try to change existing behavior: https://github.com/sass/dart-sass/compare/master...nickbehrens:fix-417#diff-872a211685a7abf16c5f7d36c967296cR1130

lib/src/visitor/serialize.dart
```diff
@@ -1078,30 +1080,47 @@ class _SerializeVisitor
...
 
-    _writeLineFeed();
+    bool indentNextChild;
+    if (children.isNotEmpty && _isTrailingComment(children.first, parent)) {
+      _writeOptionalSpace();
+      indentNextChild = false;
+    } else {
+      _writeLineFeed();
+      indentNextChild = true;
+    }
```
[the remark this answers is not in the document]
nickbehrens: No, I don't think I forgot to push anything.  The [newer logic](https://github.com/sass/dart-sass/commit/d937f878dabf46abe490c60b3fd4e3671a6ce299#diff-872a211685a7abf16c5f7d36c967296cR1091) should (a) avoid indenting at all if no indentation is necessary (vs previously just indenting one level less) and (b) special-case trailing comments.

I did also try using your suggested code but I wasn't able to make it easily work with the rest of the existing logic in the function (esp the bits around handling semicolons and newlines that follow those).  Unfortunately it's been a little while since I tried so I don't remember all the details.

lib/src/visitor/serialize.dart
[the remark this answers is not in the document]
nickbehrens: Looking at this again a little later, I realized I don't really need a duplicate impl in CssStylesheet (as there's no beforeChildren span that makes sense there).

lib/src/visitor/serialize.dart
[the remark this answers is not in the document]
nickbehrens: I took another pass at simplifying _visitChildren and managed to get rid of the indentNextChild var.  I'm guessing this is more similar to what you were originally thinking (?).  PTAL.

test/output_test.dart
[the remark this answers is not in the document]
nickbehrens: When simplifying _visitChildren, I found a relatively easy way to "fix" this but unfortunately it breaks a bunch of tests that expect the current behavior (e.g. https://github.com/sass/dart-sass/blob/master/test/dart_api_test.dart#L25 loads in SCSS like "a { b: c; }" that lacks any newlines but then expects it to be formatted with newlines after compilation: "a {\n  b: c;\n}").

I'm thinking we should leave this as is.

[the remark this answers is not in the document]
nex3: I've pushed a change that gets rid of the trailing newline only for comments, by tracking the *two* previous nodes and checking them to see if the last node was a trailing comment of the parent node.

lib/src/visitor/serialize.dart
```diff
@@ -1078,43 +1080,59 @@ class _SerializeVisitor
...
+  /// Whether [node] represents a trailing comment when it appears after
+  /// [previous] in a sequence of nodes being serialized.  Note [previous]
+  /// could either be (1) a sibling of [node] or (2) the parent of [node], with
+  /// [node] being the first visible child.
+  bool _isTrailingComment(CssNode node, CssNode previous) {
+    if (node is! CssComment) return false;
+    FileSpan previousSpan = previous.span;
+    if (previous is CssParentNode && previous.children.contains(node)) {
+      var lbracePattern = String.fromCharCode($lbrace);
+      var endOffset = math.max(0, previousSpan.text.indexOf(lbracePattern));
```
nex3: It's an edge case, but this `indexOf()` check could fail if a selector or at-rule contains curly braces (for example, `[href*="{"]` is a valid selector), or even if it contains a comment that was removed from the AST (such as `a /*{*/ b { ... }`). It would be safer to look backwards from `node.span.start`.

test/output_test.dart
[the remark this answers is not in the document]
nickbehrens: ok, thanks!

lib/src/visitor/serialize.dart
nickbehrens: I added a test case for this, confirmed it failed, and then fixed it (by doing what you suggested).  I'll push those as soon as all the other tests finish running.

nickbehrens: committed in https://github.com/nickbehrens/dart-sass/commit/45afc0a31b5115487fa816120aff32019890d244

test/output_test.dart
```diff
@@ -145,5 +145,15 @@ selector {
...
     });
+
+    // The trailing comment detection logic looks for left braces to detect
+    // whether a comment is on the same line as its parent node.  This test
+    // checks to make sure it isn't confused by syntax that uses braces for
+    // things other than starting child blocks.
+    test("selector contains left brace", () {
+      expect(compileString("""selector[href*=\"{\"]
+{ /* please don't move me */ }"""),
+          equals("""selector[href*=\"{\"] { /* please don't move me */ }"""));
```
nex3:
```suggestion
          equals('selector[href*="{"] { /* please don't move me */ }'));
```

lib/src/visitor/serialize.dart
```diff
@@ -1138,7 +1138,13 @@ class _SerializeVisitor
 
     var previousSpan = previous.span;
     if (previous is CssParentNode && previous.children.contains(node)) {
-      var endOffset = math.max(0, previousSpan.text.indexOf("{"));
+      // Walk back from just before the current node starts looking for the
+      // parent's left brace (to open the child block).  This is safer than
+      // a simple forward search of the previousSpan.text as that might contain
+      // other left braces.
+      var endOffset =
+          previousSpan.text.lastIndexOf("{", node.span.start.offset - 1);
```
nex3: I don't think this logic is quite right... `node.span.start.offset` is an index into the entire source file, while `previousSpan.text`, just covers the span itself. I think you'll need to access [`node.span.file.getText()`](https://pub.dev/documentation/source_span/latest/source_span/SourceFile/getText.html).

nickbehrens: ah, I thought it was an offset relative to the beginning of the region covered by the span.  But I see now the FileLocation APIs are all absolute (which makes sense).  

I made the test case a little richer to try to catch the issue and then updated isTrailingComment to convert the absolute offset to relative to previousSpan's start offset: https://github.com/nickbehrens/dart-sass/commit/35f51df40734af3653e8f049d1915cbf61a32704

(I didn't see a need to use node.span.file.getText() but maybe you had something else in mind?)

nickbehrens:
I looked at the crash and reduced the repro to the following:

```
@mixin loudComment {
  /* ... */
}

selector {
  @include loudComment;
}
```
In short, `@include` breaks the assumption of the current `_isTrailingComment` logic that `node.span.start.offset` will be strictly after `prevNode.span.start.offset`.

I _think_ we just need to check for `searchFrom < 0` and if so, it's never a trailing comment.  I'll commit that change (with a unit test) and see what happens with the spec tests.

nickbehrens:
After the most recent SassSpec test run, there are 4 remaining issues.  2 of them I think are reasonable formatting changes (i.e. we could update SassSpec test cases to match):

1) [SassSpec::Test#test__spec/libsass-closed-issues/issue_894](https://github.com/sass/sass-spec/blob/master/spec/libsass-closed-issues/issue_894.hrx)
```
--- expected
+++ actual
-"a { /**/
-}
+"a { /**/ }
```

2) [SassSpec::Test#test__spec/css/keyframes/bubble/empty](https://github.com/sass/sass-spec/blob/master/spec/css/keyframes.hrx)
```
--- expected
+++ actual
-"@keyframes { /**/
-}
+"@keyframes { /**/ }
```

One of them involves a standalone loud comment following a block end (rbrace) that's now getting treated as a trailing comment.  I can add a test case for this and investigate.

3) [SassSpec::Test#test__spec/libsass-closed-issues/issue_1007](https://github.com/sass/sass-spec/blob/master/spec/libsass-closed-issues/issue_1007.hrx)
```
--- expected
+++ actual
 foo baz {
   /* before */
   margin: 0; /* after */
-}
-/* end */
+} /* end */
```

The last one I could use some help parsing/understanding:

4) [SassSpec::Test#test__spec/directives/use/css/order/import_order/comments_and_imports](https://github.com/sass/sass-spec/blob/master/spec/directives/use/css/order.hrx) 
```
--- expected
+++ actual
 "/* upstream comment before import */
-@import \"upstream.css\"; /* midstream comment before use */
+@import \"upstream.css\";
+/* midstream comment before use */
 /* midstream comment before first import */
 @import \"midstream1.css\";
 /* midstream comment before second import */
-@import \"midstream2.css\"; /* input comment before use */
+@import \"midstream2.css\";
+/* input comment before use */
 /* input comment before import */
 @import \"input.css\";
 /* upstream comment after import */
```

e.g. the expected output seems to have `@import \"upstream.css\"; /* midstream comment before use */` all on one line but I don't see that in the `.hrx` file (?)

test/output_test.dart
```diff
@@ -162,5 +162,16 @@ selector[href*=\"{\"] { /* please don't move me */ }
 
 @rule3;"""));
     });
+
+    test("loud comments in mixin", () {
```
nex3: This case raises a couple other possible edge cases:

* What if the loud comment comes after the style rule but still isn't contained within it, for example when the style rule is also in a mixin?
* What if the loud comment is in a totally different file from the mixin?

I'd probably solve these by writing a `bool spanContainsLocation(SourceSpan span, SourceLocation location)` helper and checking that in the outer conditional.

Goodwine:
So taking back this conversation to what Nick spotted I saw the following problems:

1. The formatting issues spotted from before were solved. (e.g. `@keyframes { /**/ }`) 🆗 

2. Unsure what the correct format is for loud comments after open and close braces, e. g.
   ```scss
   foo { /* open */
     margin: none;
   } /* close */
   ```
   currently this becomes ... I assume this is fine?
   ```css
   foo {
     /* open */
     margin: none;
   }
   /* close */
   ```
   but there are very odd cases with nesting 😵‍💫 

3. I'm confused why the loud comments from comments and imports should become trailing comments instead of staying where they are originally 🤔

lib/src/visitor/serialize.dart
```diff
@@ -1287,44 +1289,88 @@ class _SerializeVisitor
...
+    if (node is! CssComment) return false;
+
+    var previousSpan = previous.span;
+    if (previous is CssParentNode && previous.children.contains(node)) {
+      // Walk back from just before the current node starts looking for the
+      // parent's left brace (to open the child block).  This is safer than
+      // a simple forward search of the previousSpan.text as that might contain
+      // other left braces.
+      var searchFrom = node.span.start.offset - previousSpan.start.offset - 1;
+      if (searchFrom < 0) {
```
nex3, line 1356: This isn't necessarily the most reliable determiner of whether the math on the span offsets is nonsense... it's possible that `node.span` appears late in a different file than `previousSpan`, for example. It's probably better to write a little `SpanExtensions.contains()` utility and check that `previous.span.contains(node.span)`. You could even use that instead of the `previous.children.contains(node)` check.

test/output_test.dart
```diff
@@ -110,4 +110,210 @@ void main() {
...
+selector[href*="{"]
+{ /* please don't move me */ }
+
+@rule3;"""), equals("""@rule1;
+@rule2;
+selector[href*="{"] { /* please don't move me */ }
+
+@rule3;"""));
+    });
+    group("loud comments in mixin", () {
```
nex3, line 216: Nit: missing newline above

test/output_test.dart
```diff
@@ -110,4 +110,210 @@ void main() {
...
+@mixin loudComment {
+  /* ... */
+}
+
+selector {
+  @include loudComment;
+}"""), "selector {\n  /* ... */\n}");
+      });
+
+      test("with stylerule", () {
```
nex3, line 228: Nit

```suggestion
      test("with style rule", () {
```

lib/src/visitor/serialize.dart
Goodwine: WIP

lib/src/visitor/serialize.dart
```diff
@@ -1287,44 +1289,88 @@ class _SerializeVisitor
...
-      return;
-    }
 
-    _writeLineFeed();
-    CssNode? previous_;
-    _indent(() {
-      for (var i = 0; i < children.length; i++) {
-        var child = children[i];
-        if (_isInvisible(child)) continue;
+    CssNode? prePrevious_;
```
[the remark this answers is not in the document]
Goodwine: nvm, the previous code does have the underscore, but I removed it anyways

Goodwine (commented):
one comment still WIP

lib/src/visitor/serialize.dart
Goodwine: IIUC then I think this is done :)

nex3 (approved):

PR #855, opened 2019-10-18
Reset the forwarded config to empty for @use and meta.load-module()

### LINKED ISSUE #854: Reusing a module with @use and arguments
I have modules that I call in different files.
I use mixins from modules.
When I want to pass arguments to different files using the same module using the 'with' operator, I get errors.

> Error: This module was already loaded, so it can't be configured using "with".
>   ╷
> 1 │ @use "../promises/transition";

I have a file whose load a file from the 'promises' folder: _transition.scss.

Code:
```
// src/sass/controls/_button.scss
@use "../promises/transition";

// some code
```
```
// src/sass/controls/_input.scss
@use "../promises/transition";

// some code
```

The website.scss file contains code:
```
@use "package-name/src/sass/controls/button"  with (
  $font-family: $font-family
);
@use "package-name/src/sass/controls/input";
// some code
```

It's work. But this code dont' work:
```
@use "package/src/sass/controls/button" with (
  $font-family: $font-family
);
@use "one/src/sass/controls/input" with (
  $font-family: $font-family
);
```

These modules use the same module. 
Such calls are quite common in the code, but I can not use them because of these errors.

nex3:
In #827 I made configuration never be null, but this turns out to be
overkill: we still need a way for _loadModule() and _execute() to
distinguish between "use the existing configuration" (for @forward)
and "use no configuration" (for an unconfigured @use or
meta.load-module()). We now use null as a sentinel value there, while
still ensuring that _configuration is non-nullable.

Closes #854
See sass/sass-spec#1486

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -413,9 +413,9 @@ class _EvaluateVisitor
         var url = Uri.parse(arguments[0].assertString("module").text);
         var withMap = arguments[1].realNull?.assertMap("with")?.contents;
 
-        Map<String, _ConfiguredValue> configuration;
+        var configuration = const <String, _ConfiguredValue>{};
         if (withMap != null) {
-          configuration = <String, _ConfiguredValue>{};
+          configuration = {};
           var span = _callableNode.span;
           withMap.forEach((variable, value) {
             var name =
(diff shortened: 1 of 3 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -525,11 +525,9 @@ class _EvaluateVisitor
       {Uri baseUrl,
       Map<String, _ConfiguredValue> configuration,
       bool namesInErrors = false}) async {
-    configuration ??= const {};
-
     var builtInModule = _builtInModules[url];
     if (builtInModule != null) {
-      if (configuration.isNotEmpty) {
+      if (configuration != null && configuration.isNotEmpty) {
```
jathak: Would the better style here and in line 661 be `configuration?.isNotEmpty ?? false`?

nex3: I considered that, but I think the null logic is easier to follow here than trying to follow a bunch of `?`s. It's close, though.

PR #861, opened 2019-10-23
Fix broken links in perf.md

howardjing:


perf.md
```diff
@@ -165,7 +165,7 @@ overall.
 
 Dart Sass on Node is still substantially slower than on the Dart VM, and that
 relative slowdown becomes more pronounced as the raw Dart code becomes faster.
-Solutions for this such as [the embedded protocol][] or [WebAssembly support][]
+Solutions for this such as [the embedded protocol][embedded Dart Sass] or [WebAssembly support][Dart WebAssembly support]
```
nex3: Rather than changing the references here, can you just update the text referenced below to refer to the text that's actually being linked so we can continue to use short-style links?

howardjing: Done.

PR #862, opened 2019-10-25
Don't compile .css files in directories to themselves

### LINKED ISSUE #853: --watch infinite loop
hi,
i update my sass version to "1.23.0 compiled with dart2js 2.5.1" recently and my command for watch my files as un unexpected behavior.

the problem come from this : [Fix a bug where changes in .css files would be ignored in --watch mode.](https://github.com/sass/dart-sass/blob/master/CHANGELOG.md#command-line-interface)

The problem happend when the input folder and the outpout folder are the same.
When my files *.scss are compiled that create files like *.css then sass detect these new files and watch it so it's creating an infinite loop.

My command is : sass --watch .:.

my solutions are : 
- create a command with all my files (exemple : sass --watch style.scss:style.css colors.scss:color.css anything.scss:anything.css ...) but it's not efficient
- outpout my files in a diferent folder (exemple : sass --watch stylesheets:public/stylesheets) but it's changing my workflow

the changes that can be made :
- Adding the suport of globs, this is already an issue : #616
- Adding the possibility to ignore patterns (exemple : sass --watch --ignore *.css .:.)

Thank you for your time !
Sorry for my english.

nex3:
This also adds support for compiling .css files in directories *at
all*, which had previously only worked in --watch mode.

Closes #853

### DIFF
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -375,7 +375,7 @@ class ExecutableOptions {
       for (var path in listDir(source, recursive: true))
         if (_isEntrypoint(path) &&
             // Don't compile a CSS file to its own location.
-            (source != destination || p.extension(path) != '.css'))
+            !(source == destination && p.extension(path) == '.css'))
           path: p.join(destination,
               p.setExtension(p.relative(path, from: source), '.css'))
     };
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -373,7 +373,9 @@ class ExecutableOptions {
   Map<String, String> _listSourceDirectory(String source, String destination) {
     return {
       for (var path in listDir(source, recursive: true))
-        if (_isEntrypoint(path))
+        if (_isEntrypoint(path) &&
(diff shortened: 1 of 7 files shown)


lib/src/executable/options.dart
```diff
@@ -373,7 +373,9 @@ class ExecutableOptions {
   Map<String, String> _listSourceDirectory(String source, String destination) {
     return {
       for (var path in listDir(source, recursive: true))
-        if (_isEntrypoint(path))
+        if (_isEntrypoint(path) &&
+            // Don't compile a CSS file to its own location.
+            (source != destination || p.extension(path) != '.css'))
```
Awjin: Total nit: might be clearer deMorgans'ed:

!(source == destination && p.extension(path) == '.css')

nex3: Done.

jakeparis:
What is the advantage of compiling a .css file?

nex3:
Some people use Sass as a way of minifying their plain CSS.

PR #877, opened 2019-11-07
Avoid depending on `iterator.current != null`.

lrhn:
With the NNBD change to Dart, it's no longer safe to rely on an iterator returning `null` when it has hit the end (or before calling `moveNext` the first time). For non-nullable element types, it will have to throw instead.

This PR rewrites code that currently rely on a `null` value to recognize the end of an iterator.

### DIFF
--- a/lib/src/extend/functions.dart
+++ b/lib/src/extend/functions.dart
@@ -522,15 +522,19 @@ List<List<T>> paths<T>(Iterable<List<T>> choices) => choices.fold(
 QueueList<List<ComplexSelectorComponent>> _groupSelectors(
     Iterable<ComplexSelectorComponent> complex) {
   var groups = QueueList<List<ComplexSelectorComponent>>();
-  var iterator = complex.iterator..moveNext();
-  while (iterator.current != null) {
-    var group = <ComplexSelectorComponent>[];
-    do {
-      group.add(iterator.current);
-    } while (iterator.moveNext() &&
-        (iterator.current is Combinator || group.last is Combinator));
+  var iterator = complex.iterator;
+  if (iterator.moveNext()) {
+    var group = <ComplexSelectorComponent>[iterator.current]   
     groups.add(group);
-  }
+    while (iterator.moveNext()) {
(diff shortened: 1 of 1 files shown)


lib/src/extend/functions.dart
```diff
@@ -522,15 +522,19 @@ List<List<T>> paths<T>(Iterable<List<T>> choices) => choices.fold(
...
   var groups = QueueList<List<ComplexSelectorComponent>>();
-  var iterator = complex.iterator..moveNext();
-  while (iterator.current != null) {
-    var group = <ComplexSelectorComponent>[];
-    do {
-      group.add(iterator.current);
-    } while (iterator.moveNext() &&
-        (iterator.current is Combinator || group.last is Combinator));
+  var iterator = complex.iterator;
+  if (iterator.moveNext()) {
```
nex3:
```suggestion
  if (!iterator.moveNext()) return groups;
```

Might as well save some indentation.

nex3:
@lrhn Are you planning to continue working on this?

lrhn:
Ack, yes. I admit I had forgotten this PR existed.

lib/src/extend/functions.dart
lrhn: Done.

lib/src/extend/functions.dart
```diff
@@ -522,16 +522,18 @@ List<List<T>> paths<T>(Iterable<List<T>> choices) => choices.fold(
...
       group.add(iterator.current);
-    } while (iterator.moveNext() &&
-        (iterator.current is Combinator || group.last is Combinator));
-    groups.add(group);
+    } else {
+      group = [iterator.current];
+      groups.add(group);
+    }
   }
-  return groups;
```
nex3: It looks like you lost this `return`.

lrhn: Well, doh. Caught red-handed using the Github editor :)

nex3:
Thanks for the fix!

PR #878, opened 2019-11-09
Print call-site for @errors in mixins and functions

### LINKED ISSUE #474: Print the invocation site for @errors in mixins and functions
Currently, if an `@error` is encountered in a mixin or a function, the source span that's associated with that error (and thus the snippet that's printed in error messages) is the `@error` rule itself. This isn't super useful; `@error` is usually used to validate *calls* to an API, so if it fails it should show the call site instead.

The stack trace shouldn't be affected.

For example:

```scss
@mixin validate-arg($arg) {
  @if $arg < 10 {
    @error "Arg #{$arg} should be >= 10.";
  }
}

.foo {
  @include validate-arg(9);
}
```

Currently prints:

```
Error: "Arg 9 should be >= 10."
    @error "Arg #{$arg} should be >= 10.";
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  test.scss 3:5  validate-arg()
  test.scss 8:3  root stylesheet
```

but it should print:

```
Error: "Arg 9 should be >= 10."
    @include validate-arg(9);
    ^^^^^^^^^^^^^^^^^^^^^^^^
  test.scss 3:5  validate-arg()
  test.scss 8:3  root stylesheet
```

Awjin:
For any @errors encountered in mixins or functions, use the call-site
(instead of the @error rule) as the span printed in the error message.

Closes #474
sass/sass-spec#1494

### DIFF
--- a/lib/src/exception.dart
+++ b/lib/src/exception.dart
@@ -81,14 +81,8 @@ body::before {
 class SassRuntimeException extends SassException {
   final Trace trace;
 
-  SassRuntimeExceptionType type;
-
-  SassRuntimeException(String message, FileSpan span, this.trace, [this.type])
+  SassRuntimeException(String message, FileSpan span, this.trace)
       : super(message, span);
-}
-
-enum SassRuntimeExceptionType {
-  error,
 }
 
 /// An exception thrown when Sass parsing has failed.
--- a/lib/src/exception.dart
+++ b/lib/src/exception.dart
@@ -81,8 +81,14 @@ body::before {
 class SassRuntimeException extends SassException {
   final Trace trace;
 
-  SassRuntimeException(String message, FileSpan span, this.trace)
+  SassRuntimeExceptionType type;
+
+  SassRuntimeException(String message, FileSpan span, this.trace, [this.type])
       : super(message, span);
+}
+
+enum SassRuntimeExceptionType {
+  error,
 }
 
 /// An exception thrown when Sass parsing has failed.
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -1071,8 +1071,8 @@ class _EvaluateVisitor
   }
 
   Future<Value> visitErrorRule(ErrorRule node) async {
-    throw _exception(
-        (await node.expression.accept(this)).toString(), node.span);
+    throw _exception((await node.expression.accept(this).toString()), node.span,
+        SassRuntimeExceptionType.error);
   }
 
   Future<Value> visitExtendRule(ExtendRule node) async {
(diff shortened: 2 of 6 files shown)


lib/src/exception.dart
```diff
@@ -81,10 +81,16 @@ body::before {
...
   final Trace trace;
 
-  SassRuntimeException(String message, FileSpan span, this.trace)
+  SassRuntimeExceptionType type;
+
+  SassRuntimeException(String message, FileSpan span, this.trace, [this.type])
       : super(message, span);
 }
 
+enum SassRuntimeExceptionType {
```
nex3: Unintuitively, it's generally *not* good practice to use Dart's native enum type. It's quite inflexible once you have to do anything beyond the absolute basics, so rather than rewriting enums entirely once that comes up we generally manually write pseudo-enums from the beginning. See [`ListSeparator`](https://github.com/sass/dart-sass/blob/master/lib/src/value/list.dart#L59-L85) for an example of the manual enum pattern.

I think we may be able to get away with not using an enum here, though. It's only being used to detect errors thrown by `@error`, but I think you can determine that by checking `error.span.text.startsWith("@error")`.

lib/src/visitor/async_evaluate.dart
```diff
@@ -1469,7 +1469,14 @@ class _EvaluateVisitor
...
             for (var statement in mixin.declaration.children) {
-              await statement.accept(this);
+              try {
+                await statement.accept(this);
+              } on SassRuntimeException catch (error) {
+                if (error.type != SassRuntimeExceptionType.error) rethrow;
+
+                throw SassRuntimeException(
+                    error.message, node.span, _stackTrace(error.span));
+              }
```
nex3: Since this `try`/`catch` is repeated twice, consider putting it in a helper method, similar to the existing `_addExceptionSpanAsync` helper.

test/cli/shared/errors.dart
```diff
@@ -72,6 +72,58 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
...
+    var sass = await runSass(["--no-unicode", "test.scss"]);
+    expect(
+        sass.stderr,
+        emitsInOrder([
+          "Error: \"Within A.\"",
+          "  ,",
+          "6 |   c: a();",
+          "  |      ^^^",
+          "  '",
+          "  test.scss 2:3  a()",
```
nex3: I know this is what my example in the original issue looked like, but looking at it closer I think it's confusing that the top line of the stack trace is pointing to a different place than the span. Can you make it regenerate the stack trace at the same time it regenerates the span itself?

lib/src/visitor/async_evaluate.dart
```diff
@@ -2853,6 +2854,31 @@ class _EvaluateVisitor
...
+  /// source span.
+  T _addErrorSpan<T>(AstNode nodeWithSpan, T callback()) {
+    try {
+      return callback();
+    } on SassRuntimeException catch (error) {
+      if (!error.span.text.startsWith("@error")) rethrow;
+      throw SassRuntimeException(
+          error.message, nodeWithSpan.span, _stackTrace());
+    }
+  }
```
nex3: I wouldn't add this until it's actually used (or rather, I'd rename `_addErrorSpanAsync()` to `_addErrorSpan()`).

Awjin: Done.

PR #883, opened 2019-11-14
Fix arity in overloaded function argument error

### LINKED ISSUE #520: Misleading error message for overloaded function arity
When the wrong number of arguments are passed to a function with overloads, Dart Sass's error message sometimes claims that the function accepts fewer arguments than it actually does. For example:

```
$ sass --interactive
>> hsla(1, 2, 3, 4, 5)
   ^^^^^^^^^^^^^^^^^^^
Error: Only 1 argument allowed, but 5 were passed.
```

`hsla()` allows up to 4 arguments, so this error message is confusing and incorrect.

Awjin:
When overloaded functions receive an incorrect number of positional
arguments, determine which overload has the most similar number of
arguments, and then correctly display that number in the error.

Closes #520
sass/sass-spec#1496

### DIFF
--- a/lib/src/callable/built_in.dart
+++ b/lib/src/callable/built_in.dart
@@ -63,13 +63,13 @@ class BuiltInCallable implements Callable, AsyncBuiltInCallable {
     Tuple2<ArgumentDeclaration, _Callback> fuzzyMatch;
     int minMismatchDistance;
 
-    for (var i = 0; i < _overloads.length; i++) {
+    for (var overload in _overloads) {
       var overload = _overloads[i];
 
       // Ideally, find an exact match.
       if (overload.item1.matches(positional, names)) return overload;
 
-      int mismatchDistance = overload.item1.arguments.length - positional;
+      var mismatchDistance = overload.item1.arguments.length - positional;
 
       if (minMismatchDistance != null) {
         if (mismatchDistance.abs() > minMismatchDistance.abs()) continue;
--- a/lib/src/callable/built_in.dart
+++ b/lib/src/callable/built_in.dart
@@ -55,13 +55,36 @@ class BuiltInCallable implements Callable, AsyncBuiltInCallable {
   /// Returns the argument declaration and Dart callback for the given
   /// positional and named arguments.
   ///
-  /// Note that this doesn't guarantee that [positional] and [names] are valid
-  /// for the returned [ArgumentDeclaration].
+  /// If no exact match is found, finds the closest approximation. Note that this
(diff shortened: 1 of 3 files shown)


lib/src/callable/built_in.dart
```diff
@@ -55,13 +55,36 @@ class BuiltInCallable implements Callable, AsyncBuiltInCallable {
...
   Tuple2<ArgumentDeclaration, _Callback> callbackFor(
-          int positional, Set<String> names) =>
-      _overloads.take(_overloads.length - 1).firstWhere(
-          (overload) => overload.item1.matches(positional, names),
-          orElse: () => _overloads.last);
+      int positional, Set<String> names) {
+    Tuple2<ArgumentDeclaration, _Callback> fuzzyMatch;
+    int minMismatchDistance;
+
+    for (var i = 0; i < _overloads.length; i++) {
```
nex3:
```suggestion
    for (var overload in _overloads) {
```

lib/src/callable/built_in.dart
```diff
@@ -55,13 +55,36 @@ class BuiltInCallable implements Callable, AsyncBuiltInCallable {
...
+      var overload = _overloads[i];
+
+      // Ideally, find an exact match.
+      if (overload.item1.matches(positional, names)) return overload;
+
+      int mismatchDistance = overload.item1.arguments.length - positional;
```
nex3:
```suggestion
      var mismatchDistance = overload.item1.arguments.length - positional;
```

You don't need to type-annotate variables that can be inferred.

lib/src/callable/built_in.dart
Awjin: Thanks

lib/src/callable/built_in.dart
Awjin: Gotcha

lib/src/callable/built_in.dart
```diff
@@ -55,13 +55,34 @@ class BuiltInCallable implements Callable, AsyncBuiltInCallable {
...
+      if (minMismatchDistance != null) {
+        if (mismatchDistance.abs() > minMismatchDistance.abs()) continue;
+        // If two overloads have the same mismatch distance, favor the overload
+        // that has more arguments.
+        if ((mismatchDistance.abs() == minMismatchDistance.abs()) &&
+            (mismatchDistance < 0)) continue;
```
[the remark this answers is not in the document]
Awjin: Will fix

PR #885, opened 2019-11-19
Support configuring modules through imports

### LINKED ISSUE #882: Support configuring modules with @import
See https://github.com/sass/sass/issues/2772

jathak:
Fixes #882

See sass/sass-spec#1497

### DIFF
--- a/lib/src/async_environment.dart
+++ b/lib/src/async_environment.dart
@@ -13,6 +13,7 @@ import 'ast/css.dart';
 import 'ast/node.dart';
 import 'ast/sass.dart';
 import 'callable.dart';
+import 'configuration.dart';
 import 'exception.dart';
 import 'extend/extender.dart';
 import 'module.dart';
@@ -728,6 +729,22 @@ class AsyncEnvironment {
       }
       _nestedForwardedModules?.removeLast();
     }
+  }
+
+  /// Creates an implicit configuration from the variables declared in this
+  /// environment.
+  Configuration toImplicitConfiguration() {
+    var configuration = <String, ConfiguredValue>{};
+    for (var i = 0; i < _variables.length; i++) {
+      var values = _variables[i];
+      var nodes =
+          _variableNodes == null ? <String, AstNode>{} : _variableNodes[i];
+      for (var name in values.keys) {
+        configuration[name] =
+            ConfiguredValue(values[name], nodes[name]?.span, nodes[name]);
+      }
+    }
+    return Configuration(configuration, isImplicit: true);
   }
 
   /// Returns a module that represents the top-level members defined in [this],
--- a/lib/src/async_environment.dart
+++ b/lib/src/async_environment.dart
@@ -14,6 +14,7 @@ import 'ast/node.dart';
 import 'ast/sass.dart';
 import 'callable.dart';
 import 'configuration.dart';
+import 'configured_value.dart';
 import 'exception.dart';
 import 'extend/extender.dart';
 import 'module.dart';
@@ -740,8 +741,9 @@ class AsyncEnvironment {
       var nodes =
           _variableNodes == null ? <String, AstNode>{} : _variableNodes[i];
       for (var name in values.keys) {
-        configuration[name] =
-            ConfiguredValue(values[name], nodes[name]?.span, nodes[name]);
+        // Implicit configurations are never invalid, making [configurationSpan]
+        // unnecessary, so we pass null here to avoid having to compute it.
+        configuration[name] = ConfiguredValue(values[name], null, nodes[name]);
       }
     }
     return Configuration(configuration, isImplicit: true);
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -11,15 +11,6 @@ import 'util/unprefixed_map_view.dart';
 
 /// A set of variables meant to configure a module by overriding its
 /// `!default` declarations.
-///
-/// There are two types of configuration: an explicit one that's created by
-/// the `with` clause of a `@use` rule and an implicit one that's created any
-/// time a file containing a `@forward` rule is imported.
-///
-/// Both types of configuration pass through `@forward` rules, but explicit
-/// configurations will cause an error if attempting to use them on a module
-/// that has already been loaded, while implicit configurations will be silently
-/// ignored in this case.
 class Configuration {
   /// A map from variable names (without `$`) to values.
   ///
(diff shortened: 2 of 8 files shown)


jathak:
I think that makes sense, but could there ever be a case where the `Environment` is mutated after the `Configuration` is created but before it's used?

nex3:
Good question... I guess that does run into the problem where variables defined in the midstream file suddenly become part of the upstream file's configuration, which we don't want, so copying everything over is probably better in the end.

lib/src/async_environment.dart
```diff
@@ -670,6 +671,22 @@ class AsyncEnvironment {
...
+  /// environment.
+  Configuration toImplicitConfiguration() {
+    var configuration = <String, ConfiguredValue>{};
+    for (var i = 0; i < _variables.length; i++) {
+      var values = _variables[i];
+      var nodes =
+          _variableNodes == null ? <String, AstNode>{} : _variableNodes[i];
+      for (var name in values.keys) {
+        configuration[name] =
+            ConfiguredValue(values[name], nodes[name]?.span, nodes[name]);
```
nex3: Materializing `AstNode`s' spans can be expensive, which is why we pass them around as `AstNode`s in the first place, so I'd like to avoid `nodes[name]?.span` here. In practice this span is also never used, since an implicit configuration's variables are never invalid.

We could just pass `null` here, or we could encode this more statically by having a subclass with a constructor `ImplicitConfiguredValue(Value, AstNode)` and defining `Configuration.implicit(Map<String, ImplicitConfiguredValue>)`. Up to you.

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+///
+/// Both types of configuration pass through `@forward` rules, but explicit
+/// configurations will cause an error if attempting to use them on a module
+/// that has already been loaded, while implicit configurations will be silently
+/// ignored in this case.
+class Configuration {
+  /// A map from variable names (without `$`) to values.
+  ///
+  /// When this is empty, it may be unmodifiable, so [Map.remove] should not be
+  /// called on this.
```
nex3: Now that we have a full `Configuration` class, we can enforce this more aggressively by only exposing an unmodifiable view of this map and exposing a `Configuration.remove()` method that handles empty configurations correctly.

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+  final Map<String, ConfiguredValue> values;
+
+  /// Whether or not this configuration is implicit.
+  final bool isImplicit;
```
nex3: I might move the discussion of what an implicit configuration means from the class documentation to this getter.

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+  Configuration(this.values, {this.isImplicit = false});
+
+  /// The empty configuration, which indicates that the module has not been
+  /// configured.
+  const Configuration.empty()
+      : values = const {},
+        isImplicit = false;
+
+  bool get isEmpty => values.isEmpty;
+  bool get isNotEmpty => values.isNotEmpty;
```
nex3: Typically classes that aren't implementing core library interfaces don't define `isNotEmpty`. (Defining `isNotEmpty` at all was a design mistake—having a different way to negate exactly one boolean getter doesn't make sense—but there's no going back on that for the core libraries at this point.)

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+
+  /// Creates a new configuration from this one based on a `@forward` rule.
+  Configuration throughForward(ForwardRule forward) {
```
nex3: Moving this logic here is nice.

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+      newValues = LimitedMapView.whitelist(newValues, forward.shownVariables);
+    } else if (forward.hiddenVariables?.isNotEmpty ?? false) {
+      newValues = LimitedMapView.blacklist(newValues, forward.hiddenVariables);
+    }
+    return Configuration(newValues, isImplicit: isImplicit);
+  }
+
+  /// Creates a copy of this configuration.
+  Configuration clone() => isEmpty
+      ? const Configuration.empty()
```
nex3: This will return an explicit configuration even if `this` is implicit. Does that matter? Maybe all empty configurations should be considered implicit, since they don't cause duplicate-configuration errors?

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+      : Configuration({...values}, isImplicit: isImplicit);
+}
+
+/// A variable value that's been configured using `@use ... with`.
+class ConfiguredValue {
```
nex3: I usually put every class definition in its own file, even if two classes are related, unless one is private or they have to access private members from one another.

lib/src/configuration.dart
```diff
@@ -0,0 +1,85 @@
...
+    return Configuration(newValues, isImplicit: isImplicit);
+  }
+
+  /// Creates a copy of this configuration.
+  Configuration clone() => isEmpty
+      ? const Configuration.empty()
+      : Configuration({...values}, isImplicit: isImplicit);
+}
+
+/// A variable value that's been configured using `@use ... with`.
```
nex3: Some values are configured without `@use ... with`—maybe just say "configured for a `Configuration`"?

lib/src/visitor/async_evaluate.dart
```diff
@@ -576,22 +574,21 @@ class _EvaluateVisitor
...
   Future<Module> _execute(AsyncImporter importer, Stylesheet stylesheet,
-      {Map<String, _ConfiguredValue> configuration,
-      bool namesInErrors = false}) async {
+      {Configuration configuration, bool namesInErrors = false}) async {
     var url = stylesheet.span.sourceUrl;
 
     var alreadyLoaded = _modules[url];
     if (alreadyLoaded != null) {
-      if ((configuration ?? _configuration).isNotEmpty) {
+      configuration ??= _configuration;
```
nex3: I don't think you want to make this assignment. We use the fact that this was passed as null later on to determine whether to throw errors about unused configuration variables.

lib/src/visitor/async_evaluate.dart
```diff
@@ -635,8 +632,7 @@ class _EvaluateVisitor
       _inKeyframes = false;
 
       if (configuration != null) {
-        _configuration =
-            configuration.isEmpty ? const {} : Map.of(configuration);
+        _configuration = configuration.clone();
```
nex3: Style nit: you can make this a single-line `if` now.

lib/src/visitor/async_evaluate.dart
```diff
@@ -661,11 +657,11 @@ class _EvaluateVisitor
       if (configuration != null && _configuration.isNotEmpty) {
```
nex3: Shouldn't this also check whether the configuration is implicit?

lib/src/visitor/async_evaluate.dart
```diff
@@ -1306,13 +1285,20 @@ class _EvaluateVisitor
...
         _importer = importer;
         _stylesheet = stylesheet;
         _root = ModifiableCssStylesheet(stylesheet.span);
         _parent = _root;
         _endOfImports = 0;
         _outOfOrderImports = null;
 
+        // This configuration is only used if it passes through a `@forward`
+        // rule, so we avoid creating unnecessary ones for performance reasons.
+        if (stylesheet.forwards.isNotEmpty) {
```
nex3: Before you submit, try removing this condition and *always* creating the environment, and run the result against sass-spec. It shouldn't change the behavior, so if it does you'll have found a bug.

jathak:
Okay, I think I've addressed all comments. I also tested with the condition for creating the implicit configuration removed and confirmed that all of the specs still pass.

PR #906, opened 2019-12-20
Adds built-in clamp() and hypot()

Awjin:
See sass/sass-spec#1500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## 1.25.0
 
 * Add functions to the built-in "sass:math" module.
+  * `clamp()`: given a `$min`, $number`, and `$max` values, clamps the `$number`
+    in between `$min` and `$max`.
+  * `hypot()`: given *n* numbers, outputs the length of the *n*-dimensional
+    vector that has components equal to each of the inputs.
 
 ## 1.24.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.25.0
+
+* Add functions to the built-in "sass:math" module.
+
 ## 1.24.0
 
 * Add an optional `with` clause to the `@forward` rule. This works like the
--- a/lib/src/functions/math.dart
+++ b/lib/src/functions/math.dart
@@ -56,12 +56,14 @@ final _clamp = BuiltInCallable("clamp", r"$min, $number, $max", (arguments) {
     return number;
   }
 
-  var arg1 = "$min";
-  var arg2 = "${min.hasUnits != number.hasUnits ? number : max}";
-  var unit1 = "${min.hasUnits ? "has units" : "is unitless"}";
-  var unit2 = "${min.hasUnits ? "is unitless" : "has units"}";
+  var arg2 = min.hasUnits != number.hasUnits ? number : max;
+  var arg2Name = min.hasUnits != number.hasUnits ? "\$number" : "\$max";
+  var unit1 = min.hasUnits ? "has unit ${min.unitString}" : "is unitless";
+  var unit2 = arg2.hasUnits ? "has unit ${arg2.unitString}" : "is unitless";
+
   throw SassScriptException(
-      "$arg1 $unit1 but $arg2 $unit2. Arguments must all have units or all be unitless.");
+      "\$min $unit1 but $arg2Name $unit2. Arguments must all have units or " +
+          "all be unitless.");
 });
 
 final _floor = _numberFunction("floor", (value) => value.floor());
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.25.0
+
+* Add functions to the built-in "sass:math" module.
```
nex3: List the functions here, and provide brief descriptions of them.

lib/src/functions/math.dart
```diff
@@ -21,14 +22,16 @@ final global = UnmodifiableListView([
   _round, _ceil, _floor, _abs, _max, _min, _randomFunction, _unit, //
   _percentage,
   _isUnitless.withName("unitless"),
-  _compatible.withName("comparable")
+  _compatible.withName("comparable"),
+  _div.withName("divide")
```
nex3: As with the spec CL, I'd leave the addition of the division function to later.

lib/src/functions/math.dart
```diff
@@ -21,14 +22,16 @@ final global = UnmodifiableListView([
...
 ]);
 
 /// The Sass math module.
 final module = BuiltInModule("math", functions: [
   _round, _ceil, _floor, _abs, _max, _min, _randomFunction, _unit,
   _isUnitless, //
-  _percentage, _compatible
+  _percentage, _compatible,
+  _clamp, _div, _hypot
```
nex3: I don't know why the formatting got weird here before, but this should be flowed onto as few lines as possible:

```dart
final module = BuiltInModule("math", functions: [
  _round, _ceil, _floor, _abs, _max, _min, _randomFunction, _unit, //
  _isUnitless, _percentage, _compatible, _clamp, _div, _hypot
]);
```

lib/src/functions/math.dart
```diff
@@ -98,3 +101,57 @@ BuiltInCallable _numberFunction(String name, num transform(num value)) {
         denominatorUnits: number.denominatorUnits);
   });
 }
+
+final _clamp = BuiltInCallable("clamp", r"$min, $preferred, $max", (arguments) {
```
nex3: `$preferred` is a different parameter name than what the spec says (`$number`).

lib/src/functions/math.dart
```diff
@@ -98,3 +101,57 @@ BuiltInCallable _numberFunction(String name, num transform(num value)) {
...
+  var min = arguments[0].assertNumber("min");
+  var preferred = arguments[1].assertNumber("preferred");
+  var max = arguments[2].assertNumber("max");
+
+  if (min.hasUnits != preferred.hasUnits ||
+      preferred.hasUnits != max.hasUnits) {
+    throw SassScriptException(
+        "Mixing unitless arguments with arguments that have units is not allowed.");
```
nex3: It would be good to provide more detail here. Which numbers are in conflict? What are their values? The user may be confused about how they got into this state, and this information could help them figure it out.

lib/src/parse/stylesheet.dart
```diff
@@ -2367,6 +2367,9 @@ relase. For details, see http://bit.ly/moz-document.
     number += _tryDecimal(allowTrailingDot: scanner.position != start.position);
     number *= _tryExponent();
 
+    // Preserve the sign of -0 by representing it as a double.
+    number = (sign.isNegative && number == 0) ? -0.0 : sign * number;
```
nex3: Is this necessary? I don't mind having `-0` parse to a normal `0`.

Awjin: We technically need it to parse inputs like `math.div(1, -0)`

nex3: Could we generate an in-memory `-0` value by doing weird stuff with math instead?

lib/src/functions/math.dart
Awjin: Hmm good question. In this context it could make sense to have the majority "win out"—so if `arg1` and `arg2` are unitless but `arg3` has units, the error could be `Error: Expected $arg3 to have no units`.

This error might be confusing for a call to `hypot()` with lots of args (e.g. 4 unitless and 5 unitful) but that seems like the uncommon case.

Awjin: Actually, thinking about it more, if the user reads `Error: Expected $arg3 to have no units`, in the future they might assume that `clamp` only takes unitless values.

Awjin: How about—`Error: $arg1 has units but $arg2 does not. Mixing unitless arguments with arguments that have units is not allowed.` It's wordier than most messages we currently have, but it feels pretty clear.

lib/src/parse/stylesheet.dart
Awjin: Not sure if I'm following. Don't we lose all information about `-0` if we parse it as `0` here, causing `math.div` to be invoked on the wrong input?

CHANGELOG.md
Awjin: Done.

lib/src/functions/math.dart
Awjin: Done.

lib/src/functions/math.dart
Awjin: Done. I also decided to reorganize this file, let me know your thoughts.

lib/src/functions/math.dart
Awjin: Good catch, thanks.

lib/src/functions/math.dart
```diff
@@ -13,34 +13,58 @@ import '../module/built_in.dart';
...
+    if (number.greaterThanOrEquals(max).isTruthy) return max;
+    return number;
+  }
+
+  var arg1 = "$min";
+  var arg2 = "${min.hasUnits != number.hasUnits ? number : max}";
+  var unit1 = "${min.hasUnits ? "has units" : "is unitless"}";
+  var unit2 = "${min.hasUnits ? "is unitless" : "has units"}";
+  throw SassScriptException(
+      "$arg1 $unit1 but $arg2 $unit2. Arguments must all have units or all be unitless.");
```
nex3: I'd also include the names of the arguments in question here—it may not be obvious from context.

Also, this is a long line (the formatter won't break strings for you).

Same below.

lib/src/parse/stylesheet.dart
nex3: You're right that we don't generate the floating-point value `-0.0`, but I don't think that's inherently a problem—you can still get that value through arithmetic (such as `-1 * 0.0`). Unless there's a compelling case for a user to be able to write a terse negative zero literal, I think preserving backwards compatibility is more important.

lib/src/functions/math.dart
nex3: LGTM

lib/src/parse/stylesheet.dart
Awjin: Ah, I'm understanding now. Agreed.

lib/src/functions/math.dart
Awjin: Done.

PR #907, opened 2019-12-26
Add built-in variables e and pi.

Awjin:
See https://github.com/sass/sass-spec/pull/1501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   * `hypot()`: given *n* numbers, outputs the length of the *n*-dimensional
     vector that has components equal to each of the inputs.
 
-* Add the variables `pi` and `e` to the built-in "sass:math" module.
+* Add the variables `$pi` and `$e` to the built-in "sass:math" module.
 
 ## 1.24.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * `hypot()`: given *n* numbers, outputs the length of the *n*-dimensional
     vector that has components equal to each of the inputs.
 
+* Add the variables `pi` and `e` to the built-in "sass:math" module.
+
 ## 1.24.0
 
 * Add an optional `with` clause to the `@forward` rule. This works like the
--- a/lib/src/module/built_in.dart
+++ b/lib/src/module/built_in.dart
@@ -31,7 +31,9 @@ class BuiltInModule<T extends AsyncCallable> implements Module<T> {
       : url = Uri(scheme: "sass", path: name),
         functions = _callableMap(functions),
         mixins = _callableMap(mixins),
-        variables = _variableMap(variables);
+        variables = variables == null
+            ? UnmodifiableMapView({})
+            : UnmodifiableMapView(variables);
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
   * `hypot()`: given *n* numbers, outputs the length of the *n*-dimensional
     vector that has components equal to each of the inputs.
 
+* Add the variables `pi` and `e` to the built-in "sass:math" module.
```
nex3:
```suggestion
* Add the variables `$pi` and `$e` to the built-in "sass:math" module.
```

lib/src/module/built_in.dart
```diff
@@ -39,8 +41,16 @@ class BuiltInModule<T extends AsyncCallable> implements Module<T> {
           : UnmodifiableMapView(
               {for (var callable in callables) callable.name: callable}));
 
+  /// Returns a map from [variables]' names to their values.
+  static Map<String, V> _variableMap<V>(Map<String, V> variables) =>
```
nex3: Since this logic isn't re-used, I don't think you need to factor it out into a function.

lib/src/module/built_in.dart
```diff
@@ -39,8 +41,16 @@ class BuiltInModule<T extends AsyncCallable> implements Module<T> {
...
+      UnmodifiableMapView(
+          variables == null ? {} : UnmodifiableMapView(variables));
+
   void setVariable(String name, Value value, AstNode nodeWithSpan) {
-    throw SassScriptException("Undefined variable.");
+    if (!variables.containsKey(name)) {
+      throw SassScriptException("Undefined variable.");
+    }
+    throw SassScriptException("Cannot modify built-in variable \$$name.");
```
nex3: You don't need to include the variable name here, because it'll always be present in the error's source span (that's why we don't mention it above).

Awjin: Makes sense, fixed.

lib/src/module/built_in.dart
Awjin: Done.

CHANGELOG.md
Awjin: Done.

lib/src/module/built_in.dart
```diff
@@ -17,19 +17,23 @@ class BuiltInModule<T extends AsyncCallable> implements Module<T> {
...
 
-  BuiltInModule(String name, {Iterable<T> functions, Iterable<T> mixins})
+  BuiltInModule(String name,
+      {Iterable<T> functions, Iterable<T> mixins, Map<String, Value> variables})
       : url = Uri(scheme: "sass", path: name),
         functions = _callableMap(functions),
-        mixins = _callableMap(mixins);
+        mixins = _callableMap(mixins),
+        variables = variables == null
+            ? UnmodifiableMapView({})
```
nex3:
```suggestion
            ? const {}
```

A constant map is implicitly unmodifiable, and it'll save some allocations.

Awjin: Thanks. Javascript habits are hard to shake ;)

PR #908, opened 2019-12-26
Cache separate canonical URLs for @use and @import

### LINKED ISSUE #899: Incorrect behavior when a stylesheet in a load path is loaded by both `@use` and `@import`
For example:
```scss
// app.scss
@use "library";
@import "library";

// load-path/_library.scss
a { b: regular; }

// load-path/_library.import.scss
a { b: import-only; }
```
should output the regular styles followed by the import-only styles (and it does so if the rules' URLs are relative), but instead it outputs the regular styles twice.

I believe this is due to `ImportCache` using the same canonicalization cache for both `@import` and `@use` rules (I actually ran into this with the migrator initially).

jathak:
Fixes #899.

The cache for canonical URLs is now keyed on *both* the rule URL and
whether that URL was canonicalized for an `@import` rule.

### DIFF
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -11,6 +11,7 @@ import 'package:tuple/tuple.dart';
 import 'ast/sass.dart';
 import 'importer.dart';
 import 'importer/result.dart';
+import 'importer/utils.dart';
 import 'io.dart';
 import 'logger.dart';
 import 'sync_package_resolver.dart';
@@ -26,11 +27,15 @@ class AsyncImportCache {
 
   /// The canonicalized URLs for each non-canonical URL.
   ///
+  /// The second item in each key's tuple is true when this canonicalization is
+  /// for an `@import` rule. Otherwise, it's for a `@use` or `@forward` rule.
+  ///
   /// This map's values are the same as the return value of [canonicalize].
   ///
   /// This cache isn't used for relative imports, because they're
   /// context-dependent.
-  final Map<Uri, Tuple3<AsyncImporter, Uri, Uri>> _canonicalizeCache;
+  final Map<Tuple2<Uri, bool>, Tuple3<AsyncImporter, Uri, Uri>>
+      _canonicalizeCache;
 
   /// The parsed stylesheets for each canonicalized import URL.
   final Map<Uri, Stylesheet> _importCache;
@@ -109,18 +114,20 @@ class AsyncImportCache {
   /// If any importers understand [url], returns that importer as well as the
   /// canonicalized URL. Otherwise, returns `null`.
   Future<Tuple3<AsyncImporter, Uri, Uri>> canonicalize(Uri url,
-      [AsyncImporter baseImporter, Uri baseUrl]) async {
+      {AsyncImporter baseImporter, Uri baseUrl, bool forImport = false}) async {
     if (baseImporter != null) {
       var resolvedUrl = baseUrl != null ? baseUrl.resolveUri(url) : url;
-      var canonicalUrl = await _canonicalize(baseImporter, resolvedUrl);
+      var canonicalUrl =
+          await _canonicalize(baseImporter, resolvedUrl, forImport);
       if (canonicalUrl != null) {
         return Tuple3(baseImporter, canonicalUrl, resolvedUrl);
       }
     }
 
-    return await putIfAbsentAsync(_canonicalizeCache, url, () async {
+    return await putIfAbsentAsync(_canonicalizeCache, Tuple2(url, forImport),
+        () async {
       for (var importer in _importers) {
-        var canonicalUrl = await _canonicalize(importer, url);
+        var canonicalUrl = await _canonicalize(importer, url, forImport);
         if (canonicalUrl != null) {
           return Tuple3(importer, canonicalUrl, url);
         }
@@ -132,8 +139,14 @@ class AsyncImportCache {
 
   /// Calls [importer.canonicalize] and prints a deprecation warning if it
   /// returns a relative URL.
-  Future<Uri> _canonicalize(AsyncImporter importer, Uri url) async {
-    var result = await importer.canonicalize(url);
+  Future<Uri> _canonicalize(
+      AsyncImporter importer, Uri url, bool forImport) async {
+    Uri result;
+    if (forImport) {
+      result = await inImportRule(() => importer.canonicalize(url));
+    } else {
+      result = await importer.canonicalize(url);
+    }
     if (result?.scheme == '') {
       _logger.warn("""
 Importer $importer canonicalized $url to $result.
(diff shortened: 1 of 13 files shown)


lib/src/async_import_cache.dart
```diff
@@ -132,8 +139,14 @@ class AsyncImportCache {
 
   /// Calls [importer.canonicalize] and prints a deprecation warning if it
   /// returns a relative URL.
-  Future<Uri> _canonicalize(AsyncImporter importer, Uri url) async {
-    var result = await importer.canonicalize(url);
+  Future<Uri> _canonicalize(
+      AsyncImporter importer, Uri url, bool forImport) async {
+    Uri result;
+    if (forImport) {
```
nex3: Style nit: I'd use `? :` here to save a few lines of boilerplate.

lib/src/importer/utils.dart
```diff
@@ -6,43 +6,42 @@ import 'package:path/path.dart' as p;
...
-    _inUseRule = wasInUseRule;
+    _inImportRule = wasInImportRule;
   }
 }
 
 /// Resolves an imported path using the same logic as the filesystem importer.
 ///
-/// This tries to fill in extensions and partial prefixes and check if a directory default. If no file can be
-/// found, it returns `null`.
+/// This tries to fill in extensions and partial prefixes and check if a
```
nex3:
```suggestion
/// This tries to fill in extensions and partial prefixes and check for a
```

lib/src/stylesheet_graph.dart
```diff
@@ -92,14 +93,17 @@ class StylesheetGraph {
 
   /// Returns a map from non-canonicalized imported URLs in [stylesheet], which
```
nex3: Document the new map value.

lib/src/stylesheet_graph.dart
```diff
@@ -237,8 +242,9 @@ class StylesheetNode {
   /// stylesheets those imports refer to.
   ///
   /// This may have `null` values, which indicate failed imports.
-  Map<Uri, StylesheetNode> get upstream => UnmodifiableMapView(_upstream);
-  Map<Uri, StylesheetNode> _upstream;
+  Map<Tuple2<Uri, bool>, StylesheetNode> get upstream =>
```
nex3: It's plausible that we'll want to expose this class publicly relatively soon, and I really don't want to have to break this API once we remove imports. WDYT about making `upstream` refer only to `@use` and `@import`, and have a separate `upstreamImports` field that only contains imports?

jathak: Yeah, that makes sense. I went back and forth on whether to key on a tuple or just have two maps while I was writing this. Would it make sense to do the same in `ImportCache` for consistency?

nex3: Since `ImportCache`'s map isn't part of its public API, I think it's fine to have it stay as tuple-keyed.

lib/src/executable/watch.dart
```diff
@@ -298,15 +299,22 @@ class _Watcher {
...
+            // If the call to canonicalize failed, do nothing. We'll surface
+            // the error more nicely when we try to recompile the file.
           }
-          importChanged = newCanonicalUrl != node.upstream[url]?.canonicalUrl;
+          importChanged = newCanonicalUrl != node?.canonicalUrl;
         }
       }
+
+      node.upstream.forEach(recanonicalize);
+      forImport = true;
```
nex3: Rather than using this as implicit state, I'd just pass it to `recanonicalize()`. It does mean you'll have to expand those `forEach()` calls, but I think the clarity of flow is worth it.

lib/src/stylesheet_graph.dart
```diff
@@ -39,7 +40,10 @@ class StylesheetGraph {
     DateTime transitiveModificationTime(StylesheetNode node) {
       return _transitiveModificationTimes.putIfAbsent(node.canonicalUrl, () {
         var latest = node.importer.modificationTime(node.canonicalUrl);
-        for (var upstream in node.upstream.values) {
+        for (var upstream in {
+          ...node.upstream.values,
+          ...node.upstreamImports.values
```
nex3: I'd use [`Iterable.followedBy()`](https://api.dartlang.org/stable/2.7.0/dart-core/Iterable/followedBy.html) here to avoid copying everything into a set eagerly. There's a possibility we'll visit a file twice, but that seems unlikely and it won't affect the result.

Similar below.

lib/src/stylesheet_graph.dart
nex3: It looks like this doc comment still needs to be updated.

lib/src/visitor/find_dependencies.dart
```diff
@@ -2,21 +2,28 @@
...
-    _FindImportsVisitor().run(stylesheet);
-
-/// A visitor that traverses a stylesheet and records all the [DynamicImport]s
-/// it contains.
-class _FindImportsVisitor extends RecursiveStatementVisitor<void> {
+/// Returns two lists of dependencies for [stylesheet].
+///
+/// The first is a list of URLs from all `@use` and `@forward` rules in
+/// [stylesheet]. The second is a list of all imports in [stylesheet].
+Tuple2<List<Uri>, List<DynamicImport>> findDependencies(
```
nex3: Why aren't imports represented as `Uri`s as well?

lib/src/stylesheet_graph.dart
```diff
@@ -233,29 +241,45 @@ class StylesheetNode {
   /// The canonical URL of [stylesheet].
   final Uri canonicalUrl;
 
-  /// A map from non-canonicalized import URLs in [stylesheet] to the
-  /// stylesheets those imports refer to.
+  /// A map from non-canonicalized `@use` and `@forward` URLs in [stylesheet] to
+  /// the stylesheets those rules refer to.
   ///
   /// This may have `null` values, which indicate failed imports.
```
nex3:
```suggestion
  /// This may have `null` values, which indicate failed loads.
```

jathak:
@nex3: Can you restart the AppVeyor build? I don't think I have permission

nex3:
Done.

lib/src/executable/watch.dart
```diff
@@ -288,8 +288,8 @@ class _Watcher {
     var changed = <StylesheetNode>[];
     for (var node in _graph.nodes.values) {
       var importChanged = false;
-      for (var url in node.upstream.keys) {
-        if (_name(p.url.basename(url.path)) != name) continue;
+      void recanonicalize(Uri url, StylesheetNode upstream, bool forImport) {
```
nex3: Nit:

```suggestion
      void recanonicalize(Uri url, StylesheetNode upstream, {@required bool forImport}) {
```

This just helps make it clearer at the callsite what the boolean means.

PR #910, opened 2019-12-28
Update chokidar to v3

paulmillr:
See https://paulmillr.com/posts/chokidar-3-save-32tb-of-traffic/ - it's like 10x better. Chokidar v2 will keep getting worse slowly, because it's kinda unsupported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "name": "sass",
   "devDependencies": {
     "node-sass": "^4.11.0",
-    "chokidar": "^2.0.0",
+    "chokidar": "^3.3.0",
     "fibers": ">=1.0.0 <5.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -7,7 +7,7 @@
   "name": "sass",
   "devDependencies": {
     "node-sass": "^4.11.0",
-    "chokidar": "^2.0.0",
+    "chokidar": "^3.3.0",
```
nex3: I'd like this to match the distributed `package.json`'s constraint, which is `>=2.0.0 <4.0.0`.

paulmillr:
@nex3 whatever -- I won't sign the "CLA" anyway, it's the piece of shit. Maintainers would need to do this on their own.

PR #917, opened 2020-01-02
Replace whitelist/blacklist with safelist/blocklist

nex3:
Inspired by https://twitter.com/amlyhamm/status/1202684742069604353.


nex3:
Yeah, it wasn't something that had occurred to me either until I saw that tweet!

PR #919, opened 2020-01-03
Fix import-only files for Node importers

jathak:


test/node_api/importer_test.dart
```diff
@@ -165,6 +165,32 @@ void main() {
...
+      await writeTextFile(
+          p.join(sandbox, 'target.import.scss'), 'a {b: import-only}');
+
+      expect(
+          renderSync(RenderOptions(
+              data: "@use 'foo'; @import 'foo';",
+              importer: allowInterop((void _, void __) =>
+                  NodeImporterResult(file: p.join(sandbox, 'target.scss'))))),
+          equalsIgnoringWhitespace('a { b: regular; } a { b: import-only; }'));
+    });
```
nex3: I'd also add a test to the top-level `node_api_test` that the default import logic supports both of these cases.

PR #921, opened 2020-01-07
Add trigonometric math functions.

Awjin:
See https://github.com/sass/sass-spec/pull/1504

### DIFF
--- a/lib/src/functions/math.dart
+++ b/lib/src/functions/math.dart
@@ -140,6 +140,7 @@ final _hypot = BuiltInCallable("hypot", r"$numbers...", (arguments) {
 
 final _log = BuiltInCallable("log", r"$number, $base: null", (arguments) {
   var number = arguments[0].assertNumber("number");
+
   if (number.hasUnits) {
     throw SassScriptException("\$number: Expected $number to have no units.");
   }
@@ -149,6 +150,7 @@ final _log = BuiltInCallable("log", r"$number, $base: null", (arguments) {
   if (arguments[1] == sassNull) return SassNumber(math.log(numberValue));
 
   var base = arguments[1].assertNumber("base");
+
   if (base.hasUnits) {
     throw SassScriptException("\$base: Expected $base to have no units.");
   }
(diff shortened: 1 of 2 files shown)


lib/src/functions/math.dart
```diff
@@ -213,6 +214,115 @@ final _sqrt = BuiltInCallable("sqrt", r"$number", (arguments) {
...
+  var acos = math.acos(numberValue) * 180 / math.pi;
+  return SassNumber.withUnits(acos, numeratorUnits: ['deg']);
+});
+
+final _asin = BuiltInCallable("asin", r"$number", (arguments) {
+  var number = arguments[0].assertNumber("number");
+  number.assertNoUnits();
+  var numberValue = number.value;
+
+  if (fuzzyEquals(numberValue, 0) && numberValue.isNegative) {
```
nex3: Why don't we round a positive number that fuzzy-equals 0?

Awjin: Good question. The spec only specifies this behavior for `-0.0`. But it would probably be more consistent if the spec also handled `0`.

Awjin: The same would apply to `sqrt()`, which currently only handles `-0.0`.

lib/src/functions/math.dart
```diff
@@ -140,6 +140,7 @@ final _hypot = BuiltInCallable("hypot", r"$numbers...", (arguments) {
 
 final _log = BuiltInCallable("log", r"$number, $base: null", (arguments) {
   var number = arguments[0].assertNumber("number");
+
```
nex3: Style nit: I don't think these newlines necessarily improve readability; having every statement separated by an empty line doesn't really provide any more information than having no empty lines would.

Generally speaking, I think clustering a variable assignment in the same vertical block as a validity check for that variable, or that uses that variable, is good.

lib/src/functions/math.dart
```diff
@@ -207,11 +209,18 @@ final _pow = BuiltInCallable("pow", r"$base, $exponent", (arguments) {
...
   if (number.hasUnits) {
     throw SassScriptException("\$number: Expected $number to have no units.");
   }
 
-  return SassNumber(math.sqrt(number.value));
+  var numberValue = number.value;
+
+  if (fuzzyEquals(numberValue, 0)) {
+    numberValue = numberValue.isNegative ? -0.0 : 0;
+  }
```
nex3: Consider factoring this logic out into a helper function.

PR #926, opened 2020-01-11
Throw multi-span exceptions

nex3:
This shouldn't land until after dart-lang/source_span#49 is released.

See sass/sass-spec#1505

PR #932, opened 2020-01-15
NodeJS `run_` API method should return native JavaScript `Promise`

devversion:
Currently the `run_` method, exposed through the `dart.sass.js` file,
returns a `Future`. The `Future` object is not accessible due to the
Dart compilation process and `Future` not being a thing in the
JavaScript/NodeJS platform.

In order to provide a better interface to the Sass API for
JavaScript/NodeJS consumers, a native `Promise` is now returned.

Fixes https://github.com/bazelbuild/rules_sass/issues/96.

### DIFF
--- a/lib/src/node.dart
+++ b/lib/src/node.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:js';
 import 'dart:js_util';
 import 'dart:typed_data';
 
@@ -21,6 +22,7 @@ import 'importer/node.dart';
 import 'node/error.dart';
 import 'node/exports.dart';
 import 'node/function.dart';
+import 'node/promise.dart';
 import 'node/render_context.dart';
 import 'node/render_context_options.dart';
 import 'node/render_options.dart';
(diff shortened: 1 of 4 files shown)


devversion:
I tried figuring out if the test failures are due to changes in this PR, but it doesn't look like it. I sent an empty test PR to confirm this: https://github.com/sass/dart-sass/pull/933

jathak:
It looks like tests are failing because the CIs have started to use a version of the `source_span` package with slightly different output from our current specs. This should be resolved once #926 is merged.

nex3:
We're in the process of transitioning the way we build JS executables to use `cli_pkg` (#924), so it's probably a good idea to update this there rather than in this package.

jelbourn:
Would someone from the Sass team be able to follow up on that? The Angular folks can only go so deep into the dart ecosystem before we're over our heads

nex3:
I can, but it'll have to be after we do this migration. Can you file an issue on http://github.com/google/dart_cli_pkg?

devversion:
Just came across this again. I've created the issue on the `dart_cli_pkg` and also experimented a bit with it, but looks like it will be time consuming to set up the project/tests on my end: google/dart_cli_pkg#50

Issue #868, opened 2019-10-30
Use worker threads to implement render()


nex3:
Currently, we implement the asynchronous `render()` method in Node.js in one of two ways:

* By default, it's run through the `AsyncEvaluator` which is synchronized via code gen with the synchronous `Evaluator`. This is *much* slower than running synchronously, and imposes developer overhead by requiring synchronization.

* If the user passes in a reference to the [`fibers` package](https://www.npmjs.com/package/fibers) (which uses a C++ extension to add a coroutine-style asynchrony model to Node), we run the synchronous code path and use coroutines to call out to async helpers. This is as fast as running synchronously but requires the end user to opt in to a C++ compilation which may be difficult to compile.

Sass's business logic itself is fully synchronous, so we get no benefit from this asynchrony other than the ability to invoke asynchronous function and importer callbacks. I think we can avoid the downsides of both of these methods by running the whole Sass compiler in a [worker thread](https://nodejs.org/api/worker_threads.html) and using [`Atomics.wait()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait) to block that thread until the callback completes.

hegelocampus:
I am thinking about taking on the implementation of this and I had a few quick clarification questions:
- Is this still a desired enhancement?
- To be explicitly clear, this would involve the porting of the needed components of the node.js `worker_threads` library, right?
- Am I correct that the check for `fibers` along with the call to `renderAsync` would all be removed in favor of `renderSync`?
- Then in `src/node.dart` a worker thread would be setup to deal with the rendering within `_renderSync`?
Sorry I've got all of these clarification questions, I don't want to write a bunch of code only to realize I completely misinterpreted what you were asking for here.

nex3:
> First of all, is this still a desired enhancement?

Yes, definitely!

> To be explicitly clear, this would involve the porting of the needed components of the node.js `worker_threads` library, right?

You wouldn't need to port anything. Dart compiled to JS can invoke JS libraries, so you'd just need to call this library.

> Am I correct that the check for `fibers` along with the call to `renderAsync` would all be removed in favor of `renderSync`?

We'd definitely be getting rid of `fibers`. I think we want to retain a distinction between `renderAsync()` and `renderSync()` both for backwards compatibility, and so we can avoid the overhead of spinning up a worker thread when only synchronous callbacks are necessary.

[...]

PR #939, opened 2020-01-22
WIP: Use worker threads in render

hegelocampus:
This isn't quite in a working state but I think it's at least 75% of the way there right now. 

I've hit a bit of a snag with using the `js.dart` package and I'm pretty sure I'm doing something very wrong in [`lib/src/node/worker_threads.dart`](https://github.com/sass/dart-sass/compare/master...hegelocampus:use-worker-threads-to-implement-render?expand=1#diff-186d5a1200b74a315242676c5ca915a2) but I can't quite figure out exactly what.   
The present issue is that essentially that nothing is actually being imported by node. My first approach was very similar to what is suggested in the api docs, but nothing was defined. So then I tried to implement the same JS `require` interop that I found being done in `test/node_api/intercept_stdout.dart`, but I'm not sure if my translation of that implementation makes sense either. 
  
If someone could point me in the right direction as to how to use the `js.dart` package to import something that is not globally accessible, I would really appreciate it because I feel like I've my lost my footing and don't really know how one would/should go about importing `worker_threads` for use in Dart.

Todo:
- [ ] Fix `lib/src/node/worker_threads.dart`
- [ ] Clean up worker thread implementation in `lib/src/node.js`
- [ ] Check that all of tests that are no longer relevant have been removed
- [ ] Do performance tests to see benefits attained through the use of a worker thread
- [ ] Possibly more...

Implements #868


nex3 (commented):
I'm sorry you have to face the nightmare that is Dart's JS interop... it's confusing, incomplete, and very poorly documented. I'll try to walk you through the necessary steps.

First, you don't need to worry about manually simulating a `require()`—there's an easier way. In the JS compilation tooling, [we add a few requires](https://github.com/sass/dart-sass/blob/master/tool/grind/npm.dart#L68-L70) to the beginning of the JS output file. You can add a `worker_thread` require there as well, and then in the JS interop `worker_thread` will just be a pre-defined name.

The reason your existing code isn't working is because, when you call `require()` through JS interop, the object being returned isn't globally available. As far as Dart's concerned, it's just some object somewhere in the program, and it doesn't know that it matches the `worker_threads` module name that you declared for the rest of the library.

ronkorving:
Hi, very interested party here :) Is this still going places?

hegelocampus:
Hi, @ronkorving, I had a major change in the amount of free time I have and I definitely don't have time to complete this so it's all yours if you want it. I pushed all of my local changes to this branch and you can feel free to use those as a starting place. Good luck and godspeed.

ronkorving:
@hegelocampus First off, I appreciate your work so far. But I need to be very clear about my own situation... I am not in a situation right now or anytime soon where I can make this kind of contribution for a host of reasons. I hope the SASS community can help out, as this seems to be very important work (especially since fibers now officially printed a warning of obsolescence).

PR #945, opened 2020-02-06
Fix a number of defects related to imported environments

nex3:


nex3:
Good idea; I've updated a95c5b8ccba88e4d07311d6650389ed9f9b79837 to describe the bug in detail.

PR #952, opened 2020-02-10
Eagerly initialize Environment._globalModules

nex3:
We had been lazily initializing this to be more efficient when no
global modules were used, but this meant that the object wasn't shared
with closures created for mixins and functions that were created when
it was still `null`, so later imported forwards weren't visible to
those members.

sass/sass-spec#1513

### DIFF
--- a/lib/src/async_environment.dart
+++ b/lib/src/async_environment.dart
@@ -40,15 +40,11 @@ class AsyncEnvironment {
   final Map<String, AstNode> _namespaceNodes;
 
   /// The namespaceless modules used in the current scope.
-  ///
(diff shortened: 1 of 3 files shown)


lib/src/async_environment.dart
```diff
@@ -40,15 +40,11 @@ class AsyncEnvironment {
   final Map<String, AstNode> _namespaceNodes;
 
   /// The namespaceless modules used in the current scope.
-  ///
-  /// This is `null` if there are no namespaceless modules.
-  Set<Module> _globalModules;
+  final Set<Module> _globalModules;
```
nshahan: Cool, more final!

PR #962, opened 2020-02-25
Fix a --watch bug

nex3:


Awjin:
Wow, this was an elusive bug.

nex3:
Yeah, it's taken over a year to finally nail it down.

Issue #904, opened 2019-12-19
The fix for issue #806 leads to an infinity loop!


WolverinDEV:
By adding the `.css` file ending to the "allow compile" list creates an infinity loop when the output file is within the source directory.

Steps to reproduce:
1. Create a file called `a.scss` with this content:
```
body { }
```
2. Execute the watch command: `sass --watch .`
3. Edit the file.

If you're lucky it will only compile the css file twice.
If you're not you might end up within an infinity loop.

nex3:
Duplicate of #853. I think you're using an outdated Sass version.

WolverinDEV:
Updated to `1.23.7` issue persist.

nex3:
I can't reproduce this locally, and we have tests in place that verify that this exact problem doesn't reoccur. I suspect you're somehow still using an old Sass version.

GenesisTMS:
I can still reproduce this issue

- `npm install sass`
- create file `a.scss` and write to it

`sass --watch`
```
Compiled a.scss to a.css.
Compiled a.css to css.
```

`sass --version`
```
1.25.0 compiled with dart2js 2.7.0
```

Working on fedora

nex3:
@GenesisTMS I can't reproduce this.

georgpukk:
On Windows 10

`sass --version` [dart-sass-1.26.0-windows-x64.zip](https://github.com/sass/dart-sass/releases/tag/1.26.0)
```
1.26.0
```

`sass ./:./ --watch` _when saving creates an infinite loop_
```
Compiled main.scss to main.css.
Compiled main.css to main.css.
Compiled main.css to main.css.
Compiled main.css to main.css.
and so on...
```
`sass . --watch` _works fine_
```
Compiled main.scss to main.css.
```

georgpukk:
Updated to `1.26.1` and `sass . --watch` now too creates an infinite loop.

WolverinDEV:
Yeahr, good to see that I'm not the only one.
I suppose to may add a switch to ignore css files. This will quickly solve this.

georgpukk:
@nex3 This reproduces it on Windows 10
```
test("goes to infinite loop", () async {
      await d.file("test.scss", "a {b: c}").create();

      var sass = await watch(["."]);
      await expectLater(
          sass.stdout, emits('Compiled test.scss to test.css.'));
      await expectLater(sass.stdout, _watchingForChanges);
      await tickIfPoll();

[...]

PR #968, opened 2020-03-01
Fix a check that prevents compiling .css files to themselves

georgpukk:
* fixes the infinite loop issue in --watch mode  (#904, #853)
* previously it compared relative destination to absolute source
* p.relative(destination) is used to replace backslashes

### DIFF
--- a/lib/src/executable/watch.dart
+++ b/lib/src/executable/watch.dart
@@ -267,7 +267,7 @@ class _Watcher {
           p.setExtension(p.relative(source, from: sourceDir), '.css'));
 
       // Don't compile ".css" files to their own locations.
-      if (destination != source) return destination;
+      if (p.relative(destination) != p.relative(source)) return destination;
     }
 
     return null;
--- a/lib/src/executable/watch.dart
+++ b/lib/src/executable/watch.dart
@@ -267,7 +267,7 @@ class _Watcher {
           p.setExtension(p.relative(source, from: sourceDir), '.css'));
 
       // Don't compile ".css" files to their own locations.
-      if (p.relative(destination) != p.relative(source)) return destination;
+      if (!p.equals(destination, source)) return destination;
     }
 
     return null;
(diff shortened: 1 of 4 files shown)


crexalbo:
Thank you so much for this commit! I just learned SASS and this bug is the only thing preventing me from fully integrating SASS into my workflow. For now I'm just running sass from the command line every time I make an edit which is super tedious, but with this I can just set it as a daemon and forget it.

lib/src/executable/watch.dart
```diff
@@ -267,7 +267,7 @@ class _Watcher {
           p.setExtension(p.relative(source, from: sourceDir), '.css'));
 
       // Don't compile ".css" files to their own locations.
-      if (destination != source) return destination;
+      if (p.relative(destination) != p.relative(source)) return destination;
```
nex3:
```suggestion
      if (!p.equals(destination, source)) return destination;
```

[`p.equals()`](https://pub.dev/documentation/path/latest/path/equals.html) is the best way to determine whether two paths actually refer to the same location.

georgpukk:
This is the reproducing test case (at least on Windows 10).
```
test("goes to infinite loop", () async {
	await d.file("test.scss", "a {b: c}").create();

	var sass = await watch(["."]);
	await expectLater(
		sass.stdout, emits('Compiled test.scss to test.css.'));
	await expectLater(sass.stdout, _watchingForChanges);
	await tickIfPoll();

	await d.file("test.scss", "r {o: g}").create();
	await expectLater(
		sass.stdout, emits('Compiled test.scss to test.css.'));

	await tickIfPoll();

	await d.file("test.scss", "x {y: z}").create();
	await expectLater(
		sass.stdout, emits('Compiled test.scss to test.css.'));

	await sass.kill();

	await d
		.file("test.css", equalsIgnoringWhitespace("x { y: z; }"))
		.validate();
});
```

nex3:
Can you add that to the existing watch tests?

PR #979, opened 2020-04-14
Remove two unused constructors named '_none'

srawlins:


nex3:
Looks like you need to re-run `pub run grinder`.

srawlins:
Thank you! I think my `pub run grinder` picked up some new dartfmt changes; let me know if those need to be reverted.

PR #990, opened 2020-04-24
Add missing constants to JS API

### LINKED ISSUE #977: [node-sass incompatibility] Missing constants: sass.NULL, sass.TRUE and sass.FALSE
dart-sass doesn't support the node-sass constants `sass.NULL`, `sass.TRUE` and `sass.FALSE` as convenience accessors to `sass.types.Null.NULL`, `sass.types.Boolean.TRUE`, and `sass.types.Boolean.FALSE` respectively.

jathak:
Fixes #977.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.26.6
+
+### JavaScript API
+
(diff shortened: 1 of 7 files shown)


jathak:
Done

Issue #975, opened 2020-03-28
[node-sass incompatibility] Importers that return both file and contents keys.


chriseppstein:
In `node-sass` if both `contents` and `file` are returned, the presence of `contents` wins and the file is not attempted to be read, instead the file name provided is used as that file's canonical url.

In `dart-sass` the presence of `file` causes the importer result to attempt to read the file from disk even if the contents are already present.

The relevant code in dart-sass is here:
https://github.com/sass/dart-sass/blob/master/lib/src/importer/node/implementation.dart#L192

The behavior of `node-sass` is preferable here because it allows importers to return the correct path to the file while also having already processed the results. For instance, the importer might do some preprocessing on the file to get it ready for parsing by Sass, but also wants to return the full url to the resolved location for error reporting purposes.

PR #991, opened 2020-04-25
Don't read `file` if Node importer returns `file` and `contents`.

Awjin:
Instead, use `contents` with `file` as the canonical url.

Closes https://github.com/sass/dart-sass/issues/975.


Awjin:
@nex3 For context, this is equivalent to the behavior implemented in libsass: https://github.com/sass/libsass/blob/master/src/context.cpp#L448

Awjin:
> Looks good! Can you add a changelog entry and bump the pubspec if necessary as well?

Sure thing, slotted this under 1.26.6

PR #1005, opened 2020-05-12
Fix a selector parsing bug in indented syntax

### LINKED ISSUE #1002: Wrong parsing of valid SASS syntax
When I try to parse the following SASS file I get a strange warning
```
WARNING on line 1, column 1 of normalize.sass:
This selector doesn't have any properties and won't be rendered.
1 │ ┌ button,
2 │ └ input, /* 1 */
```
However, if I remove the comment from line 2 - the warning goes away.
Also, if I move the comma after the comment - the warning goes away.
If just does not make sense to me.
```sass
button,
input, /* 1 */
select /* 2, 3 */
  overflow: visible
  text-transform: none
```

jathak:
Fixes #1002.

In the indented syntax, a selector list may continue onto another line if the previous line ends with a comma. Previously, if there was a comment after the comma, it wouldn't be recognized and the selector would be broken in two (with the first selector having no properties).

This fixes the parser to ignore comments when looking for the comma at the end of a line.

See sass/sass-spec#1532.

### DIFF
--- a/lib/src/parse/sass.dart
+++ b/lib/src/parse/sass.dart
@@ -46,23 +46,12 @@ class SassParser extends StylesheetParser {
 
     var buffer = InterpolationBuffer();
     do {
-      buffer.addInterpolation(almostAnyValue());
+      buffer.addInterpolation(almostAnyValue(omitComments: true));
       buffer.writeCharCode($lf);
-    } while (_endsWithComma(buffer.trailingString) && scanCharIf(isNewline));
+    } while (buffer.trailingString.trimRight().endsWith(',') &&
+        scanCharIf(isNewline));
 
     return buffer.interpolation(scanner.spanFrom(start));
-  }
-
-  // Returns true if [text] ends with a comma, ignoring whitespace and comments.
-  bool _endsWithComma(String text) {
-    var line = text.trimRight().split('\n').last;
-    var silentComment = line.lastIndexOf('//');
-    if (silentComment != -1) line = line.substring(0, silentComment);
-    line = line.trimRight();
-    while (line.endsWith('*/')) {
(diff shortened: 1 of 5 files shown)


lib/src/parse/sass.dart
```diff
@@ -48,12 +48,23 @@ class SassParser extends StylesheetParser {
...
       buffer.writeCharCode($lf);
-    } while (buffer.trailingString.trimRight().endsWith(",") &&
-        scanCharIf(isNewline));
+    } while (_endsWithComma(buffer.trailingString) && scanCharIf(isNewline));
 
     return buffer.interpolation(scanner.spanFrom(start));
   }
 
+  // Returns true if [text] ends with a comma, ignoring whitespace and comments.
+  bool _endsWithComma(String text) {
```
nex3, line 57: I'm concerned that this doesn't respect CSS tokenization rules. For example, if `"//"` somehow made it into the selector, it would be considered a comment here rather than a string. Since `almostAnyValue()` does fully tokenize, maybe we can get it to be a little more clever here. Maybe add an `omitComments` parameter that tells it not to include comments in the buffer?

jathak: Yeah, that makes sense. Originally, I didn't think that would work, since the comments would be omitted in the output, but since even loud comments are omitted within selectors, it should be fine.

jathak: Done

nex3:
Don't forget to add a CHANGELOG entry for this!

jathak:
Do you know what's going on with Appveyor here?

nex3:
Looks like possibly a flake? I'll try restarting it.

PR #1014, opened 2020-05-27
Properly handle escaped `@else`

### LINKED ISSUE #1011: Fails to parse escaped @else
The below SCSS fails to parse using `dart-sass`
```scss
@if false {}

@\65lse {
  a {
    color: red;
  }
}
```
with the error
```scss
Error: This at-rule is not allowed here.
  ╷
3 │ @\65lse {
  │ ^^^^^^^^
  ╵
  foo.scss 3:1  root stylesheet
```
```
dart-sass --version
1.26.3
```

<br/>

`sassc` does not fail, but fails to recognize the escape,
```scss
@\65lse {
  a {
    color: red; } }
```

```
sassc --version
sassc: [NA]
libsass: [NA]
sass2scss: 1.1.1
sass: 3.5
```
(`sassc` was built from source using the most recent commit, https://github.com/sass/sassc/commit/3e73baee972f5f46f4ce1a7a84e0efd3c3563c30, and a very recent libsass commit)

<br>

An escaped version of `@if` does parse and evaluate correctly,
```scss
@\69 f true {
  a {
    color: red;
  }
}
```
(though again libsass does not recognize the escape)

jathak:
Fixes #1011.

This also fixes a bug where `@else` was parsed case-insensitively
(unlike all other Sass at-rules, which must be lowercase).

### DIFF
--- a/lib/src/parse/parser.dart
+++ b/lib/src/parse/parser.dart
@@ -522,11 +522,9 @@ class Parser {
   /// Matching will be case-insensitive unless [caseSensitive] is true.
   @protected
   bool scanIdentChar(int char, {bool caseSensitive = false}) {
-    bool matches(int actual) {
-      if (caseSensitive) return actual == char;
-      if (char >= $a && char <= $z) return equalsLetterIgnoreCase(char, actual);
-      return characterEqualsIgnoreCase(char, actual);
-    }
+    bool matches(int actual) => caseSensitive
+        ? actual == char
+        : characterEqualsIgnoreCase(char, actual);
 
     var next = scanner.peekChar();
     if (matches(next)) {
--- a/lib/src/parse/parser.dart
+++ b/lib/src/parse/parser.dart
@@ -517,33 +517,36 @@ class Parser {
     return true;
   }
 
-  /// Consumes the next character or escape sequence if it matches [letter].
+  /// Consumes the next character or escape sequence if it matches [expected].
   ///
-  /// [letter] must be a lowercase ASCII character. It will be matched
-  /// case-insensitively unless [caseSensitive] is true.
+  /// Matching will be case-insensitive unless [caseSensitive] is true.
   @protected
-  bool scanLetter(int letter, {bool caseSensitive = false}) {
+  bool scanIdentChar(int char, {bool caseSensitive = false}) {
+    bool matches(int actual) {
+      if (caseSensitive) return actual == char;
+      if (char >= $a && char <= $z) return equalsLetterIgnoreCase(char, actual);
+      return characterEqualsIgnoreCase(char, actual);
+    }
+
     var next = scanner.peekChar();
-    if (equalsLetter(letter, next, caseSensitive: caseSensitive)) {
+    if (matches(next)) {
       scanner.readChar();
       return true;
     } else if (next == $backslash) {
       var start = scanner.state;
-      next = escapeCharacter();
-      if (equalsLetter(letter, next, caseSensitive: caseSensitive)) return true;
+      if (matches(escapeCharacter())) return true;
       scanner.state = start;
     }
     return false;
   }
 
   /// Consumes the next character or escape sequence and asserts it matches
-  /// [letter].
+  /// [char].
   ///
-  /// [letter] must be a lowercase ASCII character. It will be matched
-  /// case-insensitively unless [caseSensitive] is true.
+  /// Matching will be case-insensitive unless [caseSensitive] is true.
   @protected
-  void expectLetter(int letter, {bool caseSensitive = false}) {
-    if (scanLetter(letter, caseSensitive: caseSensitive)) return;
+  void expectIdentChar(int letter, {bool caseSensitive = false}) {
+    if (scanIdentChar(letter, caseSensitive: caseSensitive)) return;
 
     scanner.error('Expected "${String.fromCharCode(letter)}".',
         position: scanner.position);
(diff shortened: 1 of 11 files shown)


lib/src/parse/parser.dart
```diff
@@ -520,21 +520,26 @@ class Parser {
   /// Consumes the next character if it's equal to [letter], ignoring ASCII
   /// case.
```
nex3, line 521: This doc comment isn't always accurate anymore.

lib/src/parse/parser.dart
```diff
@@ -520,21 +520,26 @@ class Parser {
...
   @protected
-  bool scanCharIgnoreCase(int letter) {
-    if (!equalsLetterIgnoreCase(letter, scanner.peekChar())) return false;
-    scanner.readChar();
-    return true;
+  bool scanAsciiChar(int letter, {bool caseSensitive = false}) {
```
nex3, line 523: It's confusing to me that this is called `scanAsciiChar()` when the only relationship it has to ASCII is ignoring the case of letters, and then only if `caseSensitive` is `false`. As a reader, I'd expect `scanAsciiChar($unicodeSnowman)` to be invalid, but it's totally fine.

I'd probably just call this `scanChar()`, and explain in the doc comment that you'd use it over `scanner.scanChar()` when you want the case-sensitivity to be configurable.

lib/src/parse/parser.dart
```diff
@@ -520,21 +520,26 @@ class Parser {
...
+    var next = scanner.peekChar();
+    if (caseSensitive ? letter == next : equalsLetterIgnoreCase(letter, next)) {
```
nex3, line 525: The relationship between this function and `StringScanner.scanChar()` would probably be clearer if you explicitly fell back to the latter when `caseSensitive` is `true`:

```dart
if (caseSensitive) return scanner.scanChar(letter);
if (!equalsLetterIgnoreCase(letter, scanner.peekChar())) return false;
scanner.readChar();
return true;
```

lib/src/parse/parser.dart
```diff
@@ -598,14 +603,17 @@ class Parser {
   }
 
   /// Consumes an identifier if its name exactly matches [text].
+  ///
+  /// [text] must only contain lowercase ASCII characters. It will be matched
```
nex3, line 607: This restriction could either mean "any alphabetic ASCII characters in `text` must be lowercase" or "`text` may not contain any characters other than lowercase alphabetic ASCII characters". I think you mean the former, since it seems like it would make sense to write `scanIdentifier("else-if")`, but you should be explicit.

It's probably also not a bad idea to add an assertion that verifies this requirement as well.

lib/src/parse/parser.dart
```diff
@@ -615,15 +623,35 @@ class Parser {
...
+    if (scanAsciiChar(letter, caseSensitive: caseSensitive)) return true;
+    var start = scanner.state;
+    if (scanner.peekChar() == $backslash) {
+      var escaped = escape().codeUnitAt(0);
+      if (!caseSensitive) escaped = toLowerCase(escaped);
+      if (escaped == letter) return true;
+      scanner.state = start;
+    }
+    return false;
+  }
```
nex3, line 640: Is there a reason not to check for escapes for every call to `scanAsciiChar()`?

lib/src/parse/parser.dart
nex3: I'd like to make it possible for this to scan *any* identifier. I think it's mostly coincidental that we've managed to get away with only using it for alphabetical identifiers so far. Maybe have it check if the character is a letter and choose to do `scanLetter()` or `scanIdentChar()`, where the latter handles escape characters? Or else replace `scanLetter()` *with* `scanIdentChar()` and build in the case check there.

lib/src/parse/parser.dart
```diff
@@ -517,24 +517,33 @@ class Parser {
...
+  ///
+  /// [letter] must be a lowercase ASCII character. It will be matched
+  /// case-insensitively unless [caseSensitive] is true.
   @protected
-  bool scanCharIgnoreCase(int letter) {
-    if (!equalsLetterIgnoreCase(letter, scanner.peekChar())) return false;
-    scanner.readChar();
-    return true;
+  bool scanLetter(int letter, {bool caseSensitive = false}) {
+    var start = scanner.state;
```
nex3, line 526: Storing the state for every letter could get expensive, since it involves an allocation of a state object, and parse speed can be important for a lot of ancillary tools like static analyzers. It may be worth only storing this if we actually see a backslash in practice, which is likely to be very rare.

lib/src/parse/parser.dart
jathak: Whoops. Missed this reply. Will do that

lib/src/parse/parser.dart
```diff
@@ -517,24 +517,39 @@ class Parser {
...
   @protected
-  bool scanCharIgnoreCase(int letter) {
-    if (!equalsLetterIgnoreCase(letter, scanner.peekChar())) return false;
-    scanner.readChar();
-    return true;
+  bool scanIdentChar(int char, {bool caseSensitive = false}) {
+    bool matches(int actual) {
+      if (caseSensitive) return actual == char;
+      if (char >= $a && char <= $z) return equalsLetterIgnoreCase(char, actual);
+      return characterEqualsIgnoreCase(char, actual);
```
nex3, line 528: I forgot we had this function! Why not just use this instead of toggling between it and `equalsLetterIgnoreCase()`? It's already optimized to work with either letters or not.

PR #1028, opened 2020-06-17
Release 1.26.9

Mstrodl:
(Update to `node_preamble` is available)

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -44,5 +44,5 @@ dev_dependencies:
   stream_channel: ">=1.0.0 <3.0.0"
   test_descriptor: "^1.1.0"
   test_process: "^1.0.0-rc.1"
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -44,5 +44,6 @@ dev_dependencies:
   stream_channel: ">=1.0.0 <3.0.0"
   test_descriptor: "^1.1.0"
   test_process: "^1.0.0-rc.1"
-  test: ">=0.12.42 <2.0.0"
+  # Update this constraint once dart-lang/test#1278 is resolved.
+  test: "1.14.7"
```
nex3, line 48:
```suggestion
  test: ">=0.12.42 <=1.14.7"
```

It's generally good practice to have a range even when you're effectively pinning to the top of that range.

PR #1034, opened 2020-06-22
Replace use of deprecated mapMap with map literals

jathak:
The test package fixed the macOS issue in 1.15.1, so I've also bumped that version.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -44,5 +44,5 @@ dev_dependencies:
   stream_channel: ">=1.0.0 <3.0.0"
   test_descriptor: "^1.1.0"
   test_process: "^1.0.0-rc.1"
-  test: ">=0.12.42 <2.0.0"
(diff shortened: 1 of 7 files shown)


pubspec.yaml
```diff
@@ -44,5 +44,5 @@ dev_dependencies:
   stream_channel: ">=1.0.0 <3.0.0"
   test_descriptor: "^1.1.0"
   test_process: "^1.0.0-rc.1"
-  test: ">=0.12.42 <2.0.0"
+  test: ">=1.15.1 <2.0.0"
```
nex3, line 47: Usually we don't increase the lower bound to get bug fixes. It's still (presumably) the case that version 0.12.42 works fine, it's just 1.15.0 *specifically* that doesn't. Now that 1.15.1 is out, there's no reason the old version would be selected anyway.

PR #1037, opened 2020-06-25
Fix release notes to include changes

jathak:


jathak:
I'm not sure why this is failing on Windows (https://travis-ci.org/github/sass/dart-sass/jobs/702208809). I can't replicate it on my own Windows machine

jathak:
I had already restarted it once, so it's now failed three times in a row

nex3:
Dang. It looks like it could be some sort of newline normalization issue. Maybe your Windows machine's Git configuration is different than Travis's? Either way, I wonder if it would be fixed if the regexp for the changelog header was changed to search for `\r?\n` rather than just `\n`.

jathak:
Yeah, that was my guess. IIRC, I cloned the repo with Git on WSL even tough I was using the Windows version of Dart when testing.

PR #1055, opened 2020-08-06
Improve the error messages for syntax errors

nex3:
See https://github.com/sass/sass-spec/pull/1548

### DIFF
--- a/lib/src/parse/parser.dart
+++ b/lib/src/parse/parser.dart
@@ -527,7 +527,7 @@ class Parser {
         : characterEqualsIgnoreCase(char, actual);
 
     var next = scanner.peekChar();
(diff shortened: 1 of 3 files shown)


lib/src/parse/parser.dart
```diff
@@ -459,7 +460,13 @@ class Parser {
     }
 
     if (identifierStart ? isNameStart(value) : isName(value)) {
-      return String.fromCharCode(value);
+      try {
+        return String.fromCharCode(value);
+      } on RangeError catch (error) {
```
Awjin, line 465:
```suggestion
      } on RangeError {
```

nex3: Thanks!

PR #1073, opened 2020-09-04
Don't crash when using Infinity or NaN as a key in a map

nex3:
Closes #3126
See sass/sass-spec#1556


connorskees:
Would it make sense to instead deny both `NaN` and `Infinity` from being map keys? That `map-get(((0/0): a), (0/0))` is equal to `null` may be surprising to some. As this was already a crash, it should not break any existing code relying on `dart-sass`. In fact, denying `NaN` and `Infinity` as map keys may prevent silent errors that result from the disparity in equality between `libsass` and `dart-sass` (see https://github.com/sass/libsass/issues/3126).

Awjin (approved):
I'm ok with this workaround, and revisiting once floating point is officially defined in Sass.

PR #1075, opened 2020-09-14
Add support for nested maps to has-key

jathak:
See https://github.com/sass/sass-spec/pull/1558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.27.0
+
+* Add support for nested maps in `map.has-key`.
+
 ## 1.26.11
 
 * **Potentially breaking bug fix:** `selector.nest()` now throws an error
--- a/lib/src/functions/map.dart
+++ b/lib/src/functions/map.dart
@@ -67,10 +67,19 @@ final _values = _function(
     (arguments) => SassList(
         arguments[0].assertMap("map").contents.values, ListSeparator.comma));
 
-final _hasKey = _function("has-key", r"$map, $key", (arguments) {
+final _hasKey = _function("has-key", r"$map, $key, $keys...", (arguments) {
   var map = arguments[0].assertMap("map");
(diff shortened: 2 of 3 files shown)


lib/src/functions/map.dart
```diff
@@ -67,10 +67,19 @@ final _values = _function(
...
         arguments[0].assertMap("map").contents.values, ListSeparator.comma));
 
-final _hasKey = _function("has-key", r"$map, $key", (arguments) {
+final _hasKey = _function("has-key", r"$map, $key, $keys...", (arguments) {
   var map = arguments[0].assertMap("map");
-  var key = arguments[1];
-  return SassBoolean(map.contents.containsKey(key));
+  var allKeys = [arguments[1], ...arguments[2].asList];
+  var intermediateKeys = allKeys.sublist(0, allKeys.length - 1);
+  for (var key in intermediateKeys) {
```
nex3, line 74:
```suggestion
  for (var key in allKeys.take(allKeys.length - 1)) {
```

Not only is this terser, it also avoids creating an unnecessary intermediate list :smiley:.

nex3 (changes requested):
Please add a CHANGELOG entry as well.

lib/src/functions/map.dart
jathak: Done

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.27.0
+
+* Add support for nested maps in `map.has-key`.
```
Awjin, line 3: This should also link to the docs.

jathak: Will do. I'm going to wait until #1076 is merged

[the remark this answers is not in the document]
jathak: Done

nex3 (approved):

PR #1076, opened 2020-09-14
Add support for nested maps to map-get

jathak:
See https://github.com/sass/sass-spec/pull/1559

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ## 1.27.0
 
-* Add support for nested maps in `map.get`.
+* Add support for nested maps to `map.get()`.
+  For example, `map.get((a: (b: (c: d))), a, b, c)` would return `d`.
+  See [the documentation][maps] for more details.
+
+[maps]: https://sass-lang.com/documentation/modules/map
 
 ## 1.26.11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.27.0
+
+* Add support for nested maps in `map.get`.
+
 ## 1.26.11
 
 * **Potentially breaking bug fix:** `selector.nest()` now throws an error
--- a/lib/src/functions/map.dart
+++ b/lib/src/functions/map.dart
@@ -24,10 +24,19 @@ final global = UnmodifiableListView([
 final module = BuiltInModule("map",
     functions: [_get, _merge, _remove, _keys, _values, _hasKey]);
 
-final _get = _function("get", r"$map, $key", (arguments) {
+final _get = _function("get", r"$map, $key, $keys...", (arguments) {
   var map = arguments[0].assertMap("map");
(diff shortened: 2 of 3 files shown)


lib/src/functions/map.dart
```diff
@@ -24,10 +24,19 @@ final global = UnmodifiableListView([
...
     functions: [_get, _merge, _remove, _keys, _values, _hasKey]);
 
-final _get = _function("get", r"$map, $key", (arguments) {
+final _get = _function("get", r"$map, $key, $keys...", (arguments) {
   var map = arguments[0].assertMap("map");
-  var key = arguments[1];
-  return map.contents[key] ?? sassNull;
+  var allKeys = [arguments[1], ...arguments[2].asList];
+  var intermediateKeys = allKeys.sublist(0, allKeys.length - 1);
+  for (var key in intermediateKeys) {
```
nex3, line 31: Same as the other CL, you can use `.take()` to avoid an intermediate list.

nex3 (changes requested):
Please add a CHANGELOG entry here as well.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.27.0
+
+* Add support for nested maps in `map.get`.
```
nex3, line 3: Can you add a bit more detail here? For some percentage of users, this will be their first encounter with this feature, so it would be nice to give them at least an example of how it works, as well as maybe a link to further documentation on the website.

nex3 (changes requested):

CHANGELOG.md
jathak: Done.

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.27.0
+
+* Add support for nested maps to `map.get()`.
+  For example, `map.get((a: (b: (c: d))), a, b, c)` would return `d`.
+  See [the documentation][maps] for more details.
+
+[maps]: https://sass-lang.com/documentation/modules/map
```
Awjin, line 7:
```suggestion
[map-get]: https://sass-lang.com/documentation/modules/map#get
```

(for consistency with the other PRs)

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.27.0
+
+* Add support for nested maps to `map.get()`.
+  For example, `map.get((a: (b: (c: d))), a, b, c)` would return `d`.
+  See [the documentation][maps] for more details.
```
Awjin, line 5:
```suggestion
  See [the documentation][map-get] for more details.
```

lib/src/functions/map.dart
Awjin: @jathak I think this comment still applies (replacing `sublist` with `take`).

jathak: Whoops. Good catch

nex3 (approved):

PR #1077, opened 2020-09-15
Add a map.deep-merge() function

nex3:
This also adds a Value.tryMap() function, which was useful for
implementing this and may be more generally useful to users as well.

See sass/sass#2836
See sass/sass-spec#1560


Awjin (approved):

jathak:
Was this supposed to be merged into master or feature.nested-maps?

PR #1078, opened 2020-09-15
Add a map.set() function

Awjin:
See https://github.com/sass/sass/pull/2836
See https://github.com/sass/sass-spec/pull/1561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,18 @@
 ## 1.27.0
 
-* Adds an overloaded `map.set()` function:
-  * `set($map, $key, $value)`: Adds to or updates $map with the specified $key
-     and $value.
-  * `set($map, $args...)`: Allows adding to or updating a map that is nested
-     within $map. $args contains multiple keys, followed by a value. The keys
-     form a path to the nested map in `$map` where the value is to be added or
-     updated.
+* Adds an overloaded `map.set()` function.
+
+  `set($map, $key, $value)`: Adds to or updates `$map` with the specified
+  `$key` and `$value`.
+
+  `set($map, $args...)`: Adds to or updates a map that is nested within
+  `$map`. `$args` contains multiple keys, followed by a value. The keys
+  form a path to the nested map in `$map` where the value is to be added or
+  updated.
+
+  See [the Sass documentation][map-set] for more details.
+
+  [map-set]: https://sass-lang.com/documentation/modules/map#set
 
 * Add support for nested maps to `map.get()`.
   For example, `map.get((a: (b: (c: d))), a, b, c)` would return `d`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 * Adds an overloaded `map.set()` function.
 
-  `set($map, $key, $value)`: Adds to or updates `$map` with the specified
+  `map.set($map, $key, $value)`: Adds to or updates `$map` with the specified
   `$key` and `$value`.
 
-  `set($map, $args...)`: Adds to or updates a map that is nested within
+  `map.set($map, $args...)`: Adds to or updates a map that is nested within
   `$map`. `$args` contains multiple keys, followed by a value. The keys
   form a path to the nested map in `$map` where the value is to be added or
   updated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## 1.27.0
 
+* Adds an overloaded `map.set()` function:
+  * `set($map, $key, $value)`: Adds to or updates $map with the specified $key
+     and $value.
+  * `set($map, $args...)`: Allows adding to or updating a map that is nested
+     within $map. $args contains multiple keys, followed by a value. The keys
+     form a path to the nested map in `$map` where the value is to be added or
+     updated.
+
 * Add support for nested maps to `map.get()`.
   For example, `map.get((a: (b: (c: d))), a, b, c)` would return `d`.
   See [the documentation][map-get] for more details.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.27.0
+
+* Adds an overloaded `map.set()` function:
+  * `set($map, $key, $value)`: Adds to or updates $map with the specified $key
+     and $value.
+  * `set($map, $args...)`: Allows adding to or updating a map that is nested
+     within $map. $args contains multiple keys, followed by a value. The keys
+     form a path to the nested map in `$map` where the value is to be added or
+     updated.
```
nex3, line 9: Nit: code-format all argument names.

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.27.0
+
+* Adds an overloaded `map.set()` function:
+  * `set($map, $key, $value)`: Adds to or updates $map with the specified $key
+     and $value.
+  * `set($map, $args...)`: Allows adding to or updating a map that is nested
```
nex3, line 6: Nit: Use active voice ("Adds to or updates" rather than "Allows adding to or updating").

lib/src/functions/map.dart
```diff
@@ -21,8 +22,25 @@ final global = UnmodifiableListView([
 ]);
 
 /// The Sass map module.
-final module = BuiltInModule("map",
-    functions: [_get, _merge, _remove, _keys, _values, _hasKey]);
+final module = BuiltInModule("map", functions: [
+  _get, _merge, _remove, _keys, _values, _hasKey, //
+  BuiltInCallable.overloadedFunction("set", {
```
nex3, line 27: Please follow the pattern used for other functions by declaring this as a top-level field and then including it by name in the `functions` list.

lib/src/functions/map.dart
```diff
@@ -73,6 +91,26 @@ final _hasKey = _function("has-key", r"$map, $key", (arguments) {
   return SassBoolean(map.contents.containsKey(key));
 });
 
+SassMap _set(SassMap map, List<Value> keys, Value value) {
```
nex3, line 94: Please document this.

lib/src/functions/map.dart
```diff
@@ -73,6 +91,26 @@ final _hasKey = _function("has-key", r"$map, $key", (arguments) {
...
+  var key = keys.first;
+
+  if (keys.length == 1) {
+    mutableMap[key] = value;
+    return SassMap(mutableMap);
+  }
+
+  SassMap nestedMap;
+  try {
+    nestedMap = mutableMap[key].assertMap();
```
nex3, line 105: Once I land https://github.com/sass/dart-sass/pull/1077, you can use `.tryMap()` here to avoid the unnecessary catch.

lib/src/functions/map.dart
```diff
@@ -73,6 +91,26 @@ final _hasKey = _function("has-key", r"$map, $key", (arguments) {
...
+  } on SassScriptException {
+    nestedMap = SassMap({});
+  }
+
+  mutableMap[key] = _set(nestedMap, keys.sublist(1), value);
```
nex3, line 110: This is pretty nitty, but creating a separate sublist for every recursion could result in a lot of extra data structure allocations which are relatively expensive. It may be worth writing this in an iterative style instead of recursive to avoid that.

nex3 (changes requested):

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.27.0
+
+* Adds an overloaded `map.set()` function:
```
nex3, line 3: Can you add a link to the sass-lang.com documentation as well? It's not written yet, but when it is it'll be at https://sass-lang.com/documentation/modules/map#set.

lib/src/functions/map.dart
Awjin: nice!

lib/src/functions/map.dart
Awjin: I had trouble expressing this nicely as iterative, so instead I passed an index into the array to avoid sublisting it over and over. WDYT?

lib/src/functions/map.dart
```diff
@@ -97,23 +98,23 @@ final _hasKey = _function("has-key", r"$map, $key", (arguments) {
...
 
-  SassMap nestedMap;
-  try {
-    nestedMap = mutableMap[key].assertMap();
-  } on SassScriptException {
-    nestedMap = SassMap({});
-  }
-
-  mutableMap[key] = _set(nestedMap, keys.sublist(1), value);
+  var nestedMap = mutableMap[key].tryMap() ?? SassMap({});
```
nex3, line 116:
```suggestion
  var nestedMap = mutableMap[key].tryMap() ?? const SassMap.empty();
```

Using `const` constructors ensures that the same empty map instance is re-used everywhere, which saves allocations.

lib/src/functions/map.dart
nex3: I can see why a directly iterative solution would be messy, since you'd need to track an explicit stack of parent maps *and* their associated keys. I think this solution works well enough.

nex3 (approved):
Please don't merge this until I successfully land sass/dart-sass#1080 on the feature branch.

nex3:
Okay, this should be good to go now once you rebase onto the latest feature.nested-maps.

lib/src/functions/map.dart
Awjin: Don't we want new empty map instances to be created for each key?

nex3: SassScript values are all immutable, so there's no value to creating new instances when re-using an existing instance would do.

Awjin: Ah, I see

PR #1083, opened 2020-09-16
Adds nested map support to map-merge.

Awjin:
See https://github.com/sass/sass/pull/2836
See https://github.com/sass/sass-spec/pull/1566

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,9 @@
   `map.set($map, $key, $value)`: Adds to or updates `$map` with the specified
   `$key` and `$value`.
 
-  `map.set($map, $args...)`: Adds to or updates a map that is nested within
-  `$map`. `$args` contains multiple keys, followed by a value. The keys
-  form a path to the nested map in `$map` where the value is to be added or
-  updated.
+  `map.set($map, $keys..., $value)`: Adds to or updates a map that is nested
+  within `$map`. The `$keys` form a path to the nested map in `$map`, into
+  which `$value` is inserted.
 
   See [the Sass documentation][map-set] for more details.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,8 @@
 
 * Adds an overload to `map.merge()` that supports merging a nested map.
 
-  `map.merge($map1, $args...)`: `$args` contains multiple keys, followed by a
-  `map2`. The keys form a path to the nested map in `$map1`, into which map2
-  gets merged.
+  `map.merge($map1, $keys..., $map2)`: The `$keys` form a path to the nested map
+  in `$map1`, into which `$map2` gets merged.
 
   See [the Sass documentation][map-merge] for more details.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 ## 1.27.0
 
+* Adds an overload to `map.merge()` that supports merging a nested map.
+
+  `map.merge($map1, $args...)`: `$args` contains multiple keys, followed by a
+  `map2`. The keys form a path to the nested map in `$map1`, into which map2
+  gets merged.
+
+  See [the Sass documentation][map-merge] for more details.
+
+  [map-merge]: https://sass-lang.com/documentation/modules/map#merge
+
 * Adds an overloaded `map.set()` function.
 
   `map.set($map, $key, $value)`: Adds to or updates `$map` with the specified
--- a/lib/src/functions/map.dart
+++ b/lib/src/functions/map.dart
@@ -72,7 +72,7 @@ final _merge = BuiltInCallable.overloadedFunction("merge", {
       throw SassScriptException("Expected \$args to contain a map.");
     }
     var map2 = args.last.assertMap("map2");
-    return _modify(map1, args.sublist(0, args.length - 1), (Value oldValue) {
+    return _modify(map1, args.sublist(0, args.length - 1), (oldValue) {
       var nestedMap = oldValue?.tryMap();
       if (nestedMap == null) return map2;
       return SassMap({...nestedMap.contents, ...map2.contents});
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 ## 1.27.0
 
+* Adds an overload to `map.merge()` that supports merging a nested map.
+
+  `map.merge($map1, $args...)`: `$args` contains multiple keys, followed by a
```
nex3, line 5: I'd write this as `map.merge($map1, $keys..., $map2)`. Even though `$map2` can't *actually* be passed by name, in informal documentation like this it'll help make it clearer what the intended usage is.

lib/src/functions/map.dart
```diff
@@ -57,10 +49,26 @@ final _set = BuiltInCallable.overloadedFunction("set", {
...
+    var map1 = arguments[0].assertMap("map1");
+    var args = arguments[1].asList;
+    if (args.isEmpty) {
+      throw SassScriptException("Expected \$args to contain a key.");
+    } else if (args.length == 1) {
+      throw SassScriptException("Expected \$args to contain a map.");
+    }
+    var map2 = args.last.assertMap("map2");
+    var keys = args.sublist(0, args.length - 1);
+    var nestedMap = _getImpl(map1, keys).tryMap() ?? const SassMap.empty();
```
nex3, line 68: This currently traverses the maps twice, once for `_getImpl()` and again for `_setImpl()`. If you changed `_setImpl()` to something like `SassMap _modify(SassMap map, List<Value> keys, SassValue modify(Value old))` then you could handle the traversal once in a more general way.

nex3 (changes requested):

lib/src/functions/map.dart
Awjin: Good call, thanks

lib/src/functions/map.dart
```diff
@@ -43,7 +43,7 @@ final _get = _function("get", r"$map, $key, $keys...", (arguments) {
 final _set = BuiltInCallable.overloadedFunction("set", {
   r"$map, $key, $value": (arguments) {
     var map = arguments[0].assertMap("map");
-    return _setImpl(map, [arguments[1]], arguments[2]);
+    return _modify(map, [arguments[1]], (oldValue) => arguments[2]);
```
nex3, line 46:
```suggestion
    return _modify(map, [arguments[1]], (_) => arguments[2]);
```

This helps make it clear at a glance that the argument isn't used.

lib/src/functions/map.dart
```diff
@@ -53,14 +53,39 @@ final _set = BuiltInCallable.overloadedFunction("set", {
     } else if (args.length == 1) {
       throw SassScriptException("Expected \$args to contain a value.");
     }
-    return _setImpl(map, args.sublist(0, args.length - 1), args.last);
+    return _modify(
+        map, args.sublist(0, args.length - 1), (oldValue) => args.last);
```
nex3, line 57:
```suggestion
        map, args.sublist(0, args.length - 1), (_) => args.last);
```

lib/src/functions/map.dart
```diff
@@ -53,14 +53,39 @@ final _set = BuiltInCallable.overloadedFunction("set", {
...
+    var map1 = arguments[0].assertMap("map1");
+    var args = arguments[1].asList;
+    if (args.isEmpty) {
+      throw SassScriptException("Expected \$args to contain a key.");
+    } else if (args.length == 1) {
+      throw SassScriptException("Expected \$args to contain a map.");
+    }
+    var keys = args.sublist(0, args.length - 1);
+    var map2 = args.last.assertMap("map2");
+    SassMap _mergeCallback(Value oldMap) {
```
nex3, line 77: Since this is only used once, I'd just define write it inline in the call to `_modify`.

lib/src/functions/map.dart
```diff
@@ -53,14 +53,39 @@ final _set = BuiltInCallable.overloadedFunction("set", {
...
+      // If oldMap is null, this indicates there is no nested map at the key.
+      // If this is the case, set the value at the key to map2, which is akin to
+      // merging map2 into an empty map. This behavior is consistent with that
+      // of map.set().
+      return (oldMap == null)
```
nex3, line 82: Style nit

```suggestion
      return oldMap == null
```

lib/src/functions/map.dart
```diff
@@ -53,14 +53,39 @@ final _set = BuiltInCallable.overloadedFunction("set", {
...
+          ? map2
+          : SassMap({...oldMap.assertMap().contents, ...map2.contents});
```
nex3, line 84: According to the spec, the new map is expected to replace the old value if it's not a map, rather than throwing an error.

nex3 (changes requested):

lib/src/functions/map.dart
Awjin: I was hoping there was a better way to do this :)

lib/src/functions/map.dart
Awjin: Ah, good catch.

lib/src/functions/map.dart
```diff
@@ -72,19 +71,12 @@ final _merge = BuiltInCallable.overloadedFunction("merge", {
...
-      // If this is the case, set the value at the key to map2, which is akin to
-      // merging map2 into an empty map. This behavior is consistent with that
-      // of map.set().
-      return (oldMap == null)
-          ? map2
-          : SassMap({...oldMap.assertMap().contents, ...map2.contents});
-    }
-
-    return _modify(map1, keys, _mergeCallback);
+    return _modify(map1, args.sublist(0, args.length - 1), (Value oldValue) {
```
nex3, line 75:
```suggestion
    return _modify(map1, args.sublist(0, args.length - 1), (oldValue) {
```

This type is inferred from `_modify()`'s signature. That's one of the benefits of writing callbacks inline: you have to specify fewer types manually.

nex3 (approved):

PR #1085, opened 2020-09-17
Support HWB colors in SassColor

nex3:
See sass/sass#2835

### DIFF
--- a/lib/src/value/color.dart
+++ b/lib/src/value/color.dart
@@ -58,14 +58,14 @@ class SassColor extends Value implements ext.SassColor {
   num get whiteness {
     // Because HWB is (currently) used much less frequently than HSL or RGB, we
     // don't cache its values because we expect the memory overhead of doing so
-    // to outweight the cost of recalculating it on access.
+    // to outweigh the cost of recalculating it on access.
     return math.min(math.min(red, green), blue) / 255 * 100;
   }
 
   num get blackness {
     // Because HWB is (currently) used much less frequently than HSL or RGB, we
     // don't cache its values because we expect the memory overhead of doing so
-    // to outweight the cost of recalculating it on access.
+    // to outweigh the cost of recalculating it on access.
     return 100 - math.max(math.max(red, green), blue) / 255 * 100;
   }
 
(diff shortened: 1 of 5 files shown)


lib/src/value/color.dart
```diff
@@ -55,6 +55,20 @@ class SassColor extends Value implements ext.SassColor {
 
   num _lightness;
 
+  num get whiteness {
+    // Because HWB is (currently) used much less frequently than HSL or RGB, we
+    // don't cache its values because we expect the memory overhead of doing so
+    // to outweight the cost of recalculating it on access.
```
Awjin, line 61: nit:

```suggestion
    // to outweigh the cost of recalculating it on access.
```

lib/src/value/color.dart
```diff
@@ -55,6 +55,20 @@ class SassColor extends Value implements ext.SassColor {
...
+    return math.min(math.min(red, green), blue) / 255 * 100;
+  }
+
+  num get blackness {
+    // Because HWB is (currently) used much less frequently than HSL or RGB, we
+    // don't cache its values because we expect the memory overhead of doing so
+    // to outweight the cost of recalculating it on access.
```
Awjin, line 68: nit:

```suggestion
    // to outweigh the cost of recalculating it on access.
```

lib/src/value/color.dart
```diff
@@ -81,6 +95,31 @@ class SassColor extends Value implements ext.SassColor {
...
+      scaledBlackness /= sum;
+    }
+
+    var factor = 1 - scaledWhiteness - scaledBlackness;
+    int toRgb(num hue) {
+      var channel = _hueToRgb(0, 1, hue) * factor + scaledWhiteness;
+      return fuzzyRound(channel * 255);
+    }
+
+    return SassColor.rgb(
```
Awjin, line 116: It could be worth documenting why this returns an rgb color.

Awjin (approved):
Needs a pubspec bump, otherwise the code LGTM.

lib/src/value/color.dart
nex3: Done.

lib/src/value/color.dart
nex3: Done.

lib/src/value/color.dart
nex3: Done.

nex3 (commented):
> Needs a pubspec bump, otherwise the code LGTM.

Done.

Issue #1088, opened 2020-09-17
Support nested map functions in sass:map


nex3:
See https://github.com/sass/sass/issues/1739 and https://github.com/sass/sass/pull/2836.

nex3:
We're currently landing provisional support for this in the [feature.nested-maps](https://github.com/sass/dart-sass/commits/feature.nested-maps) branch, including the following PRs:

* https://github.com/sass/dart-sass/pull/1077
* https://github.com/sass/dart-sass/pull/1076
* https://github.com/sass/dart-sass/pull/1075
* https://github.com/sass/dart-sass/pull/1078

PR #1091, opened 2020-09-18
Add a map.deep-remove() function

jathak:
See sass/sass#1739
See #1088

See https://github.com/sass/sass-spec/pull/1569

### DIFF
--- a/lib/src/functions/map.dart
+++ b/lib/src/functions/map.dart
@@ -23,7 +23,15 @@ final global = UnmodifiableListView([
 
 /// The Sass map module.
 final module = BuiltInModule("map", functions: [
-  _get, _set, _merge, _remove, _keys, _values, _hasKey, _deepMerge //
+  _get,
+  _set,
+  _merge,
+  _remove,
+  _keys,
+  _values,
+  _hasKey,
+  _deepMerge,
+  _deepRemove
 ]);
 
 final _get = _function("get", r"$map, $key, $keys...", (arguments) {
@@ -86,6 +94,19 @@ final _deepMerge = _function("deep-merge", r"$map1, $map2", (arguments) {
   return _deepMergeImpl(map1, map2);
 });
 
+final _deepRemove =
+    _function("deep-remove", r"$map, $key, $keys...", (arguments) {
+  var map = arguments[0].assertMap("map");
+  var keys = [arguments[1], ...arguments[2].asList];
+  return _modify(map, keys.sublist(0, keys.length - 1), (value) {
+    var nestedMap = value.tryMap();
+    if (nestedMap?.contents?.containsKey(keys.last) ?? false) {
+      return SassMap(Map.of(nestedMap.contents)..remove(keys.last));
+    }
+    return value;
+  });
+});
+
 final _remove = BuiltInCallable.overloadedFunction("remove", {
   // Because the signature below has an explicit `$key` argument, it doesn't
   // allow zero keys to be passed. We want to allow that case, so we add an
@@ -139,7 +160,10 @@ final _hasKey = _function("has-key", r"$map, $key, $keys...", (arguments) {
 /// leads to the targeted map. If any value along the path is not a map, and
 /// [overwrite] is true, this inserts a new map at that key and overwrites the
 /// current value. Otherwise, this fails and returns [map] with no changes.
-SassMap _modify(SassMap map, List<Value> keys, Value modify(Value old),
+///
+/// If no keys are provided, this passes [map] directly to modify and returns
+/// the result.
+Value _modify(SassMap map, List<Value> keys, Value modify(Value old),
     [bool overwrite = true]) {
   SassMap _modifyNestedMap(SassMap map, int index) {
     var mutableMap = Map.of(map.contents);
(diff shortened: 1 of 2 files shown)


lib/src/functions/map.dart
```diff
@@ -86,6 +94,19 @@ final _deepMerge = _function("deep-merge", r"$map1, $map2", (arguments) {
   return _deepMergeImpl(map1, map2);
 });
 
+final _deepRemove =
+    _function("deep-remove", r"$map, $key, $keys...", (arguments) {
```
nex3, line 98: To match the spec:

```suggestion
    _function("deep-remove", r"$map, $keys...", (arguments) {
```

nex3 (approved):

lib/src/functions/map.dart
```diff
@@ -86,6 +94,19 @@ final _deepMerge = _function("deep-merge", r"$map1, $map2", (arguments) {
...
+  var map = arguments[0].assertMap("map");
+  var keys = [arguments[1], ...arguments[2].asList];
```
Awjin, line 100: According to the spec, this should throw an error for fewer than 2 keys.

lib/src/functions/map.dart
```diff
@@ -86,6 +94,19 @@ final _deepMerge = _function("deep-merge", r"$map1, $map2", (arguments) {
...
+  return _modify(map, keys.sublist(0, keys.length - 1), (value) {
```
Awjin, line 101: Should this call `_modify` with `overwrite = false`?

lib/src/functions/map.dart
jathak: Good catch.

@nex3: Is this the behavior we actually want? Since the arguments to `deep-remove` are supposed to parallel `get` and `has-key`, would it make sense to allow fewer than 2 keys (either requiring 1 or requiring 0)? If we do require a minimum number of keys, would it make sense to have the signature list them separately? (e.g. `$map, $key1, $key2, $keys...` or something like that)

Awjin: FWIW this is the error message for `map.set()` when `$args` has too few items: https://github.com/sass/dart-sass/pull/1083/files#diff-132af1066cf93730a37dc60a5fb3644fR69

nex3: I don't think requiring 0 keys makes sense, but only requiring 1 probably does so people can universally use `deep-replace()` instead of `replace()` if they want. You also make a good point about the `$map, $key, $keys...` signature being more in line with `get()`. Can you send a PR to update the spec?

lib/src/functions/map.dart
nex3: I'm wondering if `_modify()` should handle this automatically by allowing the callback to return `null` to indicate that nothing should be changed.

Awjin: Good call, that's much cleaner.

Awjin: Although, would that require us to recurse to the leaf node in all cases, even if we know that intermediate maps won't be necessary?

nex3: Once you reach the point where you don't have an actual map, there's nothing more to recurse through anyway—you can immediately call `modify(null)`, and choose whether to create nested maps or just exit early based on the result.

lib/src/functions/map.dart
jathak: Updated the spec

lib/src/functions/map.dart
```diff
@@ -131,36 +152,41 @@ final _hasKey = _function("has-key", r"$map, $key, $keys...", (arguments) {
...
     }
 
     nestedMap ??= const SassMap.empty();
-    mutableMap[key] = _modifyNestedMap(nestedMap, index + 1);
+    mutableMap[key] = _modifyNestedMap(nestedMap, newValue);
     return SassMap(mutableMap);
   }
 
-  return _modifyNestedMap(map, 0);
+  return keyIterator.moveNext() ? _modifyNestedMap(map) : modify(map);
```
Awjin, line 189: Could this instead be the base case of `_modify()`?

lib/src/functions/map.dart
```diff
@@ -131,36 +152,41 @@ final _hasKey = _function("has-key", r"$map, $key, $keys...", (arguments) {
...
+    if (!keyIterator.moveNext()) {
+      mutableMap[key] = newValue ?? modify(mutableMap[key]);
       return SassMap(mutableMap);
     }
 
     var nestedMap = mutableMap[key]?.tryMap();
-    if (nestedMap == null && !overwrite) {
-      return SassMap(mutableMap);
+    if (nestedMap == null) {
+      newValue = modify(null);
```
Awjin, line 180: Might be worth documenting why we pass `null` here.

Awjin (approved):

lib/src/functions/map.dart
jathak: Do you mean making this like `if (keys.isEmpty) return modify(map)` at the top of `_modify()` or as a base case of `_modifyNestedMap()`?

For the latter, since it deals with mutating a key within a map, it doesn't really work when there are no keys at all. For the former, `keyIterator.moveNext()` needs to be called once before `_modifyNestedMap()` (to ensure that `keyIterator.current` exists), so making the check for empty keys separate doesn't really make things any simpler.

nex3 (approved):

PR #1098, opened 2020-09-27
Feature: Add support for multiline expressions in indented syntax

### LINKED ISSUE #280: Multiline in indented syntax
Over at Ruby Sass there has been an [issue](https://github.com/sass/sass/issues/216) about Multiline in indented syntax for quite some time, a [pull request](https://github.com/sass/sass/pull/2094) trying to introduce it has been closed without a merge because of the upcoming Ruby Sass deprecation in favor of Dart Sass.

Dart Sass is now here and I'm wondering whether there's any chance of restarting what was happening in that PR and maybe getting multiline into Dart Sass?

alromh87:
backslash "\\" is used to wrap long lines, it most be followed by newline and spaces after will be ignored, example:

```
/* Mixins */
@mixin col($cols, $mleft: 0, $mright: 0, $include-margin: false, $border: 0, \
           $pleft: 0, $pright: 0, $include-padding: true, $extra: 0, $clear: false, \
           $lead: true, $container: false)
    width: $cols
    color: red
    display: block
```

fixes #280
fixes sass/sass#216

All test passed
Captura de pantalla de 2020-09-27 10-43-37

### DIFF
--- a/lib/src/parse/sass.dart
+++ b/lib/src/parse/sass.dart
@@ -314,6 +314,12 @@ class SassParser extends StylesheetParser {
     // newlines.
     while (!scanner.isDone) {
       var next = scanner.peekChar();
+      if (next == $backslash) {
+        // Use backslash to enable multiline
+        scanner.readChar();
+        _expectNewline();
+        continue;
+      }
       if (next != $tab && next != $space) break;
       scanner.readChar();
     }


alromh87:
@googlebot I signed it!

jathak:
Thanks for this!

Because this involves a change to the language itself (not just Dart Sass's implementation), this change needs to go through the [language proposal process](https://github.com/sass/sass/blob/master/CONTRIBUTING.md). Since this feature is already marked as planned, you can start from step 2 on the linked document.

If you'd like to write up a proposal (done as a PR on the [sass](https://github.com/sass/sass) repo), we can review it and then prepare it for public comment.

Once a formal proposal goes through the approval process, we can revisit this PR, but for now I'm going to close it and the matching sass-spec PR.

jathak:
Unfortunately no, as as there are some design decisions that are not entirely settled. For instance, it probably makes sense to support additional heuristics for when multi-line expressions are allowed rather than just relying on escapes.

The fast-track process is usually only for minor bug fixes in the spec. Introducing new features requires the full process.

PR #1108, opened 2020-10-07
Fix custom property parsing in plain CSS

### LINKED ISSUE #1104: webpack builds using sass-loader are failing on updating to sass@1.26.12
The fix committed in #1096 is causing the webpack builds using sass-loader to fail with the following error:
```
ERROR in ./web/src/shared/themes/bootstrap/bootstrap.scss (./node_modules/css-loader/dist/cjs.js??ref--11-1!./node_modules/sass-loader/dist/cjs.js??ref--11-2!./web/src/shared/themes/bootstrap/bootstrap.scss)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
TypeError: Instance of 'SassColor0': type 'SassColor0' is not a subtype of type 'SassString0'
 @ ./web/src/shared/themes/bootstrap/bootstrap.scss 4:14-168
```

The bootstrap.scss has the following code:
```scss
@import '~bootstrap-vue/dist/bootstrap-vue';
@import '~bootstrap/dist/css/bootstrap';
@import '../../css/bootstrap-overrides';
@import "../globals/globals";
```

On reverting to sass@1.26.11 the webpack build is working fine.

Any insights on what could be the issue here would be of great help. Thanks :)

nex3:
Closes #1104
See sass/sass-spec#1575

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -503,8 +503,11 @@ abstract class StylesheetParser extends Parser {
   /// This is only used in contexts where declarations are allowed but style
   /// rules are not, such as nested declarations. Otherwise,
   /// [_declarationOrStyleRule] is used instead.
-  @protected
-  Statement _propertyOrVariableDeclaration() {
+  ///
+  /// If [parseCustomProperties] is `true`, properties that begin with `--` will
+  /// be parsed using custom property parsing rules.
+  Statement _propertyOrVariableDeclaration(
+      {bool parseCustomProperties = true}) {
     var start = scanner.state;
 
     Interpolation name;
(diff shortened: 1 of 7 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -503,8 +503,11 @@ abstract class StylesheetParser extends Parser {
   /// This is only used in contexts where declarations are allowed but style
   /// rules are not, such as nested declarations. Otherwise,
   /// [_declarationOrStyleRule] is used instead.
-  @protected
```
Awjin, line 506: Does this no longer apply?

Awjin (approved):

lib/src/parse/stylesheet.dart
nex3: I'm not sure why it was there in the first place—the method is private, so it can't be referenced from subclasses anyway (unless those subclasses are in the same file which they aren't).

Issue #1089, opened 2020-09-18
Add HWB support in sass:color


nex3:
See https://github.com/sass/sass/issues/2834.

We're currently landing provisional support for this in the [feature.hwb](https://github.com/sass/dart-sass/commits/feature.hwb) branch, including https://github.com/sass/dart-sass/pull/1085.

PR #1111, opened 2020-10-10
Support HWB in adjust, change, and scale

jathak:
This adds `$whiteness` and `$blackness` parameters to the `adjust`, `change`, and `scale` functions in the color module.

In doing so, I generalized the logic for all three functions into a single `_updateComponents` helper since they all shared a lot of similar logic for validating their parameters.

See sass/sass#2834
See #1089

See sass/sass-spec#1577

### DIFF
--- a/lib/src/functions/color.dart
+++ b/lib/src/functions/color.dart
@@ -384,165 +384,14 @@ final _complement = _function("complement", r"$color", (arguments) {
 
 // Miscellaneous
 
-final _adjust = _function("adjust", r"$color, $kwargs...", (arguments) {
-  var color = arguments[0].assertColor("color");
-  var argumentList = arguments[1] as SassArgumentList;
-  if (argumentList.asList.isNotEmpty) {
-    throw SassScriptException(
-        "Only one positional argument is allowed. All other arguments must "
-        "be passed by name.");
-  }
+final _adjust = _function("adjust", r"$color, $kwargs...",
+    (arguments) => _updateComponents(arguments, adjust: true));
 
-  var keywords = Map.of(argumentList.keywords);
-  num getInRange(String name, num min, num max) =>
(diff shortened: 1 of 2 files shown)


lib/src/functions/color.dart
```diff
@@ -447,112 +420,86 @@ final _scale = _function("scale", r"$color, $kwargs...", (arguments) {
...
 
   var hasRgb = red != null || green != null || blue != null;
-  var hasHsl = saturation != null || lightness != null;
-  if (hasRgb) {
-    if (hasHsl) {
-      throw SassScriptException(
-          "RGB parameters may not be passed along with HSL parameters.");
-    }
+  var hasSl = saturation != null || lightness != null;
+  var hasWb = whiteness != null || blackness != null;
```
nex3, line 456: According to [the Dart style guide](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words), these should be `hasSL` and `hasWB`.

lib/src/functions/color.dart
```diff
@@ -447,112 +420,86 @@ final _scale = _function("scale", r"$color, $kwargs...", (arguments) {
...
-  var alpha = getScale("alpha");
+  var alpha = getParam("alpha", 1);
+  var red = getParam("red", 255);
+  var green = getParam("green", 255);
+  var blue = getParam("blue", 255);
+  var hue = scale ? null : keywords.remove("hue")?.assertNumber("hue")?.value;
+  var saturation = getParam("saturation", 100);
+  var lightness = getParam("lightness", 100);
+  var whiteness = getParam("whiteness", 100);
+  var blackness = getParam("blackness", 100);
```
nex3, line 447: This should assert that these parameters (but not `saturation` and `lightness`) have unit `%` for `adjust()` and `change()`.

nex3 (approved):

PR #1123, opened 2020-10-21
fix(importer/node):  skip exception in tryPath

### LINKED ISSUE #1121: Importer incorrectly excludes files by filename, instead of absolute path, when using contents option
Simply put, when using `contents` instead of `file`, `Importer` will not consider the full path of the file, but only look at the filename itself. If the same filename appears in two subfolders, `Importer` will not process (perhaps because of cache) any files name that match.

Example Tree:

- /index.scss
- /folder1/
  - useAandB.scss
  - a.scss
  - b.scss
- /folder2/
  - useAandB.scss
  - a.scss
  - b.scss

Sample code:

````js
const sass = require('sass');
const path = require('path');

const rootIndex = `
@use './folder1/useAandB.scss' as folder1;
@use './folder2/useAandB.scss' as folder2;
`
const useAandB = `
@use './a.scss';
@use './b.scss';
`;

const folder1a = '.folder1a{content:"folder1a"}'
const folder1b = '.folder1b{content:"folder1b"}'
const folder2a = '.folder2a{content:"folder2a"}'
const folder2b = '.folder2b{content:"folder2b"}'

function getContents(path) {
  switch (path) {
    case '/index.scss':
      return rootIndex;
    case '/folder1/useAandB.scss':
      return useAandB;
    case '/folder2/useAandB.scss':
      return useAandB;
    case '/folder1/a.scss':
      return folder1a;
    case '/folder1/b.scss':
      return folder1b;
    case '/folder2/a.scss':
      return folder2a;
    case '/folder2/b.scss':
      return folder2b;
    default:
      throw new Error("NOT_FOUND");
  }
}

const rootPath = '/index.scss';
const locationMap = new Map();
const result = sass.renderSync({
  data: getContents(rootPath),
  outputStyle: "compressed",
  importer(url, prev) {
    let requester = locationMap.get(prev);
    if (!requester) {
      console.assert(prev === 'stdin', 'stdin');
      requester = rootPath;
    }
    const dirname = path.dirname(requester);
    const request = path.join(dirname, url);
    locationMap.set(url, request)
    console.log(requester, '=>', request);
    const contents = getContents(request);
    return {
      contents,
    };
  }
})
const actual = result.css.toString();
const expected = [
  folder1a, folder1b, folder2a, folder2b
].join('');

console.assert(actual === expected , 'CSS Match', {actual, expected}, result.stats.includedFiles);
````

Output:
````
/index.scss => /folder1/useAandB.scss
/folder1/useAandB.scss => /folder1/a.scss
/folder1/useAandB.scss => /folder1/b.scss
/index.scss => /folder2/useAandB.scss
/folder2/useAandB.scss => /folder2/a.scss
/folder2/useAandB.scss => /folder2/b.scss
Assertion failed: CSS Match {
  actual: '.folder1a{content:"folder1a"}.folder1b{content:"folder1b"}',
  expected: '.folder1a{content:"folder1a"}.folder1b{content:"folder1b"}.folder2a{content:"folder2a"}.folder2b{content:"folder2b"}'
} [
  'folder1/useAandB.scss',
  'a.scss',
  'b.scss',
  'folder2/useAandB.scss',
  // Removed Symbol
]
````

Changing `folder2` to contain `c.scss` and `d.scss`, instead, produces the correct result (4 rules instead of 2). `includedFiles` shows that the full path isn't being used.

clshortfuse:
Fixes #1121:

* Add try/catch around readFile in case file is not found.
* Return null on failure, as documented

### DIFF
--- a/lib/src/importer/node/implementation.dart
+++ b/lib/src/importer/node/implementation.dart
@@ -169,9 +169,12 @@ class NodeImporter {
     var resolved = forImport
         ? inImportRule(() => resolveImportPath(path))
         : resolveImportPath(path);
-    return resolved == null
-        ? null
-        : Tuple2(readFile(resolved), p.toUri(resolved).toString());
+    if (resolved == null) return null;
+    try {
+      return Tuple2(readFile(resolved), p.toUri(resolved).toString());
+    } catch (e) {
+      return null;
+    }
   }
 
   /// Converts an importer's return [value] to a tuple that can be returned by
--- a/test/node_api/importer_test.dart
+++ b/test/node_api/importer_test.dart
@@ -139,6 +139,43 @@ void main() {
           }))));
       expect(result.stats.includedFiles, equals(['bar']));
     });
+
+    test("uses relative files within subdirectories by contents", () {
+      expect(
+          renderSync(RenderOptions(
+              data:
+                  "@use './button/rules' as button; @use './input/rules' as input",
+              importer: [
+                allowInterop((Object url, void _) {
+                  if (url != "button/rules") return null;
+                  return NodeImporterResult(
+                      contents: '@use "./styles"',
+                      file: '/__virtual__/button/rules');
+                }),
+                allowInterop((Object url, void _) {
+                  if (url != "input/rules") return null;
+                  return NodeImporterResult(
+                      contents: '@use "./styles"',
+                      file: '/__virtual__/input/rules');
+                }),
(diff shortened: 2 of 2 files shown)


nex3:
As I described in https://github.com/sass/dart-sass/issues/1121#issuecomment-713887543, this behavior is working as intended (although certainly less than ideally).

Issue #1109, opened 2020-10-09
Any recommendations for troubleshooting poor performance


ronnyek:
Ok, so I know the `sass` npm package is the dart2js version of dart-sass, but rather than continuing to complain about performance there, I was curious if there were ways to set super details verbosity or anything so I can identify whether I'm running into other issues.

I believe when sass-loader switched to defaulting to dart-sass, the sass compile step went from like 1-2 seconds to 2+ minutes. Rather than assume all of that is because of dart runtime etc... I want to see if there are problems with my scss files.

I was initially concerned that I had circular references or something, but dont get any warnings. I believe there is a verbosity flag, but for me that just gave me more information when a file include failed.

I am really open to any other ideas. I tried to run some older node apps that attempt to detect scss "graphs", but they seem to be older and somewhat unsupported, and don't complete against my scss files. That being said, these scss files do ultimately end up compiling and rendering exactly as I expect.

nex3:
That's a *much larger* compilation time increase than expected. In [our benchmarks](https://github.com/sass/dart-sass/blob/master/perf.md) the JS version ranges from 2x-5x slower than LibSass and the Dart VM, not 60x. If you want to dig into this yourself, the best way is either to use a [Node profiler](https://nodejs.org/en/docs/guides/simple-profiling/) to figure out where time is being spent or to bisect your Sass stylesheets to try to figure out what's triggering such a dramatic performance hit.

If you're able to share your stylesheets, I could also take a look and try to figure out what's going wrong.

steveetm:
I've even worse experience. I am trying to upgrade our whole dev env including webpack, loaders, everything. Also want to try to adopt yarn pnp and moved to dart-sass for this very reason.
With node@10 I see oom issues after 10 minutes of compile time with node@14 the longest I left it run was 45 minutes.
Thats only the css, in the past with webpack@3 and sass-loader our _whole_ project compiled in 2 minutes worst case, scss was around 10 seconds.

Profile of 5 minutes of runtime attached, memory around 700 megs, increase is 2MB/s but that speed is decreasing.
I am pretty sure we have some braindead legacy extends lurking somewhere in the code, but this is just not normal.
I can't compile that scss directly with dart-sass as there is a lot of ~imports in it, but I highly doubt it is a sass-loader or webpack issue.

[...]

PR #1124, opened 2020-10-23
Trim extends at the compound selector level

nex3:
This reverts a portion of #460 (15202c4). It turns out that trimming
compound selector extends allows us to avoid exponential behavior in
certain recursive @extend cases.

See sass/dart-sass#1109
See sass/sass-spec#1578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 1.27.0
+## 1.27.1
+
+* Avoid going exponential on certain recursive `@extend` edge cases.
+
+## 1.27.1
 
 * Adds an overload to `map.merge()` that supports merging a nested map.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 1.27.0
+## 1.27.1
+
+* Avoid going exponential on certain recursive `@extend` edge cases.
```
jathak, line 3: Maybe clarify here that "going exponential" refers to the number of selectors generated? On first read, I assumed this just meant exponential in runtime.

jathak (approved):

CHANGELOG.md
nex3: It's actually the runtime that's the issue here—we do end up trimming most of the selectors, but we do so late enough in the process that we have to process an exponential combination in the meantime.

Issue #1174, opened 2020-12-29
Support all <angle> units in color functions


nex3:
See https://github.com/sass/sass/issues/2904

PR #1175, opened 2020-12-29
Deprecate incorrect HSL units

nex3:
Partially addresses #1174
See sass/sass#2904

See sass/sass-spec#1608


jathak (approved):
I think a couple of tests need to be updated to use the new error messages. Otherwise LGTM

PR #1183, opened 2021-01-06
Migrate the CI to github actions

### LINKED ISSUE #1163: CI is not working anymore
According to https://travis-ci.com/github/sass/dart-sass/requests, the sass organization does not have Travis credits anymore (and they seem to have suspended the granting of free open-source credits according to https://twitter.com/ArnaudLigny/status/1336077586888318977).

stof:
Closes #1163 

The deployment part require setting up the appropriate secrets in the github settings:

- `SASSBOT_GITHUB_TOKEN`
- `NPM_TOKEN`
- `HEROKU_TOKEN`
- `CHOCOLATEY_TOKEN`
- `PUB_CREDENTIALS`

They correspond to the existing secure env variables of the Travis config (with the same names except for `GITHUB_TOKEN` because secret names are not allowed to start with `GITHUB_`)


stof:
See https://github.com/stof/dart-sass/actions/runs/467055434 for an example of running it.

stof:
This contains some failure for the watch tests in one node job. My previous attempts included some similar failures sometimes on other node versions too. So this looks like a volatile test.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+      - name: Setup Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.6
+          working-directory: sass-spec
+          bundler-cache: true
+
+      - name: Run specs
+        run: bundle exec sass-spec.rb --dart .. $extra_args
+        working-directory: ./sass-spec
```
nex3, line 64:
```suggestion
        working-directory: sass-spec
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+        dart_channel: [stable, dev]
+        async_label: [synchronous]
+        async_args: ['']
+        include:
+          - dart_channel: stable
+            async_label: asynchronous
+            async_args: '--cmd-args --async'
+
+    steps:
+      - name: Checkout
```
nex3, line 31: Style nit: for configuration-file brevity, I prefer to leave off step names when they're redundant with the name of the action repo itself.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+      - name: Setup Dart
+        uses: cedx/setup-dart@v2
+        with:
+          release-channel: "${{ matrix.dart_channel }}"
+
+      - name: Install app dependencies
+        run: dart pub get
+
+      # Setup the environment to run the sass-spec runner which is written in Ruby
+      - name: Setup sass-spec
```
nex3, line 43:
```suggestion
      - name: Set up sass-spec
```

Fun grammar fact: "setup" is the noun form of the verb "set up"

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+          release-channel: "${{ matrix.dart_channel }}"
+
+      - name: Install app dependencies
+        run: dart pub get
+
+      - name: Build
+        run: dart pub run grinder pkg-standalone-dev
+
+      - name: Run tests
+        run: dart pub run test -p vm -x node
```
nex3, line 99:
```suggestion
        run: dart pub run test -p vm -x node -r expanded
```

This will cause the test runner to only print output when a test passes or fails, rather than printing a line every second.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+          node-version: '${{ matrix.node_version }}'
+
+      - name: Install node dependencies
+        run: npm install
+
+      - name: Build
+        run: dart pub run grinder before-test
+
+      - name: Run tests
+        run: dart pub run test -j 2 -t node
```
nex3, line 156:
```suggestion
        run: dart pub run test -j 2 -t node -r expanded
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Setup Dart
+        uses: cedx/setup-dart@v2
+
+      - name: Install app dependencies
+        run: dart pub get
```
nex3, line 170: There's so much repetition in this file, it's really unfortunate that it doesn't support YAML references :confused:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,392 @@
...
+        uses: cedx/setup-dart@v2
+        with:
+          release-channel: "${{ matrix.dart_channel }}"
+
+      - name: Install app dependencies
+        run: dart pub get
+
+      # Setup the environment to run the sass-spec runner which is written in Ruby
+      - name: Setup sass-spec
+        # TODO replace the spec-ref script
```
nex3, line 44: Why?

tool/github-action/sass-spec-ref.sh
```diff
@@ -0,0 +1,35 @@
...
+  >&2 echo "Ref: $default."
+  echo "$default"
+  exit 0
+fi
+
+>&2 echo "$PR_BODY"
+
+RE_SPEC_PR="sass/sass-spec(#|/pull/)([0-9]+)"
+
+if [[ $PR_BODY =~ $RE_SPEC_PR ]]; then
```
nex3, line 28: Very exciting to be able to do this without an HTTP request :smiley:.

should `$PR_BODY` and `$RE_SPEC_PR` be quoted here?

nex3 (commented):
Thanks for diving into this! You're certainly saving me a bunch of hassle :grin:.

nex3:
The failing tests here are strange. It's true that `--watch` tests are inherently flaky, but they're also set to retry three times on failure, and this test *continued to fail* through all three retries. No other test failed, which suggests that something's wrong with this test in particular. On the other hand, it consistently passes on Travis. Maybe the GitHub bots are slow enough that it's worth trying to increase the duration of [`tick`](https://github.com/sass/dart-sass/blob/f24a2b9ca42da11753d64daa213b35c9e8d2e083/test/utils.dart#L15-L21) for non-Windows systems?

.github/workflows/ci.yml
stof: I forgot to remove the TODO after doing it (I kept that for after other work)

.github/workflows/ci.yml
stof: I totally agree with you on that point.

tool/github-action/sass-spec-ref.sh
stof: there was no quotes around them in the previous script, which I took as basis for this one.

nex3: Doing some poking around in Bash documentation, I think `$PR_BODY` should be quoted but `$RE_SPEC_PR` should not:

```suggestion
if [[ "$PR_BODY" =~ $RE_SPEC_PR ]]; then
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,300 @@
...
+        dart_channel: [stable, dev]
+        async_label: [synchronous]
+        async_args: ['']
+        include:
+          - dart_channel: stable
+            async_label: asynchronous
+            async_args: '--cmd-args --async'
+
+    steps:
+      - name: actions/checkout@v2
```
nex3, line 31: This should be `uses:`, right? Also below.

stof: yes, indeed

nex3:
Moving to https://github.com/sass/dart-sass/pull/1187 to get the CI going.

PR #1200, opened 2021-01-12
Implement proper handling of units in for loops

stof:
Closes #1199 

https://github.com/sass/sass-spec/pull/1617

PR #1201, opened 2021-01-12
Fix the CI to use a PR version of sass-spec

stof:
The previous implementation contained 2 mistakes:

- there was a mismatch between names of env variables in the workflow config and in the script
- the `github.base_ref` property in github actions contains a branch name, not a full git ref like `github.ref`

### DIFF
--- a/tool/github-action/sass-spec-ref.sh
+++ b/tool/github-action/sass-spec-ref.sh
@@ -7,15 +7,21 @@
 # run. If we're running specs for a pull request which refers to a sass-spec
 # pull request, we'll run against the latter rather than sass-spec master.
 
-GITHUB_REF=${PR_REF:-$CURRENT_REF}
-if [[ "$GITHUB_REF" == refs/heads/feature.* ]]; then
-  default="${GITHUB_REF:11}"
+if [ -z "$PR_BRANCH" ]; then
+  # Remove the "refs/heads/" prefix
+  current_branch="${CURRENT_REF:11}"
+else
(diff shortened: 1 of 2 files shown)


stof:
I discovered this thanks to https://github.com/sass/dart-sass/pull/1200/checks?check_run_id=1686654465#step:5:2 which installed the wrong version.

Regarding env variables, I decided to use `*_BRANCH` when it contains a branch name and `*_REF` when it contains a full git ref.

tool/github-action/sass-spec-ref.sh
```diff
@@ -7,15 +7,20 @@
 # run. If we're running specs for a pull request which refers to a sass-spec
 # pull request, we'll run against the latter rather than sass-spec master.
 
-GITHUB_REF=${PR_REF:-$CURRENT_REF}
-if [[ "$GITHUB_REF" == refs/heads/feature.* ]]; then
-  default="${GITHUB_REF:11}"
+if [ -z "$PR_BRANCH" ]; then
+  current_branch="${CURRENT_REF:11}"
```
nex3, line 11: Can you add a comment indicating what string exactly the `:11` is slicing off here?

nex3 (changes requested):

tool/github-action/sass-spec-ref.sh
stof: done

nex3 (approved):
Thanks!

PR #1012, opened 2020-05-24
Update node_preamble, hopefully fixing electron support

Mstrodl:
See mrbullington/node_preamble.dart#20 for more information

PR #1207, opened 2021-01-16
Bump to 1.32.5 (updated node_preamble.dart)

Mstrodl:
Similar to #1028, pulls mbullington/node_preamble.dart#21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.32.5
+
+* Use an updated version of `node_preamble` when compiling to JS.
+
 ## 1.32.4
 
 * No user-visible changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.32.5
 
-* Use an updated version of `node_preamble` when compiling to JS.
+* Fixes Electron support when `nodeIntegration` is disabled
 
 ## 1.32.4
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.32.5
+
+* Use an updated version of `node_preamble` when compiling to JS.
```
nex3, line 3: Generally, we try to keep changelog entries focused on the user's perspective. What does the latest version of `node_preamble` allow/fix that older versions did not?

nex3 (changes requested):

Mstrodl:
@nex3 I've updated the changelog as requested

Mstrodl:
@stof See the old PR, I was told not to change it because it still fits within the semver.

Mstrodl:
Ironically however this doesn't even fix my issue because node_preamble.dart didn't actually have it's version number changed... Oops.

stof:
but if the fix requires the new version of the library, bumping the min version of the constraint would make sense to be sure that `pub` cannot select the older buggy version. Otherwise, the fix is not enforced.

Mstrodl:
@stof I agree with what you're saying, I'm just explaining why I didn't bump it (I did try the first time, see #1012 .)

stof:
well, then the issue is not actually fixed, because the fix you want to bring in is not yet released on https://pub.dev/packages/node_preamble

Mstrodl:
@stof Correct! I have opened a pull request on the node_preamble repository because I forgot: mbullington/node_preamble.dart#23

PR #1210, opened 2021-01-20
Remove useless dev dependency

stof:
source_span is already a normal dependency of the package.

this has been reported to me by my IDE (I'm using the Dart plugin for Jetbrains IDEs).


Awjin:
Good catch, thanks.

PR #1215, opened 2021-01-22
Set RangeError.start and .end for color range checks

nex3:
This is useful for sass/dart-sass-embedded#32 because it makes the
range checks easier to deconstruct and reconstruct into useful
embedded error messages.

### DIFF
--- a/lib/src/util/number.dart
+++ b/lib/src/util/number.dart
@@ -89,5 +89,6 @@ num fuzzyCheckRange(num number, num min, num max) {
 num fuzzyAssertRange(num number, num min, num max, [String name]) {
   var result = fuzzyCheckRange(number, min, max);
   if (result != null) return result;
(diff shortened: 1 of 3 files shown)


lib/src/util/number.dart
```diff
@@ -89,5 +89,6 @@ num fuzzyCheckRange(num number, num min, num max) {
 num fuzzyAssertRange(num number, num min, num max, [String name]) {
   var result = fuzzyCheckRange(number, min, max);
   if (result != null) return result;
-  throw RangeError.value(number, name, "must be between $min and $max.");
+  throw RangeError.range(
+      number, min, max, name, "must be between $min and $max");
```
Awjin, line 93: Should `minValue` and `maxValue` be nums and not ints?

Awjin (approved):

PR #1217, opened 2021-01-25
Bump version

Mstrodl:
Same as  #1028, this time the relevant [node_preamble.dart version](https://pub.dev/packages/node_preamble/versions/1.4.13) has actually been published... Whoops!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.32.6
+
+### Node JS API
+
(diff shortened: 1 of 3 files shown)


Mstrodl:
Not sure why this is failing, but other commits seem to be as well so I assume it's not an issue?

Mstrodl:
Strange. The other commits didn't have _that_ issue

PR #1221, opened 2021-02-01
Split SassNumber into separate subclasses based on units

nex3:
This allows us to use more efficient implementations in the common
case where numbers don't have complex units.

See #113


jathak (approved):

stof:
@nex3 I think it would make sense to update benchmarks in perf.md to see the impact of this change

Issue #1222, opened 2021-02-02
Spaces in @forward module configuration lead to parsing error


jhony1104:
dart-sass: 1.32.5

There seems to be an error in the parser for the `with` block of the `@formard` rule,

Using the example  from the documentation (https://sass-lang.com/documentation/at-rules/forward#configuring-modules):

```scss
@forward 'library' with (
  $black: #222 !default,
  $border-radius: 0.1rem !default
);
```

leed to the following error:
```
Error: expected ")".
  ╷
3 │   $border-radius: 0.1rem !default
  │                                  ^
  ╵
  _opinionated.scss 3:34  @use
  style.scss 1:1          root stylesheet
```

The only workaround that seems to be working, is to close the with-block brackets on the same line (witout any spaces):
```scss
@forward 'library' with (
  $black: #222 !default,
  $border-radius: 0.1rem !default);
```

Awjin:
Thanks, this looks like a bug with how compiler is parsing `!default` within the `forward` rules. Another workaround for now is to add a comma after the second `!default`.

PR #1226, opened 2021-02-04
Let last !default value of @forward...with be commaless.

Awjin:
This advances the scanner to the next line even if the !default
identifier is not followed by a comma.

Fixes https://github.com/sass/dart-sass/issues/1222
See https://github.com/sass/sass-spec/pull/1619


nex3 (approved):
Can you add a changelog entry for this and make sure the pubspec is up-to-date?

Awjin:
Sure thing, added

PR #1227, opened 2021-02-04
Limit the version of source_span

nex3:
This is a workaround to get the build working again until
dart-lang/source_span#72 lands.


Awjin (commented):
Thanks!

PR #1235, opened 2021-02-18
Update chokidar.

Awjin:


nex3 (approved):
Can you add a CHANGELOG comment for this, and bump the pubspec as well if necessary?

PR #1237, opened 2021-02-18
Update CI and CONTRIBUTING to use Node spec runner

jathak:


nex3 (approved):
Thanks for thinking of this!

PR #1238, opened 2021-02-23
Allow the null safety release of args and watcher

lukaszkolodziejczyk:


lukaszkolodziejczyk:
@kevmoo

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## 1.32.8
 
+* Allow the null safety release of args and watcher.
```
nex3, line 3: 1.32.8 has already been released, so this should go in a new 1.32.9 section and the pubspec version should be bumped to `1.32.9-dev`. Also, please put this under a `### Dart API` header since it's Dart-specific.

nex3 (changes requested):

CHANGELOG.md
lukaszkolodziejczyk: Done. Thanks!

PR #1242, opened 2021-03-01
Use the official setup-dart action

stof:
https://github.com/cedx/setup-dart has been abandoned in favor of the official setup-dart action maintained by the dart team now that it exists.


Awjin:
Thank you for keeping an eye out for this!

PR #1249, opened 2021-03-08
Opt dynamically spawned isolates out of null safety

mraleph:


stof:
@jathak should this PR be resurrected and merged, to fix the node tests regarding #1254 ?

mraleph:
@stof this PR should not be needed anymore. The issue was fixed in the `package:test` (see https://github.com/dart-lang/test/issues/1363#issuecomment-794316840) so you should just roll `package:test` forward to [1.16.7](https://pub.dev/packages/test/changelog#1167) or newer.

stof:
OK, I found out why test 1.16.5 gets installed rather than the latest version. dart-sass currently depends on yaml 2.x while test now requires the 3.x version. I'll look into updating the yaml dependency.

stof:
@mraleph the issue is that the 1.16.6 release of test bumped the min version of its dependencies, allowing only the new null-safe major versions. This makes it impossible to use the fix from 1.16.7 until cli_pkg and grinder are updated, and until dart-sass stops using `package_resolver` (see my comment on https://github.com/google/dart_cli_pkg/pull/77#issuecomment-801110317).

so having a workaround to have a working CI in the meantime is a good idea.

PR #1253, opened 2021-03-09
Migrate to null safety (WIP)

jathak:


lib/sass.dart
```diff
@@ -90,7 +90,7 @@ String compile(String path,
     Logger logger,
     Iterable<Importer> importers,
     Iterable<String> loadPaths,
-    SyncPackageResolver packageResolver,
+    PackageConfig packageConfig,
```
stof, line 93: is this a BC break that requires releasing it as dart-sass 2.0.0 ?

PR #1256, opened 2021-03-12
allow debug before use/forward

JoshMcCullough:


jathak:
Thanks for this! This seems like a reasonable addition to the language, but before working on an implementation, we'd first need to write a formal proposal for the change. Would you be interested in drafting one? The instructions are [here](https://github.com/sass/sass/blob/master/CONTRIBUTING.md#proposal).

I'm going to close this PR for now, but we can re-open it once we have an proposal accepted for the language change.

Issue #1254, opened 2021-03-10
CI is broken since the Dart 2.12 stable release


stof:
There are 2 things breaking:

- the node.js tests are apparently trying to spawn an isolate with sound null safety turned on, which fails as null safety is not yet ready (#1253 is working on implementing null safety)
- the analyzer jobs fails because the new version has some new checks

jathak:
The Node tests issue is due to a bug in the Dart SDK itself. I believe they're working on a fix, but we'll likely complete the null safety migration before that's done.

If there's an urgent need to cut a release before we finish the null safety migration, we can pin the CI to Dart 2.11, but otherwise, we can just wait to merge other PRs until #1253 is done.

stof:
@jathak I see that #1253 has been closed without merging it. So what is the state of this work ?

Awjin:
@stof We're still working on the nnbd migration, but in a different branch.

PR #1261, opened 2021-03-17
Fix static analysis

stof:
Refs #1254 

This fixes all errors reported in the static_analysis job when using the new pedantic version (that now gets installed on CI) and the SDK 2.12


stof:
For reference, here is a CI run with the failed static analysis, to see the list of what gets fixed: https://github.com/sass/dart-sass/runs/2099661492?check_suite_focus=true

Awjin:
Thanks for the cleanup. However, it might make sense to hold off on merging these, as these fixes should be included in the nnbd migration.

nex3:
As Awjin mentioned, these were fixed in the NNBD migration so I'm going to close this out. Thanks for the thought, though!

Issue #863, opened 2019-10-27
Lazy load chokidar dependency


eklingen:
Is it possible to make dart-sass load chokidar only when actually watching files?

In trying to optimise my build chain, I've noticed dart-sass loads the chokidar dependency, even when the watch functionality isn't actually used. While the absolute timings are small (though depending on cpu, I/O, etc), chokidar is relatively heavy, since it accounts for ~75% of the time it takes to load dart-sass. At least, in my tests.

```
   55 ms        │   ├─ node_modules/sass/sass.dart.js (2x)
   43 ms        │   │   └─ node_modules/chokidar/index.js (10x)
    4 ms        │   │       ├─ node_modules/readdirp/index.js (2x)
    4 ms        │   │       │   └─ node_modules/picomatch/index.js
   18 ms        │   │       ├─ node_modules/chokidar/node_modules/braces/index.js (4x)
    4 ms        │   │       │   └─ node_modules/chokidar/node_modules/braces/lib/compile.js (2x)
    9 ms        │   │       └─ node_modules/chokidar/lib/fsevents-handler.js (2x)
    8 ms        │   │           └─ node_modules/fsevents/fsevents.js
```

I'm not that well versed in Dart. But i'm willing to take a crack at it.

nex3:
My understanding is that lazy-loading dependencies works poorly with WebPack's bundling (https://github.com/sass/dart-sass/issues/726). If we can find a way to resolve these two issues, I'm open to lazily loading Chokidar again.

DRoet:
Might be worth upgrading to chokidar `3.x` since it is about an 17x package & deps size reduction compared to `2.x`. so it might load faster

tchetwin:
Looks like support for 3.x has been in place since https://github.com/sass/dart-sass/pull/731, corresponding I think to version [v1.22.1](https://github.com/sass/dart-sass/releases/tag/1.22.1):

https://github.com/sass/dart-sass/blob/d155f3518e92a1b41122f0f2e3b22572329c35c6/package/package.json#L19-L21

dl748:
I don't think its being honored then, because it only installs the latest version of 2... 2.1.8......

Fresh install shows
```
├─┬ sass@1.32.7
│ └─┬ chokidar@2.1.8
```

changing the package.json of sass to ">=3.0.0 <4.0.0" correctly installs the latest version

silviuburceadev:
[...]

PR #1279, opened 2021-04-06
Limit Chokidar's lower bound to 3.0.0

nex3:
There's no reason we *should* have to do this, since we do declare
support for Chokidar 3.0.0 and npm *should* just install the latest
available version that matches the constraint. But in practice it
doesn't for some reason, so I guess we'll narrow the constraint to
work around that.

See #863


Awjin (approved):

silviuburceadev:
I can't be more happy now that chokidar 2.x doesn't block me upgrading to Node.js 14.x

PR #1283, opened 2021-04-12
Enable strict null checks

nex3:
See sass/sass-spec#1632


jathak (approved):
Overall LGTM, except for one `return true` that I think should be false.

We should convert any `andGet` calls to `?[]` at some point, but we could probably save that for a separate PR.

pubspec.yaml
```diff
@@ -9,40 +9,41 @@ executables:
...
+  source_maps: ^0.10.10
+  source_span: ^1.8.1
+  stack_trace: ^1.10.0
+  stream_transform: ^2.0.0
+  string_scanner: ^1.1.0
+  term_glyph: ^1.2.0
+  tuple: ^2.0.0-nullsafety.0
+  watcher: ^1.0.0
+
+publish_to: none
```
jathak, line 33: This needs to be removed before publishing. I think prerelease versions for dependencies also need to be removed to publish a stable version.

Issue #1295, opened 2021-04-21
@use - The target selector was not found.


larsboldt:
``` 
colors.scss
%background-white {
  background-color: white;
}

foo.scss
@use "colors";

.foo {
  @extend %background-white;
  padding: 20px;
}

bar.scss
@use "foo";

.bar {
  @extend .foo;
  border: 1px solid black;
}

baz.scss
@use "foo";

.baz {
  @extend .foo;
  border: 1px solid green;
}

total.scss
@use "bar";
@use "baz";

sass total.scss total.css

Error: The target selector was not found.
Use "@extend %background-white !optional" to avoid this error.
  ╷
4 │   @extend %background-white;
  │   ^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  foo.scss 4:3  root stylesheet
```

Can anyone explain why this is happening?

kerryj89:
Hi @Awjin, while the original issue is fixed, if I have an `@extend` within a `@mixin`, it fails with the same error.

Works now:
```scss
.foo {
  @extend %background-white;
  padding: 20px;
}
```

Doesn't work:
```scss
@mixin test {
  .foo {
    @extend %background-white;
    padding: 20px;
  }
}
```

Do you have any advice here?  I will convert the silent class to be a mixin for now since I don't use the `@extend` much, but it would be nice to know if something in the spec has changed with `@extend` behavior in SASS Module System or if it's a bug.   Thanks!

nex3:
@kerryj89 Can you supply a _full, stand-alone_ reproduction of the error?

kerryj89:
@nex3 Is this what you're requesting? https://github.com/kerryj89/dart-sass-issue-3322-extend-in-mixin-undefined

nex3:
@kerryj89 That's failing because you never actually _include_ the mixin.

eminaltan:
Use @import at-rule instead @use at-rule. 

According to sass offical site on https://sass-lang.com/documentation/at-rules/import/

"The Sass team discourages the continued use of the @import rule. Sass will [gradually phase it out](https://github.com/sass/sass/blob/main/accepted/module-system.md#timeline) over the next few years, and eventually remove it from the language entirely. Prefer the [@use rule](https://sass-lang.com/documentation/at-rules/use) instead. **(Note that only Dart Sass currently supports @use. Users of other implementations must use the @import rule instead.)"**

PR #1298, opened 2021-04-27
Fix extending selectors across multiple modules.

Awjin:
Allows more than one module to extend the same upstream selector that,
in turn, extends a selector from its upstream module.

Properly skips creating an extra extension that was being counted as
unresolved when evaluating @extends.

See https://github.com/sass/sass-spec/pull/1635
Fixes https://github.com/sass/dart-sass/issues/1295

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.32.11
+version: 1.32.12-dev
 description: A Sass implementation in Dart.
 author: Sass Team
 homepage: https://github.com/sass/dart-sass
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.32.11
+version: 1.32.12-dev
```
nex3, line 2: Do we have more changes in the pipeline? If not, might as well just release this as 1.32.12.

nex3 (approved):

pubspec.yaml
Awjin: Makes sense, I'll just release this.

Issue #663, opened 2019-04-30
Stop using / as division


nex3:
* [ ] Deprecation, `divide()`, and `slash-list()`
* [ ] Removal of /-as-division, deprecation of `slash-list()`
* [ ] Removal of `slash-list()`

See https://github.com/sass/sass/issues/2565

tysonmatanich:
Is there a perf or other difference between using `calc(9 / 16)` vs `math.div(9, 16)`? I read through the docs and couldn't figure out the reason for using one over the other. Thanks!

jathak:
For division on literals, both should be equivalent (there should be minimal, if any, performance differences), but in more complicated cases, Sass may not be able to fully simplify a calc at compile time (e.g. `40px + 10%`), so it will be emitted in the final CSS, whereas `math.div` will always return a number.

PR #1306, opened 2021-05-12
 Deprecate /-as-division and add replacements

nex3:
See #663
See sass/sass-spec#1399

### DIFF
--- a/lib/src/functions/color.dart
+++ b/lib/src/functions/color.dart
@@ -740,7 +740,7 @@ Object /* SassString | List<Value> */ _parseChannels(
   var list = channels.asList;
   if (list.length > 3) {
     throw SassScriptException("Only 3 elements allowed, but ${list.length} "
-        "${pluralize('was', list.length, plural: 'were')} passed.");
+        "were passed.");
   } else if (list.length < 3) {
     if (list.any((value) => value.isVar) ||
         (list.isNotEmpty && _isVarSlash(list.last))) {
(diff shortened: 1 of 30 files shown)


lib/src/functions/color.dart
```diff
@@ -720,18 +739,20 @@ Object /* SassString | List<Value> */ _parseChannels(
 
   var list = channels.asList;
   if (list.length > 3) {
-    throw SassScriptException(
-        "Only 3 elements allowed, but ${list.length} were passed.");
+    throw SassScriptException("Only 3 elements allowed, but ${list.length} "
+        "${pluralize('was', list.length, plural: 'were')} passed.");
```
jathak, line 743: `pluralize` is unnecessary here, since we just checked that `list.length > 3`

test/dart_api/value/list_test.dart
```diff
@@ -116,6 +116,11 @@ void main() {
     });
   });
 
+  test("a slash-separated list is space-separated", () {
```
jathak, line 119:
```suggestion
  test("a slash-separated list is slash-separated", () {
```

lib/src/visitor/async_evaluate.dart
```diff
@@ -2884,22 +2925,24 @@ class _EvaluateVisitor
 
   /// Returns the [AstNode] whose span should be used for [expression].
   ///
-  /// If [expression] is a variable reference, [AstNode]'s span will be the span
-  /// where that variable was originally declared. Otherwise, this will just
-  /// return [expression].
+  /// If [expression] is a variable reference and [_sourceMap] is `true`,
```
jathak, line 2928: Isn't `_sourceMap` no longer relevant for this function? (at least until the /-as-division warnings are gone)

jathak (approved):

PR #1309, opened 2021-05-14
Add the ability for importers to detect @import

nex3:
See sass/sass#3055
See webpack-contrib/sass-loader#905

### DIFF
--- a/lib/src/importer/utils.dart
+++ b/lib/src/importer/utils.dart
@@ -2,6 +2,8 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import 'dart:async';
+
 import 'package:path/path.dart' as p;
 
 import '../io.dart';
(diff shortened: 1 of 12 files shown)


lib/src/importer/utils.dart
```diff
@@ -13,30 +15,12 @@ import '../io.dart';
...
-  try {
-    return await callback();
-  } finally {
-    _inImportRule = wasInImportRule;
-  }
-}
+/// Runs [callback] in a context where [inImportRule] returns `true` and
+/// [resolveImportPath] uses `@import` semantics rather than `@use` semantics.
+T inImportRule<T>(T callback()) =>
+    runZoned(callback, zoneValues: {#_inImportRule: true});
```
jathak, line 23: Is there a particular reason why this is now run in a zone, or is it just a cleaner way to update `fromImport`?

jathak (approved):

lib/src/importer/utils.dart
nex3: Making the variable Zone-scoped means that, if multiple Sass compilations are running in parallel, they won't stomp on one another's values for this variable. It wasn't necessary before because this was only used by `FileSystemImporter` which was synchronous, but user-defined importers aren't necessarily.

PR #1312, opened 2021-05-18
Fix some edge cases when serializing edge-case numbers

nex3:
See sass/sass-spec#1653

### DIFF
--- a/lib/src/visitor/serialize.dart
+++ b/lib/src/visitor/serialize.dart
@@ -700,7 +700,7 @@ class _SerializeVisitor
       return;
     }
 
-    _writeDecimal(text);
+    _writeRounded(text);
   }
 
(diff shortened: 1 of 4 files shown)


lib/src/visitor/serialize.dart
```diff
@@ -759,59 +761,108 @@ class _SerializeVisitor
   /// Assuming [text] is a double written without exponent notation, writes it
   /// to [_buffer] with at most [SassNumber.precision] digits after the decimal.
   void _writeDecimal(String text) {
```
Awjin, line 763: Great docs throughout, they are super helpful.

Awjin (approved):
The changes LGTM. It looks like some tests are failing due to the slash-as-div deprecation warnings.

PR #1322, opened 2021-05-21
Only emit each warning once per source location

nex3:
See https://github.com/sass/sass-spec/pull/1656

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -148,6 +148,13 @@ class _EvaluateVisitor
   /// The logger to use to print warnings.
   final Logger _logger;
 
+  /// A set of message/location pairs for warnings that have been emitted via
+  /// [_warn].
+  ///
+  /// We only want to emit one warning per location, to avoid blowing up users'
+  /// consoles with redundant warnings.
+  final _warningsEmitted = <Tuple2<String, SourceSpan>>{};
+
   /// Whether to track source map information.
   final bool _sourceMap;
 
@@ -1936,7 +1943,7 @@ class _EvaluateVisitor
     }
 
     if (node.isGlobal && !_environment.globalVariableExists(node.name)) {
-      _logger.warn(
+      _warn(
           _environment.atRoot
               ? "As of Dart Sass 2.0.0, !global assignments won't be able to\n"
                   "declare new variables. Since this assignment is at the root "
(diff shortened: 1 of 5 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -1987,9 +1993,13 @@ class _EvaluateVisitor
...
         await _addExceptionSpanAsync(node, () => node.expression.accept(this));
-    _logger.warn(
-        value is SassString ? value.text : _serialize(value, node.expression),
-        trace: _stackTrace(node.span));
+    var message =
+        value is SassString ? value.text : _serialize(value, node.expression);
+
+    // Don't use [_warn] because we don't want to pass the span to the logger.
+    if (_warningsEmitted.add(Tuple2(message, node.span))) {
+      _warn(message, node.span);
```
Awjin, line 2001: Should this be `_logger.warn()`?

lib/src/visitor/async_evaluate.dart
```diff
@@ -1987,9 +1993,13 @@ class _EvaluateVisitor
...
     var value =
         await _addExceptionSpanAsync(node, () => node.expression.accept(this));
-    _logger.warn(
-        value is SassString ? value.text : _serialize(value, node.expression),
-        trace: _stackTrace(node.span));
+    var message =
+        value is SassString ? value.text : _serialize(value, node.expression);
+
+    // Don't use [_warn] because we don't want to pass the span to the logger.
+    if (_warningsEmitted.add(Tuple2(message, node.span))) {
```
stof, line 2000: should `@warn` be deduplicated too ?

For warnings triggered in dart code, the span is the place that you actually need to change to resolve the warning (at least in most cases).
For userland functions triggering a warning with `@warn` when being misused, the span used here would be the location of the `@warn` rule inside the function, not the location of the caller (which is what where the change would have to be done).

lib/src/visitor/async_evaluate.dart
stof: the comment above makes me thing it should

lib/src/visitor/async_evaluate.dart
nex3: Hmm, that's a good point. We don't necessarily have a good understanding of where the boundary is between "user code" and "library code" such that we can reliably tell which stack frame has the span that should actually be used for uniqueness here. I'll drop the `@warn` de-duplication.

Awjin (approved):

PR #1327, opened 2021-05-22
Only omit 5 deprecation warnings per feature by default

### LINKED ISSUE #1323: Limit the amount of printed warnings/errors
In https://github.com/sass/dart-sass/issues/1319, an overwhelming number of deprecation warnings were printed to the terminal, forcing the user to end the process.

We should add a ceiling to the number of warnings that get printed to the terminal—100?—that, when surpassed, just prints the remaining count. Something along the lines of:

```bash
Warning: This is warning #1.

Warning: This is warning #2.

...

Warning: This is warning #${ceiling}.

${remaining count} more warnings.
```

nex3:
Closes #1323
See sass/sass-spec#1657

### DIFF
--- a/lib/src/logger/terse.dart
+++ b/lib/src/logger/terse.dart
@@ -1,0 +1,58 @@
+// Copyright 2018 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:collection/collection.dart';
+import 'package:source_span/source_span.dart';
+import 'package:stack_trace/stack_trace.dart';
+
+import '../logger.dart';
+
+/// The maximum number of repetitions of the same warning [TerseLogger] will
+/// emit before hiding the rest.
+const _maxRepetitions = 5;
+
+/// A logger that wraps an inner logger to omit repeated deprecation warnings.
+///
+/// A warning is considered "repeated" if the first paragraph is the same as
+/// another warning that's already been emitted.
+class TerseLogger implements Logger {
+  /// A map from the first paragraph of a warning to the number of times this
+  /// logger has emitted a warning with that line.
+  final _warningCounts = <String, int>{};
+
+  final Logger _inner;
+
+  TerseLogger(this._inner);
+
+  void warn(String message,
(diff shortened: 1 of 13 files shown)


lib/src/logger/terse.dart
```diff
@@ -0,0 +1,58 @@
...
+  final _warningCounts = <String, int>{};
+
+  final Logger _inner;
+
+  TerseLogger(this._inner);
+
+  void warn(String message,
+      {FileSpan? span, Trace? trace, bool deprecation = false}) {
+    if (deprecation) {
+      var firstParagraph = message.split("\n\n").first;
```
Awjin, line 31: My only concern is that we're now assigning special semantics to `\n\n` in a warning, which could cause confusion. But, I'm not sure if any other heuristic would make sense here.

Awjin (approved):

lib/src/logger/terse.dart
nex3: We could theoretically try to pass in some sort of side-channel "deprecation type" identifier, maybe through Zones, but I think that would be more complicated than would be necessary. Given that we control both this heuristic *and* the messages that it applies to, I think it's fine.

lib/src/logger/terse.dart
```diff
@@ -0,0 +1,58 @@
...
+    _inner.warn(message, span: span, trace: trace, deprecation: deprecation);
+  }
+
+  void debug(String message, SourceSpan span) => _inner.debug(message, span);
+
+  /// Prints a warning indicating the number of deprecation warnings that were
+  /// omitted.
+  ///
+  /// The [node] flag indicates whether this is running in Node.js mode, in
+  /// which case it doesn't mention "verbose mode" because the Node API doesn't
```
stof, line 46: The Node API takes an option object. Adding support for a `verbose` option in it would be easy, and would allow getting all deprecation warnings for all users using the npm package.

nex3: True, but adding a new feature (even a small one) to the Node.js API is a much more involved process.

stof: this one seems easy to add: https://github.com/sass/dart-sass/pull/1332

Issue #672, opened 2019-05-08
Add a --quiet-upstream flag


nex3:
Most users only have the capacity to fix deprecation warnings in their own code, but they may end up using libraries that have their own deprecation warnings. We should add a `--quiet-upstream` flag (and similar flags for the Dart and JS APIs) to silence deprecation warnings emitted by upstream libraries (that is, libraries loaded through load paths and `package:` URLs). 

I think this flag should probably *not* silence warnings due to `@warn`, since libraries frequently use those to indicate that downstream users are using the library in an invalid or deprecated way.

vladimiry:
Since 1.33.0 I stared to see a lot of deprecated division operation related warnings. Is there a workaround to disable the warnings? The module used via sass-loader.

huineng:
a zillion messages

nex3:
Prioritizing this because of all the warnings generated by the /-as-division deprecation.

nex3:
Renaming this to `--quiet-deps` because A) it's shorter and B) some users may not understand what "upstream" refers to.

nex3:
To be clear, though: this option is only available via the Sass CLI and the Dart API. It would be much harder to add to the Node API, because Node's importer model makes it much harder to track what counts as the "main package" and what counts as a "dependency". We'll be able to support in the new JS API (sass/sass#3056) and in the [embedded host](https://www.npmjs.com/package/sass-embedded), but those will take some time.

In the meantime, Node users will still see fewer deprecation warnings in the next release due to #1322, which skips warnings that have already been emitted for a given source location. That should help mitigate a substantial amount of the noise.

huineng:
No bad, i which more packages has deprecation warnings like this 

```
DEPRECATION WARNING: Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($px, $carbon--base-font-size)

More info and automated migrator: https://sass-lang.com/d/slash-div

   ╷
25 │   @return ($px / $carbon--base-font-size) * 1rem;
   │            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
    node_modules/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/_convert.scss 25:12     carbon--rem()
    node_modules/carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/layout/_breakpoint.scss 16:23  @import
[...]

PR #1332, opened 2021-05-25
Add the verbose option in the node API

stof:
This allows getting the full list of deprecation warnings when using the npm package.


stof:
I haven't added tests for it, because I did not found any test for it for the dart API either, only for the CLI.

stof:
Note that having the verbose option in the node API seems quite useful to me due to the lack of `quietDeps` (which cannot be implemented right now, as indicated in https://github.com/sass/dart-sass/issues/672#issuecomment-846311746). The 5 displayed warnings might come from dependencies and hide the ones coming from your own code.

nex3:
Unfortunately, modifying the JS API isn't so simple. Because this API is shared across multiple implementations, it needs to go through a proposal process in [the language repo](https://github.com/sass/sass).

stof:
now done in #1353

PR #1340, opened 2021-06-02
Fix the allowed children for the for at-rule in declarations

### LINKED ISSUE #1337: Clarification about the supported statements inside a for loop in a at-rule
I'm currently rewriting the parser of scssphp from scratch, by porting the dart-sass one (this will solve lots of issues we have with cases where our existing parser is not compliant). While doing so, I found a weird thing in the dart-sass parser, which I think might be a bug rather than an intended feature.

When parsing at rules inside property declarations, the dart-sass stylesheet parser uses a different `child` callback for `@for` than for other at rules (`@each`, `@if` and `@while`). It will only allow other at rules as child instead of also allowing property or variable declarations. Is it actually intended or should it use  `_declarationChild` like the others ?

https://github.com/sass/dart-sass/blob/b098adb289cee3a16ac265e6cc69e1afcef10802/lib/src/parse/stylesheet.dart#L677-L685

stof:
Closes #1337.

See https://github.com/sass/sass-spec/pull/1665.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 * Fix the URL for the `@-moz-document` deprecation message.
 
+* Fix a bug with `@for` loops nested inside property declarations.
+
 ## 1.34.0
 
 * Don't emit the same warning in the same location multiple times.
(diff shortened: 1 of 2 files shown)


jathak (approved):
Thanks!

Issue #633, opened 2019-03-22
Add strictness flags


nex3:
A number of users have requested the ability to opt in to stricter Sass compilation, in which code that would normally just causes a warning (especially a deprecation warning) instead produces a compile error. We should add these flags for all current deprecation warnings to the CLI and both APIs, and possibly add configuration file support so that the can be consistently configured for teams.

TypeScript's [compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) provide some prior art here. I suspect we *don't* want to add a general `--strict` flag that enables all strict checks, though, since this will mean some downstream users will break when we add new deprecations.

- [x] CLI support
- [x] Dart API support
- [ ] JS API support
- [ ] Documentation

jathak:
As part of this, there should also be an option to error on `@warn` rules.

Ariane-B:
So when can we hope to see this? I'm guessing not soon, since the proposal doesn't seem to be entirely decided upon?

jathak:
We have many things we want to work on and while this is one of them, we don't currently have an ETA for this. We'll add an update here when we have news to share.

jjonescz:
> As part of this, there should also be an option to error on `@warn` rules.

Note: for failing on warnings in CI, I'm using the logger option like this (not ideal, but better than nothing):

```js
logger: process.env.CI
  ? undefined
  : {
    warn: (message: string): void => {
      throw new Error(`Warning from SASS reported: ${message}`);
    },
  },
```

jathak:
Forgot to close this issue, but this is now fully released. You can use the [`--fatal-deprecation` flag](https://sass-lang.com/documentation/cli/dart-sass/#fatal-deprecation) on the CLI or the [`fatalDeprecations` option](https://sass-lang.com/documentation/js-api/interfaces/options/#fatalDeprecations) in the JS API to make a particular deprecation type an error, or pass in a Dart Sass version to make all deprecations active in that version errors.

There's also `--silence-deprecation`/`silenceDeprecations` to silence a particular deprecation type.

u01jmg3:
Thanks for implementing this feature, @jathak

Is there any way to opt-in so that all deprecations are treated as fatal without knowing their ID (or the Sass version) ahead of time? I'd like my CI to fail and alert me as soon as a new deprecation warning is present in my build.

nex3:
[...]

PR #1346, opened 2021-06-04
Add `--fatal-deprecation` flag

jathak:
Partial fix for #633.

This restructures deprecation warnings into a class that gives each deprecation an identifier and structures the warnings themselves primarily based on the instance variables of that Deprecation.

This supports the core use case of making a deprecation fatal from the command line, but I'd like some feedback on the general approach before finishing this PR (mainly adding support to the Dart API and updating any sass-specs with warnings that were tweaked as part of this).

My main questions are:
- Do the deprecations and their identifiers make sense here? I looked for all the deprecation warnings I could find being emitted but it's possible I missed one. I also made the [color units breaking change](https://sass-lang.com/documentation/breaking-changes/color-units) into two deprecations here (requiring deg for hue as one, and requiring % for saturation and lightness as the other) since the warnings were fairly different.
- The properties of `Deprecation` and the parameters for `Deprecation.message` were mostly just based on what was necessary to replicate the existing warnings in a structured way. Are there any that don't make sense, or additional properties that should be added? Also, are there any deprecations without a `removedIn` version that should have one?
- Does the Zone-based approach for setting fatal deprecations make sense here? What about updating `withWarnCallback` to include an `error ` function to allow `Deprecation.warnOrError` to work? I'm more confident in the former, but I'm less sure about the latter.
- Should the Dart API take in deprecation identifiers as strings (like the `--fatal-deprecation` flag), or should `Deprecation` be exposed as part of the public API?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.35.0
+
+* Add a `--fatal-deprecation` flag, which errors when the given deprecated
+  behavior is encountered.
+
 ## 1.34.1
 
 * Fix a bug where `--update` would always compile any file that depends on a
--- a/bin/sass.dart
+++ b/bin/sass.dart
@@ -16,6 +16,7 @@ import 'package:sass/src/executable/watch.dart';
 import 'package:sass/src/import_cache.dart';
 import 'package:sass/src/io.dart';
 import 'package:sass/src/stylesheet_graph.dart';
+import 'package:sass/src/warn.dart';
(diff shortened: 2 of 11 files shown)


ntkme:
@jathak I think we can close this as it's already delivered in a different PR?

stof:
@jathak @nex3 this should be closed as https://github.com/sass/dart-sass/pull/1820 has been merged for that instead.

Issue #1347, opened 2021-06-07
@at-root breaks with @use above it


kerryj89:
I don't know if this is a bug or some intentional language design gotcha I'm not aware of.

Anyway, I migrated our styles to use `math.div()`.  That worked great, or so I thought...

Apparently `@at-root` breaks when used with `@use` above it:

```scss
//styles.scss

.namespace {
    @import "foobar";
}
```

```scss
//_foobar.scss

@use "sass:math";

@at-root {
    .foo {
        width: math.div(6px, 2);
    }
}

.bar {
    width: math.div(6px, 2);
}
```

If transpiles into the following:

```scss
// styles.css

.namespace .foo {
    width: 3px;
}
.namespace .bar {
    width: 3px;
}
```

I expected the following:

```scss
// styles.css

.foo {
    width: 3px;
}

.namespace .bar {
    width: 3px;
}
```

I tried the following:
- Using `@include meta.load-css("foobar")` instead of `@import`
- Wrapping `@at-root` within a mixin and calling it immediately
- Move `@use` below `@at-root` (errors out as expected, but thought I'd try)

--

If I remove `@use` and just use the division symbol like previously it outputs as expected.

jathak:
@kerryj89: Wanted to give you an update on this. This behavior is definitely a bug, but it's also very difficult to fix, so we've been having trouble getting it to work.

In the meantime, I wanted to offer a workaround that will allow you to use `math.div` without breaking `@at-root`:
```scss
// styles.scss
.namespace {
    @import "foobar";
}

//_foobar.scss
@import "division";

@at-root {
    .foo {
        width: math-div(6px, 2);
    }
}

.bar {
    width: math-div(6px, 2);
}

// _division.scss
@forward "sass:math" as math-* show math-div;
```

kerryj89:
Nifty!  Thank you so much for showing me a workaround.

kerryj89:
Working as intended.  Thanks for all you do!

jathak:
Reopening this since we never actually fixed it in the general case. We've fixed this for `@use` rules of built-in modules, but were unable to find a solution for other `@use` rules.

Marking this as "help wanted" in case someone else can find a solution, but it's not something we currently plan to work on more ourselves, since this is behavior that's not supported when fully using the new module system, so fixing such a complex bug that only manifests in code that's already partially migrated isn't something that we have the bandwidth for.

ext:
[...]

PR #1348, opened 2021-06-07
Fix a bug with implicit dependencies

jathak:
Encountered this while migrating internal code to math.div

See https://github.com/sass/sass-spec/pull/1666.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.34.2
+version: 1.34.2-dev
 description: A Sass implementation in Dart.
 author: Sass Team
 homepage: https://github.com/sass/dart-sass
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.34.1
+version: 1.34.2
```
nex3, line 2: Should we mark this as a dev version until #1347 is fixed as well?

nex3 (approved):

pubspec.yaml
jathak: Yeah, that makes sense. There's a decent chance the other bug is also causing issues with the internal migration; I just haven't noticed it because I've been focused on this one

PR #1352, opened 2021-06-11
Another implicit dependency fix

jathak:
I haven't yet worked out a fix for the at-root issue, but this fix is actually necessary for the internal division migration.

This splits out `_globalModules` into two fields: one that contains `@use as *` global modules and one that contains modules that come from `@import`-ing a `@forward`. 

See https://github.com/sass/sass-spec/pull/1667.

### DIFF
--- a/lib/src/async_environment.dart
+++ b/lib/src/async_environment.dart
@@ -43,23 +43,21 @@ class AsyncEnvironment {
   /// modules were originally loaded.
   final Map<String, AstNode> _namespaceNodes;
 
-  /// The namespaceless modules used in the current scope.
-  final Set<Module> _globalModules;
-
-  /// A map from modules in [_globalModules] to the nodes whose spans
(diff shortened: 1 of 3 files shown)


lib/src/async_environment.dart
```diff
@@ -837,7 +809,7 @@ class AsyncEnvironment {
   Module toModule(CssStylesheet css, ExtensionStore extensionStore) {
     assert(atRoot);
     return _EnvironmentModule(this, css, extensionStore,
-        forwarded: _forwardedModules);
+        forwarded: _forwardedModules?.keys.toSet());
```
nex3, line 812: It would probably be more efficient to use a `MapKeySet` here, unless there's a reason it *shouldn't* update if `_forwardedModules` is modified.

nex3 (approved):

PR #1353, opened 2021-06-12
Add quietDeps and verbose to the JS API

nex3:
To support this, we now run Node-Sass-style relative loads outside of
the Node importer. This allows the evaluator to check whether a
relative load succeeded and use that to determine whether the
stylesheet counts as a dependency.

See sass/sass#3065
See sass/sass#3078


Awjin (approved):

lib/src/node.dart
```diff
@@ -106,6 +106,8 @@ Future<RenderResult> _renderAsync(RenderOptions options) async {
         indentWidth: _parseIndentWidth(options.indentWidth),
         lineFeed: _parseLineFeed(options.linefeed),
         url: file == null ? 'stdin' : p.toUri(file).toString(),
+        quietDeps: options.quietDeps ?? false,
```
stof, line 109: shouldn't this use the `isTruthy` helper as done for other options ?

PR #1355, opened 2021-06-12
Add imported URIs getter to ImportCache

irundaia:
Hi,

I was looking to implement a feature in dart-sass-embedded, to expose all imported URIs (see https://github.com/sass/embedded-protocol/issues/62).

However, to do this using the current api's, I'd have to first setup a `StylesheetGraph` including an `ImportCache`, then use the graph to load the to-be-compiled file, and then compile the file as normal. 
I think it would be easier, to expose a getter on the `ImportCache` class to expose all imported canonical URIs.

### DIFF
--- a/lib/src/import_cache.dart
+++ b/lib/src/import_cache.dart
@@ -46,6 +46,9 @@ class ImportCache {
   /// The import results for each canonicalized import URL.
   final Map<Uri, ImporterResult> _resultsCache;
 
+  /// All imported canonical URIs.
+  Set<Uri> get cachedUris => _importCache.keys.toSet();
+
   /// Creates an import cache that resolves imports using [importers].
   ///
   /// Imports are resolved by trying, in order:


nex3:
We already track similar information in `_EvaluateVisitor._includedFiles`. Although this is specifically made to match Node Sass's API and so only handles filesystem paths, it could be made to handle URLs instead and then converted to filesystem paths in `node.dart`.

The real question is how we expose this from the public Dart API. The public `compile*` methods all return strings, which are difficult to add extra information to. Ideally we'd return a `CompileResult` like we do for the internal `compile*` methods, but that would be a breaking change at this point. It's probably easiest to just do the same thing we do for source maps, and have users pass in a callback that gets passed the URLs.

nex3:
Thanks for the contribution! I've ended up going a different direction in #1398, but I really appreciate you pitching in!

PR #1364, opened 2021-06-17
Ensure `sassIndexToStringIndex()` accepts only unitless numbers.

Awjin:


nex3:
Since this is technically a breaking change, and it's not immediately blocking anything, I think it should probably emit a deprecation warning for now rather than throwing an error.

Awjin:
Sounds good. Will probably hold off for now to avoid deprecation fatigue

lib/src/value/string.dart
```diff
@@ -89,7 +89,9 @@ class SassString extends Value implements ext.SassString {
           text, sassIndexToRuneIndex(sassIndex, name));
 
   int sassIndexToRuneIndex(ext.Value sassIndex, [String? name]) {
-    var index = sassIndex.assertNumber(name).assertInt(name);
+    var indexNumber = sassIndex.assertNumber(name);
+    indexNumber.assertNoUnits();
```
stof, line 93: this should pass `name`

PR #1375, opened 2021-06-23
Add links to CLI docs in README

Awjin:
Closes https://github.com/sass/sass-site/issues/560


nex3 (approved):
I don't think you need to link the docs from each separate way of installing the CLI—just the first link should do.

OlivierJM:
Thanks @Awjin

Issue #1343, opened 2021-06-03
Loading a module with the configuration does not work for multiple @forward blocks


lesha38:
Hi! I have a problem with [configuring modules](https://sass-lang.com/documentation/at-rules/forward#configuring-modules).

If the "opinionated" file contains only one `@forward` block then everything works, for example:
```
// _opinionated.scss
@forward 'library' with (
  $black: #222 !default,
  $border-radius: 0.1rem !default
);
```
```
// style.scss
@use 'opinionated' with ($black: #333);
```
```
// log
...compiled successfully
```

---------------------------------------------------------------------------------------------------------------

But if the "opinionated" file contains more than one `@forward` block we have an error, for example:

```
// _opinionated.scss
@forward 'library' with (
  $black: #222 !default,
  $border-radius: 0.1rem !default
);
@forward 'another-library' with (
  $yellow: #FBFF00 !default,
  $font-size: 14px !default
);
```
```
// style.scss
@use 'opinionated' with ($black: #333, $yellow: #333);
```

```
// log
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
SassError: This variable was not declared with !default in the @used module.
   $yellow: #444,
   ^^^^^^^^^^^^^^^^^^
   src\styles\index.scss 19:3  @use
```
---------------------------------------------------------------------------------------------------------------

software | version
-- | --
sass | 1.34.1

jathak:
Can you provide a full reproduction repo? It's hard to tell what's going on here without the full context.

lesha38:
> Can you provide a full reproduction repo? It's hard to tell what's going on here without the full context.

Full reproduction repo: https://github.com/lesha38/sass-problem  

1. Clone
2. Run `npm i`
3. Run `npm run build`

PR #1382, opened 2021-06-26
Allow configuring a module that contains multiple configured `@forward`s

Awjin:
Defers the check for valid `@use with` configs until all configured `@forward` rules have been evaluated.

Fixes https://github.com/sass/dart-sass/issues/1343

See https://github.com/sass/sass-spec/pull/1675

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -1344,7 +1344,9 @@ class _EvaluateVisitor
                     if (!variable.isGuarded) variable.name
                 });
 
-      _assertConfigurationIsEmpty(newConfiguration);
+      if (node == _stylesheet.forwards.last) {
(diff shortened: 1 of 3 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -1344,7 +1344,9 @@ class _EvaluateVisitor
                     if (!variable.isGuarded) variable.name
                 });
 
-      _assertConfigurationIsEmpty(newConfiguration);
+      if (node == _stylesheet.forwards.last) {
+        _assertConfigurationIsEmpty(newConfiguration);
```
nex3, line 1348: I'm not sure this is right either... since `newConfiguration` contains all configured variables that were passed into the current module, it might still have an unsatisfied configuration for the *current* module. Really we just want to verify that there all the variables configured by `node.configuration` were removed.

PR #1389, opened 2021-07-07
Fix bug in RecursiveAstVisitor

jathak:
It looks like the body of `visitParenthesizedExpression` was accidentally removed when the type parameter was removed from this class.

This will fix sass/migrator#202 once it's merged and we cut a new migrator release.


nex3 (approved):

jathak:
Should I wait for a Dart Sass release or should I work around this in the migrator to get out a release now?

nex3:
Either way is fine—you can also feel free to cut a Dart Sass release just for this if you want.

PR #1411, opened 2021-07-29
Abort sass if stdin is closed when watching

mcrumm:
Most command line tools (cat, fsevents, etc) and JS CLI watchers (postcss,
webpack, etc) listen to stdin and abort when the stdin is closed. This is
important because there is no guarantee the process who invokes sass will
terminate it, but it is guaranteed the stdin will be closed once the parent
dies.

This pull request listens to stdin when watching and exits accordingly,
in order to provide a user convenience and avoid zombie processes.

Note I chose to only listen when stdin is _not_ a TTY: this is to maintain
the ability to background the watch process without unexpectedly closing it.
This means for instance you cannot use Ctrl+D to abort the watch, however,
when invoked by another process sass --watch will exit as expected.

For context, there is a some history wrt listening to stdin. Most recently
this feature was added to esbuild, wherein the PR comments @evanw provided a
great reference summary which I will repeat here for context (+ add a few more
examples).

PS– Thanks for dart-sass!

  * Esbuild
    * evanw/esbuild#1449
  * Webpack
    * webpack/webpack#1308
    * webpack/webpack#1311
  * Parcel (they decided to not do it since there's a workaround)
    * parcel-bundler/parcel#1004
  * Brunch (they added behind a feature flag to fix background jobs)
    * brunch/brunch#920
    * brunch/brunch#922
    * brunch/brunch#998
    * brunch/brunch#1003
  * PostCSS
    * postcss/postcss-cli#358
  * Vue
    * vuejs/vue-cli#1597
    * vuejs/vue-cli#5376
  * Vite (they decided to require an env var)
    * vitejs/vite#1857
    * vitejs/vite#3659
    * vitejs/vite#4079
    * vitejs/vite#4082
    * vitejs/vite#4196
    * vitejs/vite#4232


jathak:
Thanks for this! We're definitely willing to merge this once the tests are passing.

mcrumm:
@jathak Awesome! I was able to reproduce the failing tests on ubuntu. I rebased to get the latest changes and now everything is green locally :)

nex3:
Looks like the tests are still timing out.

mcrumm:
Hi @nex3 – indeed, I spoke too soon :) I am happy to dive in deeper. I am new to Dart, but my naive impression is that maybe the exit in the stdin listener is racing another exit? Do you have any pointers on what to look for / changes to make?

nex3:
The proximate issue is that the tests are timing out while waiting for the process to exit. This is probably because the subscription to stdin is keeping the process alive when it would otherwise fall through the end of `main()`.

Ideally, we'd avoid using `exit()` entirely and either explicitly close the watcher when stdin closes, or explicitly close the stdin subscription when the watcher closes. I'll see if I can come up with a good way to model that.

nex3:
https://github.com/dart-lang/async/pull/194 adds some utilities to the `async` package that will make it much easier to model the concept of "cancel one operation when the other finishes and vice versa".

PR #1413, opened 2021-07-30
Add a sass_api package that exposes more API details

### LINKED ISSUE #236: Dart API: Provide access to the Sass AST
This will become necessary for porting tooling that's currently relying on Ruby Sass, such as [SCSS-Lint](https://github.com/brigade/scss-lint).

This is blocked on https://github.com/dart-lang/sdk/issues/28066, which is necessary to avoid exposing implementation details of the AST that should only be used by the `sass` package itself.

nex3:
Closes #236


pkg/sass_api/pubspec.yaml
```diff
@@ -0,0 +1,17 @@
...
+version: 1.0.0-beta.1
+description: Additional APIs for Dart Sass.
+author: Sass Team
+homepage: https://github.com/sass/dart-sass
+
+environment:
+  sdk: '>=2.12.0 <3.0.0'
+
+dependencies:
+  sass: 1.36.0
```
jathak, line 14: One of the tests is expecting this to be 1.37.0

lib/src/import_cache.dart
```diff
@@ -5,11 +5,12 @@
 // DO NOT EDIT. This file was generated from async_import_cache.dart.
 // See tool/grind/synchronize.dart for details.
 //
-// Checksum: 27d8582c2ab318a52d433390ec256497b6af5dec
+// Checksum: d9b0c8f576966b1b4fd2e7255342a0a76369689d
```
jathak, line 8: The tests are saying this is out of date.

jathak (commented):
The specs seem to be failing due to an issue with content blocks

lib/src/ast/sass/import/dynamic.dart
```diff
@@ -2,23 +2,36 @@
...
 
 /// An import that will load a Sass file at runtime.
+///
+/// {@category AST}
+@sealed
 class DynamicImport implements Import {
+  /// The URL of the file to import.
+  ///
+  /// If this is relative, it's relative to the containing file.
+  Uri get url => Uri.parse(urlString);
```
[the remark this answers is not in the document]
nex3: Good idea. Can you file a feature request?

nex3 (commented):
I had a logic error with the way I had refactored content block detection. I fixed that refactor and split it into its own commit; PTAL.

jathak (approved):

PR #1419, opened 2021-08-03
Manually resolve symlinks before deploying sub-packages

nex3:


tool/grind/subpackages.dart
```diff
@@ -61,6 +61,17 @@ Future<void> deploySubPackages() async {
...
     File(pubspecPath).writeAsStringSync(json.encode(pubspecYaml));
 
+    // We use symlinks to avoid duplicating files between the main repo and
+    // child repos, but `pub lish` doesn't resolve these before publishing so we
+    // have to do so manually.
+    for (var entry
+        in Directory(package).listSync(recursive: true, followLinks: false)) {
+      if (entry is! Link) continue;
+      var target = p.join(p.dirname(entry.path), entry.targetSync());
+      entry.deleteSync();
```
jathak, line 71: Is there a check somewhere that ensures this is only run on GitHub Actions, to make sure the copied files aren't checked in?

jathak (approved):

tool/grind/subpackages.dart
nex3: Not explicitly, but it'll fail if `PUB_CREDENTIALS` isn't set so I think that's probably enough protection from accidentally shooting yourself in the foot.

PR #1426, opened 2021-08-09
Fix a crash in the Node importer

### LINKED ISSUE #1410: Unable to view the compilation errors when using importer
Hello!,

I'm trying to migrate from node-sass to dart-sass and I'm having some issues when using a custom importer:

- First of all, I have seen that the importer is not called if the file exists... well, I have found a workaround to deal with by using some prefix in the url to import and I can live with it for now

- Then I have found that the _prev_ parameters that is sent to the importer function is url encoded, I'm not sure if it is an intended change or not, but it works different from node-sass. Anyway it's easy to solve wih a decodeURI, so it is not really a problem too

Solving this 2 small issues I have managed to make it work, and now I'm able to compile some sass files... but the problem is when there is some compilation problem in one of the imported files (for example, a variable that does not exists, a syntax error, etc.) it should show the compilation error but instead it throws a weird exception like this:

```
Invalid argument(s): Uri c:%5Cgitrepos%5CFRONT_Tools%5Cdart-sass-sample%5Ctests%5Csass%5Cimported.scss must have scheme 'file:'.
    at Object.wrapException (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:1240:17)
    at WindowsStyle.pathFromUri$1 (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:33219:17)
    at StaticClosure.fromUri (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:16169:37)
    at Object.NullableExtension_andThen0 (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:18495:40)
    at Object._wrapException (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:13048:14)
    at StaticClosure._renderSync (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:13029:18)
    at Object.Primitives_applyFunction (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:1136:30)
    at Object.Function_apply (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:5903:16)
    at _callDartFunctionFast (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:7605:16)
    at Object.renderSync (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\sass\sass.dart.js:7583:18) {
  dartException: ArgumentError {
    _hasValue: false,
    invalidValue: null,
    name: null,
    message: "Uri c:%5Cgitrepos%5CFRONT_Tools%5Cdart-sass-sample%5Ctests%5Csass%5Cimported.scss must have scheme 'file:'."
  },
  name: ''
}
```

So in the case that some user makes a mistake in a sass file, it's very difficult to find the real problem...

The same code in node-sass shows the correct error and makes it easy to solve it, for example an undefined variable:

```
Error: Undefined variable: "$some-var".
    at Object.module.exports.renderSync (C:\gitrepos\FRONT_Tools\dart-sass-sample\node_modules\node-sass\lib\index.js:440:23)
    at testSass (C:\gitrepos\FRONT_Tools\dart-sass-sample\tests\sass-import.js:26:29)
    at testNodeSass (C:\gitrepos\FRONT_Tools\dart-sass-sample\tests\sass-import.js:50:5)
    at Object.<anonymous> (C:\gitrepos\FRONT_Tools\dart-sass-sample\tests\sass-import.js:53:1)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10)
    at internal/main/run_main_module.js:17:11 {
  status: 1,
  file: 'C:\\gitrepos\\FRONT_Tools\\dart-sass-sample\\tests\\sass\\imported.scss',
  line: 3,
  column: 13,
  message: 'Undefined variable: "$some-var".',
  formatted: 'Error: Undefined variable: "$some-var".\n' +
    '        on line 3 of tests/sass/imported.scss\n' +
    '        from line 1 of tests/sass/main.scss\n' +
    '>>     height: $some_var;\r\n' +
    '   ------------^\n'
}
```

The reason I need to use the importer this way is to be able to read and compile non UTF-8 files as I commented long ago in this node-sass issue https://github.com/sass/node-sass/issues/2281... My workaround for this problem was to read the contents of all the files by a custom function so I can deal with the encoding before passing it to sass...

Anyway, I provide a simple example for the problem with the importer that compares the error in node-sass with the one using dart-sass, you can see the sample at:

https://gist.github.com/xelloss1012/c003a7fb0bdfcb12849be3c45197b205

Thanks!

nex3:
When a path was returned at the same time as a file's contents, it was
interpreted as a URL without first being translated to one. This
crashed for absolute Windows paths.

Closes #1410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.37.6
+
+* Don't crash when a Windows path is returned by a custom Node importer at the
+  same time as file contents.
+
 ## 1.37.5
 
 * No user-visible changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 1.37.6
 
+### JS API
+
 * Don't crash when a Windows path is returned by a custom Node importer at the
   same time as file contents.
 
--- a/lib/src/importer/node/implementation.dart
+++ b/lib/src/importer/node/implementation.dart
@@ -186,7 +186,7 @@ class NodeImporter {
     if (file == null) {
       return Tuple2(contents ?? '', url);
     } else if (contents != null) {
-      return Tuple2(contents, file);
+      return Tuple2(contents, p.toUri(file).toString());
     } else {
       var resolved =
           loadRelative(p.toUri(file).toString(), previous, forImport) ??
--- a/pkg/sass_api/CHANGELOG.md
+++ b/pkg/sass_api/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.0-beta.3
+
+* No user-visible changes.
+
 ## 1.0.0-beta.2
 
 * No user-visible changes.
--- a/pkg/sass_api/pubspec.yaml
+++ b/pkg/sass_api/pubspec.yaml
@@ -2,7 +2,7 @@ name: sass_api
 # Note: Every time we add a new Sass AST node, we need to bump the *major*
 # version because it's a breaking change for anyone who's implementing the
 # visitor interface(s).
-version: 1.0.0-beta.2
+version: 1.0.0-dev
 description: Additional APIs for Dart Sass.
 homepage: https://github.com/sass/dart-sass
 
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.12.0 <3.0.0'
 
 dependencies:
-  sass: 1.37.5
+  sass: 1.37.6
 
 dependency_overrides:
   sass: {path: ../..}
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.37.5
+version: 1.37.6-dev
 description: A Sass implementation in Dart.
 homepage: https://github.com/sass/dart-sass
 
--- a/test/double_check_test.dart
+++ b/test/double_check_test.dart
@@ -50,13 +50,18 @@ void main() {
             .convert(File("$package/CHANGELOG.md").readAsStringSync())
             .first;
         expect(firstLine, startsWith("## "));
-        var changelogVersion = firstLine.substring(3);
+        var changelogVersion = Version.parse(firstLine.substring(3));
 
         var pubspec = Pubspec.parse(
             File("$package/pubspec.yaml").readAsStringSync(),
             sourceUrl: p.toUri("$package/pubspec.yaml"));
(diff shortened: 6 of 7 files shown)


Awjin (approved):

devongovett:
This appears to have introduced a breaking change. Previously, the result of an importer was treated as a URL but now it's being treated as a path. In Parcel, we return a `file://` URL as was previously expected, but now on the latest version of Sass this is broken and the `file://` prefix gets escaped after being returned to Sass. We could change to return a path instead but then we would be broken with older versions of Sass people might have in their projects.

Should this change be reverted? Looking at the original issue, I think it was user error of returning a path instead of a URL.

nex3:
Treating the return value as a path is definitely correct here; that's what the importer [was already doing](https://github.com/sass/dart-sass/blob/fd7eec9eacb3ac1b8c46388a66220c8f7a77ae2b/lib/src/importer/node/implementation.dart#L192) if `file` was passed *without* `content`, and that's what the Node Sass API we're trying to match expects. It's also what [the documentation says](https://github.com/sass/dart-sass/blob/fd7eec9eacb3ac1b8c46388a66220c8f7a77ae2b/lib/src/importer/node/implementation.dart#L192):

> An object with the key `file` whose value is a path on disk.

I'm very sorry that this broke you, though, and more broadly that this API is so inconsistent about the distinction between paths and URLs. The [new importer API](https://sass-lang.com/blog/request-for-comments-new-js-api#importers) will be much more explicit and straightforward.

For now, I recommend returning a relative path with forward slashes from your importer. This will *usually* be interpreted correctly when parsed as a URL, and will always be interpreted correctly when parsed as a path.

Issue #1414, opened 2021-07-31
Add a common interface for UseRule, ForwardRule, and ImportRule


jathak:

stof:
@jathak should be closed as https://github.com/sass/dart-sass/pull/1445 did it (the automatic `Fixes ...` in PR description does not support `and` to list multiple issues, it must repeat the `Fixes` instead, which is why it was not closed)

PR #1445, opened 2021-08-20
Add shared interfaces for various AST nodes

### LINKED ISSUE #1401: Add more fine-grained spans to some AST nodes
It would be useful to have more fine-grained spans for certain AST nodes. Specifically:

- A span for the URL portion of a use rule.
- A span for the namespace of a use rule that has an `as foobar` portion.
- A span for the namespace of a namespaced variable.
- A span for just the variable, without the namespace, for a namespaced variable.

I would be happy to be the assignee for this issue.

jathak:
Fixes #1401 and #1414.

Adds `Dependency`, `SassDeclaration`, and `SassReference` interfaces,
which expose some getters that multiple AST nodes have in common with a
single type.

These also add getters for common subspans (URL, name, and namespace) to
the interfaces.

To allow for `FunctionExpression` to have a common interface with the
other two reference nodes, its old `name` getter has been renamed to
`interpolatedName`, and the new `name` getter now is either just a
string, or null if the name is interpolated.


lib/src/ast/sass/interface/reference.dart
```diff
@@ -0,0 +1,38 @@
...
+  /// without a namespace.
+  String? get namespace;
+
+  /// The name of the member being referenced, with underscores converted to
+  /// hyphens.
+  ///
+  /// For [VariableExpression]s and [IncludeRule]s, this will never be null.
+  /// For [FunctionExpression]s, this will be null if the actual name is an
+  /// interpolation (in which case this is a plain CSS function, not a reference
+  /// to a Sass function).
```
nex3, line 25: I think this indicates that we should split out a `CssFunctionExpression` from `FunctionExpression`. Not only will it make this easier for users of the AST to handle consistently, it'll make it clear when a function is or is not `-`-normalized.

lib/src/ast/sass/argument.dart
```diff
@@ -30,6 +31,9 @@ class Argument implements SassNode {
   String get originalName =>
       defaultValue == null ? span.text : declarationName(span);
 
+  FileSpan get nameSpan =>
+      defaultValue == null ? span : span.subspan(0, name.length + 1);
```
nex3, line 35: It's possible (albeit not very likely) that `name` has escape sequences, which would make this computation invalid. I'd probably write an `initialIdentifier()` helper in `SpanExtensions` that just walks forward through identifier characters or backslash-escapes.

lib/src/ast/sass/expression/function.dart
```diff
@@ -29,22 +31,32 @@ class FunctionExpression implements Expression, CallableInvocation {
   ///
   /// If this is interpolated, the function will be interpreted as plain CSS,
   /// even if it has the same name as a Sass function.
-  final Interpolation name;
+  final Interpolation interpolatedName;
+  String? get name => namespace == null
```
nex3, line 35: Nit: newline above this

lib/src/ast/sass/expression/variable.dart
```diff
@@ -22,6 +23,13 @@ class VariableExpression implements Expression {
 
   final FileSpan span;
 
+  FileSpan get nameSpan =>
+      namespace == null ? span : span.subspan(namespace!.length + 1);
```
nex3, line 27: Rather than using `!`, assign `namespace` to a local variable so its non-nullability is inferred.

lib/src/ast/sass/interface/reference.dart
```diff
@@ -0,0 +1,38 @@
...
+  String? get name;
+
+  /// The span containing this reference's name.
+  ///
+  /// For variables, this should include the `$`.
+  FileSpan get nameSpan;
+
+  /// The span containing this reference's namespace, or an empty span
+  /// immediately before the name if the namespace is null.
+  FileSpan get namespaceSpan;
```
nex3, line 37: I think it makes more sense to just make this `null` if `namespace` is `null`. I'm not generally a fan of using sentinel values when `null` will do.

lib/src/ast/sass/statement/forward_rule.dart
```diff
@@ -74,6 +75,9 @@ class ForwardRule implements Statement {
 
   final FileSpan span;
 
+  FileSpan get urlSpan =>
+      span.subspan(RegExp(r'@forward\s+').matchAsPrefix(span.text)!.end);
```
nex3, line 79: Similarly, I'd write a `withoutInitialAtRule()` span helper that you can use here and below.

Also, don't forget that a `@forward` can have `show`/`hide`/`with` clauses that will need to be excluded here.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.38.0
+version: 1.38.1
```
nex3, line 2: Always add a CHANGELOG entry, even if the changes aren't visible here. Otherwise people are liable to get confused, and I think it'll break the release process too.

nex3 (changes requested):

lib/src/ast/sass/expression/function.dart
```diff
@@ -24,39 +25,39 @@ class FunctionExpression
...
+  /// hyphens.
   ///
-  /// If this is interpolated, the function will be interpreted as plain CSS,
-  /// even if it has the same name as a Sass function.
-  final Interpolation interpolatedName;
-  String? get name => namespace == null
-      ? interpolatedName.asPlain?.replaceAll('_', '-')
-      : interpolatedName.asPlain;
+  /// If this function is a plain CSS function, use [originalName] instead.
+  final String name;
```
nex3, line 35: I'd make this a getter, since a single `String.replaceAll()` isn't very expensive.

lib/src/ast/sass/statement/forward_rule.dart
```diff
@@ -75,8 +76,12 @@ class ForwardRule implements Statement, Dependency {
 
   final FileSpan span;
 
-  FileSpan get urlSpan =>
-      span.subspan(RegExp(r'@forward\s+').matchAsPrefix(span.text)!.end);
+  FileSpan get urlSpan {
+    var skipRule = span.withoutInitialAtRule();
+    var quote = skipRule.text[0];
+    var end = skipRule.text.indexOf(quote, 1);
+    return skipRule.subspan(0, end + 1);
```
nex3, line 83: This might be worth making a Span utility as well.

lib/src/ast/sass/statement/include_rule.dart
```diff
@@ -41,16 +41,21 @@ class IncludeRule implements Statement, CallableInvocation, SassReference {
...
-    var match = RegExp(r'(\+|@include)\s*').matchAsPrefix(span.text);
-    var start = match!.end;
-    if (namespace != null) start += namespace!.length + 1;
-    return span.subspan(start, start + name.length);
+    var startSpan = span.text.startsWith('+')
+        ? span.subspan(1).trimLeft()
+        : span.withoutInitialAtRule();
+    if (namespace != null) {
+      startSpan = startSpan.withoutInitialIdentifier().subspan(1);
+    }
```
nex3, line 49: Consider adding `SpanExtensions.withoutNamespace()`.

lib/src/parse/parser.dart
```diff
@@ -479,35 +479,8 @@ class Parser {
...
-        return 0xFFFD;
-      } else {
-        return value;
-      }
-    } else {
-      return scanner.readChar();
-    }
-  }
+  int escapeCharacter() =>
+    consumeEscapedCharacter(scanner);
```
nex3, line 483: Nit: reformat

lib/src/utils.dart
```diff
@@ -383,24 +384,102 @@ Map<K1, Map<K2, V>> copyMapOfMap<K1, K2, V>(Map<K1, Map<K2, V>> map) =>
...
+      return 0xFFFD;
+    } else {
+      return value;
+    }
+  } else {
+    return scanner.readChar();
+  }
+}
+
 extension SpanExtensions on FileSpan {
```
nex3, line 419: This is big enough now that it should maybe go into its own file.

nex3 (approved):

PR #1446, opened 2021-08-20
Expose `charset` option in Node API [closes #567]

### LINKED ISSUE #567: Allow removing @charset atrule with expanded OutputStyle
It looks like Dart Sass always prefixes stylesheets with a `@charset` atrule [if the output style isn't `compressed`](https://github.com/sass/dart-sass/blob/759b5873641540aa73f450379fb26fe43143d284/lib/src/visitor/serialize.dart#L62).

This is unnecessary for stylesheets [embedded in <style> elements](https://www.w3.org/TR/CSS2/syndata.html#x57):

> When a style sheet is embedded in another document, such as in the STYLE element or "style" attribute of HTML, the style sheet shares the character encoding of the whole document.

The MDN has a [stronger interpretation of this](https://developer.mozilla.org/en-US/docs/Web/CSS/@charset):

> If several @charset at-rules are defined, only the first one is used, and it cannot be used inside a style attribute on an HTML element or inside the <style> element where the character set of the HTML page is relevant.

Outputting the `@charset` atrule should be optional, because removing this atrule isn't trivial when maintaining a sourcemap.

adamhooper:
My first commit -- please be gentle (and see https://github.com/sass/dart-sass/issues/567#issuecomment-902937921 for rationale)


nex3:
Unfortunately, adding a new feature to the JS API isn't so simple. Because the JS API is shared between multiple implementations, you'll first need to introduce a proposal for the new feature in https://github.com/sass/sass, following [this process](https://github.com/sass/sass/blob/main/CONTRIBUTING.md). Only once that proposal is approved can we move on to the implementation phase.

nex3 (approved):

adamhooper:
@nex3 Thank you for all your support and work on Sass! It feels awesome to be a part of it, however small :).

PR #1452, opened 2021-08-26
Add support for first-class calc()

nex3:
See sass/sass#818
See sass/sass-spec#1688

### DIFF
--- a/lib/src/functions/meta.dart
+++ b/lib/src/functions/meta.dart
@@ -43,6 +43,9 @@ final global = UnmodifiableListView([
     if (value == sassNull) return SassString("null", quotes: false);
     if (value is SassNumber) return SassString("number", quotes: false);
     if (value is SassFunction) return SassString("function", quotes: false);
+    if (value is SassCalculation) {
+      return SassString("calculation", quotes: false);
+    }
     assert(value is SassString);
     return SassString("string", quotes: false);
   }),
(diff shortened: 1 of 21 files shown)


lib/src/functions/meta.dart
```diff
@@ -60,6 +63,22 @@ final global = UnmodifiableListView([
   })
 ]);
 
+/// The definitions of Sass introspection functions that are only available from
+/// the `sass:meta` module, not as global functions.
+final local = UnmodifiableListView([
+  _function("calc-name", r"$calc", (arguments) {
```
stof, line 69: should it `calc-name` or `calculation-name` as the value type has been changed to `calculation` to avoid confusion with the fact that not all calculations are for `calc()` ?

nex3: That's something we probably should have talked through in the design process :sweat_smile:. I think it's fine though to keep this name short, since "calc" can also be read as short for "calculation".

stof: > That's something we probably should have talked through in the design process

I agree. But that's something I missed when re-reading the proposal after the renaming of the type from `calc` to `calculation`. It is still before to discuss that now than after landing the feature in a release.

lib/src/value/calculation.dart
```diff
@@ -0,0 +1,387 @@
...
+  bool get isSpecialNumber => true;
+
+  /// Creates a `calc()` calculation with the given [argument].
+  ///
+  /// The [argument] must be either a [SassNumber], a [SassCalculation], an
+  /// unquoted [SassString], a [CalculationOperation], or a
+  /// [CalculationInterpolation].
+  ///
+  /// This automatically simplifies the calculation, so it may return a
+  /// [SassNumber] rather than a [SassCalculation].
```
Awjin, line 45: Might be worth documenting that this throws if unsimplified args are not compatible (ditto for `min`, `max` and `clamp`)

Awjin (approved):

PR #1457, opened 2021-09-01
Fix pre-existing HintCode.UNNECESSARY_TYPE_CHECK_TRUE

scheglov:
This prepares for landing https://dart-review.googlesource.com/c/sdk/+/190360 into the analyzer.

### DIFF
--- a/lib/src/extend/functions.dart
+++ b/lib/src/extend/functions.dart
@@ -404,14 +404,18 @@ List<List<List<ComplexSelectorComponent>>>? _mergeFinalCombinators(
       result.addFirst([
         [compound2, combinator2]
       ]);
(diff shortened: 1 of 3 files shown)


nex3 (changes requested):
This code is actually more complicated than it looks. Although the type check in question isn't necessary in `async_evaluate.dart`, that code is compiled to the synchronous `evaluate.dart` and transformed into a `callable is Callable` check which *is* relevant. Can you just add an ignore for the hint instead?

scheglov:
Done.

nex3 (approved):

scheglov:
Could you merge these changes? It seems that I don't have corresponding permissions.

PR #1465, opened 2021-09-05
Upgrade to analyzer 2

stof:
This package is reported by `pub outdated`.

However, I'm not sure whether this package is actually useful. AFAICT, it is not used anywhere as the CI is using the `dart analyze` CLI instead of running the analyzer manually.


jathak (approved):
It's used by `tool/grind/synchronize.dart`.

PR #1469, opened 2021-09-08
Fix `@at-root` bug for common case of built-in use

jathak:
`@at-root` should now work properly in a nested import as long as the
only `@use` or `@forward` rules present are for built-in modules.

This is a partial but not complete fix for #1347. It's a bit hacky to
special case built-in modules here, but since they represent the only
real legitimate case for using this combination of features, it lets us
fix the bug for this case and avoid the bordering-on-infeasible work of
fixing this for user-defined modules.

See sass/sass-spec#1699.

### DIFF
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -1481,6 +1481,15 @@ class _EvaluateVisitor
         return;
       }
 
+      // If only built-in modules are loaded, we still need a separate
+      // environment to ensure their namespaces aren't exposed in the outer
+      // environment, but we don't need to worry about `@extend`s, so we can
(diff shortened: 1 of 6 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -1481,6 +1481,13 @@ class _EvaluateVisitor
         return;
       }
 
+      // If only built-in modules are loaded, we still need a separate
+      // environment to ensure their namespaces aren't exposed in the outer
+      // environment, but we don't need to worry about `@extend`s.
```
nex3, line 1486: Explain what not having to worry about extends means concretely. Maybe something like "...so we can add styles directly to the existing stylesheet rather than creating a new one"?

nex3 (approved):

PR #1472, opened 2021-09-09
Fix the leftover-configuration check for `@forward ... with`

nex3:
This check was previously checking whether *any* variables were left
in this configuration, which could include variables that were adopted
from outer configurations. This threw invalid errors when that outer
configuration would have been satisfied by another variable (or
forward) later in the file.

Closes sass/sass#1460
See sass/sass-spec#1701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.39.2
+
+* Fix a bug where configuring with `@use ... with` would throw an error when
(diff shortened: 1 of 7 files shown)


jathak (approved):
I believe you also need to increment the `sass_api` version

nex3:
> I believe you also need to increment the `sass_api` version

You'd think since I added the checks for that I'd remember to satisfy them...

PR #1481, opened 2021-09-14
Allow the + operator for calculations and strings

nex3:
See sass/sass#3142
See sass/sass#3146
See sass/sass-spec#1705


jathak (commented):
It looks like the specs are failing because this is outputting quoted strings, while the specs are expecting unquoted. Which is right?

nex3:
The quoted string behavior is right. I'll update the specs.

jathak (approved):

PR #1485, opened 2021-09-15
Add back support for min/max calculations

nex3:
See sass/sass#3142
See sass/sass#3150
See sass/sass-spec#1710


lib/src/value/calculation.dart
```diff
@@ -161,15 +161,28 @@ class SassCalculation extends Value {
   /// [SassCalculation], an unquoted [SassString], a [CalculationOperation], or
   /// a [CalculationInterpolation].
   static Object operate(
-      CalculationOperator operator, Object left, Object right) {
+          CalculationOperator operator, Object left, Object right) =>
+      operateInternal(operator, left, right, inMinMax: false);
+
+  /// Like operator, but with the internal-only [inMinMax] parameter.
```
stof, line 167: should be `Like operate` (and maybe even `Like [operate]`

Awjin (approved):

stof (approved):

PR #1500, opened 2021-09-21
Fix interpolation lookahead for calculations

### LINKED ISSUE #1499: Calculations are parsed incorrectly if followed by parenthesized interpolation
Dart Sass's lookahead logic for determing if a calculation contains an interpolation appears to be subtly wrong, so if the stylesheet after a calculation contains an interpolation within parentheses, the entire calculation is parsed as though it were an interpolation.

For example:

```scss
$variable: 1;
@debug calc($variable);
.foo { :not(#{&}) {} }
```

prints `calc($variable)` rather than `1`.

Originally reported in https://github.com/sass/sass/issues/818#issuecomment-923869341

nex3:
Closes #1499
See sass/sass-spec#1715

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.42.1
+
+* Fix a bug where Sass variables and function calls in calculations weren't
+  being resolved correctly if there was a parenthesized interpolation elsewhere
+  in the file.
+
 ## 1.42.0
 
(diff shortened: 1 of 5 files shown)


jathak (approved):
LGTM, but I think the space in the sass-spec reference means this was run against main, rather than the PR branch.

nex3:
Good catch, I'll re-run the tests

PR #1501, opened 2021-09-22
Cordon the Node Sass API off into a legacy directory

nex3:


lib/src/importer/node/implementation.dart
```diff
@@ -10,10 +10,10 @@ import 'package:tuple/tuple.dart';
 
 import '../../io.dart';
 import '../../node/function.dart';
-import '../../node/importer_result.dart';
+import '../../node/legacy/importer_result.dart';
+import '../../node/legacy/render_context.dart';
 import '../../node/utils.dart';
 import '../../util/nullable.dart';
-import '../../node/render_context.dart';
```
Awjin, line 16: (wish Github would let me leave file-level comments)

Should this implementation be marked legacy?

Awjin (approved):

PR #1503, opened 2021-09-22
Add basic compile functions for the new JS API

nex3:
See sass/sass-spec#1717


lib/src/node/exception.dart
```diff
@@ -0,0 +1,62 @@
...
+          value: dartException,
+          enumerable: false
+        });
+      }
+
+      toString() {
+        return this.message;
+      }
+    }
+  ''') as Function;
```
Awjin, line 40: 🙈

Awjin (approved):

Issue #1436, opened 2021-08-11
Error if filename starts with number


claudia-romano:
Hi there, is it possible to use a number as the first character of a file name? I get the 'Invalid Sass identifier' error if the filename starts with a number:
```
Error: Invalid Sass identifier "3d-cards"
   ╷
14 │ @use "components/_3d-cards.scss";
```

Thanks!

stof:
you need to alias it as to a valid identifier instead of relying on the auto-generated one that takes the file name (as the file name does not produce a valid identifier):

```scss
@use "components/_3d-cards.scss" as cards;
```

claudia-romano:
Thank you!

PR #1513, opened 2021-09-28
Improve the error message for invalid default namespaces

stof:
Closes https://github.com/sass/sass/issues/2906 (and also https://github.com/sass/dart-sass/issues/1436)

https://github.com/sass/sass-spec/pull/1725


lib/src/importer/legacy_node.dart
```diff
@@ -2,4 +2,5 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
-export 'legacy_node/interface.dart' if (dart.library.js) 'legacy_node/implementation.dart';
+export 'legacy_node/interface.dart'
```
stof, line 5: this change is the result of running `dart run grinder`.

PR #1521, opened 2021-10-05
Update to the latest linter package and fix lints

nex3:


lib/src/visitor/recursive_ast.dart
```diff
@@ -24,10 +24,6 @@ abstract class RecursiveAstVisitor extends RecursiveStatementVisitor
     implements ExpressionVisitor<void> {
   const RecursiveAstVisitor();
 
-  void visitAtRootRule(AtRootRule node) {
-    super.visitAtRootRule(node);
```
jathak, line 28: Instead of deleting this, shouldn't this also visit the query if it's non-null?

jathak (approved):

lib/src/visitor/recursive_ast.dart
nex3: Good catch!

PR #1547, opened 2021-11-04
Let inspect() print nested, empty, bracketed lists.

Awjin:
https://github.com/sass/sass-spec/pull/1740

Fixes https://github.com/sass/sass/issues/3160

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.45.1-dev
+version: 1.45.1
 description: A Sass implementation in Dart.
 homepage: https://github.com/sass/dart-sass
 
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.45.0-rc.2
+version: 1.45.1-dev
 description: A Sass implementation in Dart.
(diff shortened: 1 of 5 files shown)


lib/src/visitor/serialize.dart
```diff
@@ -611,7 +611,8 @@ class _SerializeVisitor
     _writeBetween<Value>(
         _inspect
             ? value.asList
-            : value.asList.where((element) => !element.isBlank),
+            : value.asList
+                .where((element) => element.hasBrackets || !element.isBlank),
```
stof, line 615: is this the right fix ? Or should bracketed lists not be considered blank ?

Side note: the commit message says this is for `inspect`, but the fix is for the CSS serialization, not the inspect one.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.45.0-rc.2
+version: 1.45.1-dev
```
nex3, line 2: I think we can mark this as a release version.

stof (approved):

nex3 (approved):
LGTM, although as discussed offline let's wait to merge this until we've officially released 1.45.0.

PR #1554, opened 2021-11-16
Add support for css empty namespace attribute selector

pjeweb:
sass/sass-spec#1742

Fixes: sass/sass#3199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 * Fix a bug where calculations with different operators were incorrectly
   considered equal.
 
+* Properly parse attribute selectors with empty namespaces.
+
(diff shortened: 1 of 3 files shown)


stof:
Please add a reference to the sass-spec PR in your PR description, so that the CI runs the patched sass-spec (once the workflow run is approved by a maintainer)

stof:
the change looks good to me

pjeweb:
I think the CI did not actually test with the new spec. I think it may be either because of a newline before the link or just because i changed the pr message afterwards. Maybe someone can trigger a new run? Thanks!

nex3 (approved):
Thanks for the contribution!

PR #1556, opened 2021-11-16
Fix equality checks for CalculationOperation

### LINKED ISSUE #1548: `calc()` with different operators wrongly considered equal
The following example:

```sass
@debug calc(40px + 2rem) == calc(40px - 2rem)
```

returns `true` when should return `false`, since there are different operators (`+` and `-`), which in CSS will produce different results.

However, the following does return `false`:

```sass
@use "sass:meta"

@debug meta.calc-args(calc(40px + 2rem)) == meta.calc-args(calc(40px - 2rem))
```

But I believe we shouldn't have to rely on `meta.calc-args()` to do this check otherwise we would have to always be checking for type `calculation`.

---

Using Dart Sass 1.43.4

stof:
Fixes #1548

I haven't covered it in sass-spec for now, as sass-spec does not contain equality specs for values yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.43.5
 
+* Fix a bug where calculations with different operators were incorrectly
+  considered equal.
+
 ### JS API
 
 * Print more detailed JS stack traces. This is mostly useful for the Sass team's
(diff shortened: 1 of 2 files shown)


stof:
`hashCode` was already taking the operator into account properly

jathak (approved):
Thanks!

PR #1558, opened 2021-11-19
Add support for custom functions in the new JS API

nex3:
See sass/sass-spec#1744

### DIFF
--- a/lib/src/node/compile.dart
+++ b/lib/src/node/compile.dart
@@ -6,8 +6,12 @@ import 'package:js/js.dart';
 import 'package:node_interop/js.dart';
 import 'package:node_interop/util.dart' hide futureToPromise;
 import 'package:term_glyph/term_glyph.dart' as glyph;
+import 'package:tuple/tuple.dart';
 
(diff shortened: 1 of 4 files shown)


lib/src/node/compile.dart
```diff
@@ -40,7 +46,8 @@ NodeCompileResult compile(String path, [CompileOptions? options]) {
         sourceMap: options?.sourceMap ?? false,
         logger: NodeToDartLogger(options?.logger, Logger.stderr(color: color),
             ascii: ascii),
-        importers: options?.importers?.map(_parseImporter));
+        importers: options?.importers?.map(_parseImporter),
+        functions: _parseFunctions(options?.functions).cast());
```
Awjin, line 50: Does this need to be `cast`ed to a specific type?

Awjin (approved):

lib/src/node/compile.dart
nex3: No, it'll automatically infer the type of the parameter to which it's being passed.

Issue #1534, opened 2021-10-14
16x performance difference between sass and node-sass


dcwarwick:
We have a jest test for a component library that compiles a stylesheet and compares the result to a snapshot. This test is still using `node-sass`, although we've moved to using `sass` for our build, so I tried to update the test to use `sass` too.

The test looks something like this:
```js
const { renderSync } = require('node-sass');

it("doesn't change the exported CSS for released components", () => {
    expect(
      renderSync({
        file: resolve(__dirname, '../index-without-carbon-released-only.scss'),
        includePaths: [resolve(__dirname, '../../../../node_modules')],
        outputStyle: 'expanded',
      }).css.toString()
    ).toMatchSnapshot();
  });
```

Now this SCSS file loads, indirectly, a big and complex tree of dependent SCSS files, and on my machine it takes about 20 secs to complete the test:
![image](https://user-images.githubusercontent.com/6385315/137381136-5b40a5bc-ecbe-4aec-80f1-ceb24ed26df6.png)

I changed 'node-sass' to 'sass' on the require, as `sass` is already a dependency. I also added `* @jest-environment node` to the header comment. Then I ran the test again, and using `sass` instead of `node-sass` it takes about 345 secs to complete the test:
![image](https://user-images.githubusercontent.com/6385315/137381483-cc05e94c-44bd-4d14-b4de-e92547872dbc.png)

This is a very substantial increase in time, and indeed would cause the total run time of the entire test suite to grow by a factor of more than 3. I am aware that `sass` will not be quite as fast as `node-sass`, but this represents a much larger factor than I was expecting. I'm interested in whether this is what would be expected, and whether there is something about the way I am calling or using `sass` that could be changed to give me better/faster results.

The full test file: https://github.com/carbon-design-system/ibm-cloud-cognitive/blob/main/packages/cloud-cognitive/src/__tests__/styles.test.js

Steps to reproduce the issue:
- Clone the repo https://github.com/carbon-design-system/ibm-cloud-cognitive
- Run `yarn`
- Run `yarn jest styles`. This will run the test using `node-sass`.
- Edit the file /packages/cloud-cognitive/src/__tests__/styles.test.js to add `* @jest-environment node` into the copyright comment at the top of the file and to change 'node-sass' to 'sass' in the first `require`.
- Run `yarn jest styles` again. This will run the test using `sass`.

nex3:
[...]

PR #1573, opened 2021-12-15
Use a more recent version of Carbon for our benchmarks

nex3:
See sass/dart-sass#1534


jathak (approved):
Should we also update perf.md with a new round of benchmarks that uses this version?

PR #1592, opened 2022-01-08
Set LegacyResult.map to undefined rather than null

nex3:
See sass/sass-spec#1764

### DIFF
--- a/lib/src/node/legacy.dart
+++ b/lib/src/node/legacy.dart
@@ -357,7 +357,7 @@ RenderResult _newRenderResult(
   var end = DateTime.now();
 
   var css = result.css;
-  Uint8List? sourceMapBytes;
+  Uint8List? sourceMapBytes = undefined;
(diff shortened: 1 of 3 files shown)


lib/src/node/legacy.dart
```diff
@@ -357,7 +357,7 @@ RenderResult _newRenderResult(
   var end = DateTime.now();
 
   var css = result.css;
-  Uint8List? sourceMapBytes;
+  Uint8List? sourceMapBytes = undefined;
```
jathak, line 360: I think this needs to be `dynamic` to satisfy the type checker since `undefined` isn't null.

jathak (approved):

lib/src/node/legacy.dart
nex3: Oh, right, I guess https://github.com/pulyaevskiy/node-interop/pull/109 hasn't been released yet.

Issue #1604, opened 2022-01-17
Untruthful package.json engines.node?


webketje:
The package.json `engines.node` version is incorrect. While doing unit tests for a dependent library with nvm on Node 8.17.0 & Node 9.11.2 and sass 1.48.0, I got the following error:

```
TypeError: self.URL is not a constructor
    at StaticClosure.dartToJSUrl (/home/username/dev/path/to/repo/sass/node_modules/sass/sass.dart.js:22395:14)
```
I am not exactly sure why this happens, the only thing I know for sure is that it works without issues on Node 10.x.x +

Reproduce simply by creating a 2 files `main.scss` and `test.js` on Node <=9.x.x: 
```js
var sass= require('sass');
sass.compile('./main.scss');
```

PR #1612, opened 2022-01-26
Drop support for non-LTS Node (< 12).

Awjin:
Fixes https://github.com/sass/dart-sass/issues/1604

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.49.0
+version: 1.49.1
 description: A Sass implementation in Dart.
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.49.0
+version: 1.49.1
```
nex3, line 2: Let's make this a -dev version for now, since we're still in the process of setting up the Embedded Sass infrastructure.

nex3 (approved):

pubspec.yaml
Awjin: Right, good call.

Issue #1605, opened 2022-01-18
JS API changed or is broken in the latest versions?


waltercruz:
with sass < 1.45

`Welcome to Node.js v17.3.1.
Type ".help" for more information.
> const sass = require("sass");
> new sass.types.String('sass');
SassString {
  dartValue: SassString0 {
    text: 'sass',
    hasQuotes: false,
    '_string$__SassString_sassLength': null,
    '_string$_hashCache': null
  }
}`

With sass > 1.45

Welcome to Node.js v17.3.1.
Type ".help" for more information.
> const sass = require("sass");
undefined
> new sass.types.String('sass');
Uncaught Error [ERR_INTERNAL_ASSERTION]: NoSuchMethodError: method not found: 'call'
Receiver: Closure 'JSClassExtension_setCustomInspect_closure'
Arguments: [Instance of 'SassString0', 1, Instance of 'PlainJavaScriptObject', Instance of 'JavaScriptFunction']
    at Object.wrapException (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:1241:17)
    at JSClassExtension_setCustomInspect_closure.noSuchMethod$1 (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:32056:15)
    at Object.noSuchMethod$1$ (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:24396:42)
    at Object.Primitives_functionNoSuchMethod (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:1114:16)
    at Object.Primitives__generalApplyFunction (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:1167:18)
    at Object.Primitives_applyFunction (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:1146:16)
    at Object.Function_apply (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:5254:16)
    at _callDartFunctionFastCaptureThis (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:22966:16)
    at SassString0.<anonymous> (/Users/waltercruz/devel/src/node_modules/sass/sass.dart.js:22953:18)
    at formatValue (node:internal/util/inspect:780:19)
This is caused by either a bug in Node.js or incorrect usage of Node.js internals.
Please open an issue with this stack trace at https://github.com/nodejs/node/issues

running on node v17.3.1 on macOS Catalina

[...]

PR #1631, opened 2022-02-16
Allow Node 17 to pass extra args to inspect.

Awjin:
Fixes https://github.com/sass/dart-sass/issues/1625
Fixes https://github.com/sass/dart-sass/issues/1605

See https://github.com/sass/sass-spec/pull/1777

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.49.8
+
+* Fix a bug where `require('sass')` was broken on Node 17 due to Dart's JS
+  interop.
+
 ## 1.49.7
 
 ### Embedded Sass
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.49.8
+
+* Fix a bug where `require('sass')` was broken on Node 17 due to Dart's JS
+  interop.
```
nex3, line 4: This kind of makes it sound like loading the module at all was impossible. Maybe instead, "Fix a bug where inspecting the Sass module in the Node.js console crashed on Node 17."

Also, put this under a `## JS API` heading.

nex3 (approved):

PR #1633, opened 2022-02-16
Fix CSS import bug

### LINKED ISSUE #1628: CSS imports not emitted in some cases
Given something like:
```scss
<==> main.scss
@import "css-rules";
@import "css-import";

<==> css-rules.scss
a{b:c}

<==> css-import.scss
@use "sass:math";
@import url(//example.com)
```

Sass should emit
```css
@import url(//example.com);
a {
  b: c;
}
```
but it instead omits the CSS import.

This only seems to be the case when both of the following are true:
- the file containing the CSS import is imported after another file emitting CSS
- the file containing the CSS import contains a `@use` or `@forward` rule.

jathak:
Fixes #1628.

See https://github.com/sass/sass-spec/pull/1778

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.49.8-dev
+version: 1.49.8
 description: A Sass implementation in Dart.
 homepage: https://github.com/sass/dart-sass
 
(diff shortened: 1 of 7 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.49.7
+version: 1.49.8
```
nex3, line 2: Let's mark this as -dev for now, there are a number of other changes in flight.

nex3 (approved):

PR #1640, opened 2022-02-25
Remove dead code in the stylesheet parser

stof:
At that point, the scanner is still at the same position than the previous check for children returning a nested declaration, so it is
impossible for it to be looking at children.

As part of my rewrite of scssphp to a more modern and more spec-compliant codebase, I ported the dart-sass parser to PHP. This dead code was reported by our static analyzer in its updated version, and my own investigation of the code makes me agree with the tool.


stof:
@nex3 has this PR been missed ?

jathak (approved):
Thanks! Sorry for the delay in reviewing

PR #1651, opened 2022-03-15
Preserve rgb/rgba/hsl/hsla formats in expanded mode

### LINKED ISSUE #1634: Preserve rgb()/hsl() color format in expanded mode only
Now that >95% of browsers support `hsl()` ([source](https://caniuse.com/mdn-css_types_color_hsl)) and `hsla()` ([source](https://caniuse.com/mdn-css_types_color_alpha)), we can bring them into the set of color formats that we preserve in expanded mode for readability purposes.

This will require some amount of refactoring of how we handle "original colors". Currently we store the [`originalSpan`](https://github.com/sass/dart-sass/blob/c66ea3f9a53e00c8a81e9781c2e62c5940dbe6fe/lib/src/value/color.dart#L104) and use its text to determine the actual format of the color, but this won't work for color functions where the source span may contain variable references and so on. While we could create a synthetic span, it's probably cleaner to change the "original representation" to an enum of different possible formats (`longHex`, `shortHex`, `name` (which may need a string attached since some colors have multiple names), `rgb`, `hsl`) and use those to recreate the appropriate serialization format.

Note that we should serialize these colors using the comma-separated function formats even if the author used a space-separated format since that's not yet widely supported. Similarly, we should serialize colors with non-`1` alpha channels using `rgba()` and `hsla()` even though that's no longer necessary per spec.

**Note:** This issue does not cover emitting `hsl()` in the compressed output. We never intend to support that for the reasons outlined in https://github.com/sass/sass/issues/469, and attempts to argue that case in this issue will be marked as spam.

nex3:
This also fixes a bug where four- and eight-digit hex numbers weren't
being translated to more compatible formats.

Closes #1634
See sass/sass-spec#1780

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -2301,14 +2301,14 @@ abstract class StylesheetParser extends Parser {
 
     var first = scanner.peekChar();
     if (first != null && isDigit(first)) {
-      return ColorExpression(_hexColorContents(start));
+      return ColorExpression(_hexColorContents(start), scanner.spanFrom(start));
     }
 
     var afterHash = scanner.state;
     var identifier = interpolatedIdentifier();
     if (_isHexColor(identifier)) {
       scanner.state = afterHash;
-      return ColorExpression(_hexColorContents(start));
+      return ColorExpression(_hexColorContents(start), scanner.spanFrom(start));
     }
 
     var buffer = InterpolationBuffer();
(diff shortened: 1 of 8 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -2351,7 +2351,14 @@ abstract class StylesheetParser extends Parser {
       }
     }
 
-    return SassColor.rgb(red, green, blue, alpha, scanner.spanFrom(start));
+    return SassColor.rgbInternal(
+        red,
+        green,
+        blue,
+        alpha,
+        // Don't four- or eight-digit hex colors as hex, since that's not yet
```
jathak, line 2359: Missing a word here?

jathak (approved):

PR #1653, opened 2022-03-15
Stop simplifying calculations in supports

### LINKED ISSUE #1652: Don't simplify calculations within supports declarations
See sass/sass#3259.

jathak:
Fixes #1652.

See https://github.com/sass/sass-spec/pull/1781

### DIFF
--- a/lib/src/value/calculation.dart
+++ b/lib/src/value/calculation.dart
@@ -35,6 +35,13 @@ class SassCalculation extends Value {
   @internal
   bool get isSpecialNumber => true;
 
+  /// Creates a new calculation with the given [name] and [arguments]
+  /// that will not be simplified.
+  @internal
+  static Value unsimplified(String name, Iterable<Object> arguments) {
+    return SassCalculation._(name, List.unmodifiable(arguments));
+  }
+
   /// Creates a `calc()` calculation with the given [argument].
   ///
   /// The [argument] must be either a [SassNumber], a [SassCalculation], an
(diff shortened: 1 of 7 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -1949,9 +1954,12 @@ class _EvaluateVisitor
     } else if (condition is SupportsInterpolation) {
       return await _evaluateToCss(condition.expression, quote: false);
     } else if (condition is SupportsDeclaration) {
-      return "(${await _evaluateToCss(condition.name)}:"
+      _inSupportsDeclaration = true;
+      var result = "(${await _evaluateToCss(condition.name)}:"
           "${condition.isCustomProperty ? '' : ' '}"
           "${await _evaluateToCss(condition.value)})";
+      _inSupportsDeclaration = false;
```
stof, line 1961: this should restore the previous value instead of setting it to `false`

lib/src/value/calculation.dart
```diff
@@ -162,17 +169,23 @@ class SassCalculation extends Value {
...
+      operateInternal(operator, left, right,
+          inMinMax: false, inSupportsDeclaration: false);
 
   /// Like [operate], but with the internal-only [inMinMax] parameter.
   ///
   /// If [inMinMax] is `true`, this allows unitless numbers to be added and
   /// subtracted with numbers with units, for backwards-compatibility with the
   /// old global `min()` and `max()` functions.
+  ///
+  /// If [inSupportsDeclaration] is `true`, no simplification will be done.
```
nex3, line 181: I'd call this parameter `simplify`, since that more directly represents what it *does* rather than what the context is.

nex3 (approved):
Can you mark this as a release version?

PR #1662, opened 2022-04-05
Add support for :where()

jathak:
See https://github.com/sass/sass/pull/3283.

See https://github.com/sass/sass-spec/pull/1783.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.49.12
 
-* [`:where()`] now behaves similarly to `:is()`, except for |something complex about extend?...]
+* `@extend` now treats [`:where()`] the same as `:is()`.
 
 [`:where()`]: https://developer.mozilla.org/en-US/docs/Web/CSS/:where
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.49.12
+
+* [`:where()`] now behaves similarly to `:is()`, except for |something complex about extend?...]
+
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.49.12
```
nex3, line 1: I think this should bump the minor version, since it's effectively a new feature.

lib/src/extend/extension_store.dart
```diff
@@ -831,14 +831,15 @@ class ExtensionStore {
           // become `.foo:not(.bar)`. However, this is a narrow edge case and
           // supporting it properly would make this code and the code calling it
           // a lot more complicated, so it's not supported for now.
-          if (innerPseudo.normalizedName != 'is' &&
-              innerPseudo.normalizedName != 'matches') {
+          if (!{'is', 'matches', 'where'}
```
nex3, line 834:
```suggestion
          if (!const {'is', 'matches', 'where'}
```

nex3 (approved):

stof:
I find something weird in this PR: it seems to treat `where` exactly the same than `is`. But the reason why both exist is because `:where` has a specificity of 0, which I don't see in this PR

jathak:
While the browser treats them differently, there aren't really any cases where the specificity difference matters from Sass's perspective. For the one place where this was previously the case (the laws of extend), we decided to tweak the spec there since the previous language led to somewhat surprising output (that would also have been more difficult to implement here).

stof:
thanks. I missed the spec tweak.

PR #1664, opened 2022-04-06
Use NoOpImporter for compileString unless url is a `file:` url

ntkme:
Per discussion: https://github.com/sass/dart-sass-embedded/pull/83#issuecomment-1090905405

Spec: https://github.com/sass/sass/pull/3278


ntkme:
I'm not familiar with dealing with JS Object in Dart so maybe there is a better way to write this.

nex3:
Is this change necessary? https://github.com/sass/sass-spec/pull/1785 is already passing without it, because if `url` isn't a `file:` URL the default `FilesystemImporter` won't be able to resolve it anyway.

ntkme:
@nex3 Technically not because it is not going to resolve anyways. It was just for the sake of precisely matching what spec says.

PR #1665, opened 2022-04-09
Ensure onStdinClose only completes for non-tty streams

mcrumm:
First, thank you so much for all the work to get #1411 merged! I realized only too late that there was a small typo in my initial version that has led to a bit of a gnarly bug. 😳 

The initial stdin commit 144cd35 inverted the TTY conditional
b/w the `io/node` and `io/vm` implementations of `onStdinClose`.
The node implementation incorrectly checked for the presence
of TTY to complete the stream, while the vm implementation
correctly checked for its _absence_. The commit that landed
upstream c7ab426 normalized the incorrect behaviour, which
means that sass was still not closing when stdin was closed,
unless stdin was a TTY.

Unfortunately that created a "worst of both worlds" situation
because programs that start sass and then close unexpectedly
will still leave zombie sass processes running in the
background, and wrapper scripts designed to mitigate this
exact problem will stop working because moving the process to
the background now incorrectly causes the job to stop.

This change ensures we only complete the CancelableOperation
onStdinClose for non-tty standard input streams.

### DIFF
--- a/lib/src/io/interface.dart
+++ b/lib/src/io/interface.dart
@@ -95,10 +95,12 @@ String? getEnvironmentVariable(String name) => throw '';
 int get exitCode => throw '';
 set exitCode(int value) => throw '';
 
-/// If stdin is a TTY, returns a [CancelableOperation] that completes once it
-/// closes.
+/// Returns a [CancelableOperation] that completes when stdin closes.
 ///
-/// Otherwise, returns a [CancelableOperation] that never completes.
+/// Note that if stdin is a TTY, the operation never completes. This is to
+/// avoid interfering with background job systems where reading from stdin
+/// and then moving the process to the background would incorrectly cause
+/// the job to stop.
 ///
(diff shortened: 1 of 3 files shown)


nex3:
Can you add a changelog entry as well?

nex3:
Thanks! Are you able to figure out why the Windows tests are failing? It seems unlikely to be a flake.

mcrumm:
@nex3 I can test Windows in a VM which is not ideal but it works :) I agree there's an issue and I will try to figure it out– right now running `dart test` from PowerShell hangs forever on a `--watch` test.

nex3:
@mcrumm If you don't have time to debug this for now, maybe we should roll back #1411 for now?

mcrumm:
@nex3 Yes I think that's likely the best course. Other responsibilities have unfortunately taken precedence for the moment, but I would very much like to get back to this as time permits :)

nex3 (changes requested):
Marking as Request Changes pending the windows debugging.

PR #1675, opened 2022-04-16
Add support for layer condition in CSS imports

stof:
Closes https://github.com/sass/sass/issues/3285

See https://github.com/sass/sass/pull/3290
See https://github.com/sass/sass-spec/pull/1787

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -1089,7 +1089,9 @@ abstract class StylesheetParser extends Parser {
       var queries = tryImportQueries();
       return StaticImport(Interpolation([url], scanner.spanFrom(start)),
           scanner.spanFrom(start),
-          supports: queries?.item1, media: queries?.item2);
+          layer: queries?.item3,
+          supports: queries?.item1,
+          media: queries?.item2);
     }
 
     var url = string();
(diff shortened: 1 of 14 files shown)


stof:
If #1671 is merged first, changelogs will need to be adapted.

stof:
The failing test is because the testsuite does not accept `1.0.0-beta.42-dev` as version in the sass_api package, expecting `1.0.0-dev` instead. It looks like it does not like when the dev version says that we will still have a beta release coming.

lib/src/parse/stylesheet.dart
```diff
@@ -1135,10 +1139,33 @@ abstract class StylesheetParser extends Parser {
     return url.startsWith("http://") || url.startsWith("https://");
   }
 
-  /// Consumes a supports condition and/or a media query after an `@import`.
+  /// Consumes a supports condition and/or a media query and/or a layer after an `@import`.
   ///
   /// Returns `null` if neither type of query can be found.
-  Tuple2<SupportsCondition?, Interpolation?>? tryImportQueries() {
+  Tuple3<SupportsCondition?, Interpolation?, Interpolation?>?
```
stof, line 1145: here I added the layer as third item in the tuple, to reduce the diff in places using this function. But I'm wondering whether the layer should come first instead, to be in the same order than in CSS.

stof: @nex3 any preference about the order here ?

PR #1676, opened 2022-04-16
Escape the source map URL before using it in a CSS comment

### LINKED ISSUE #1351: Closing block comment `*/` for embedded source content produces invalid css and breaks source map
A comment block, e.g., for a license block, from source `in.scss`:
```scss
/* comment block */
```
generating `out.css` with `1.34.1 compiled with dart2js 2.13.1` and options `--source-map --embed-sources --embed-source-map`:
```css
/* comment block */

/*# sourceMappingURL=data:application/json;charset=utf-8,%7B%22version%22:3,%22sourceRoot%22:%22%22,%22sources%22:%5B%22in.scss%22%5D,%22names%22:%5B%5D,%22mappings%22:%22AAAA%22,%22file%22:%22out.css%22,%22sourcesContent%22:%5B%22/*%20comment%20block%20*/%5Cn%22%5D%7D */
```

results in a sourceMapURL parsed as `data:application/json;charset=utf-8,%7B%22version%22:3,%22sourceRoot%22:%22%22,%22sources%22:%5B%22in.scss%22%5D,%22names%22:%5B%5D,%22mappings%22:%22AAAA%22,%22file%22:%22out.css%22,%22sourcesContent%22:%5B%22/*%20comment%20block%20`

Notably `*/%5Cn%22%5D%7D` is missing from that URI.

If the JSON:
```json
{
	"version": 3,
	"sourceRoot": "",
	"sources": ["in.scss"],
	"names": [],
	"mappings": "AAAA",
	"file": "out.css",
	"sourcesContent": ["/* comment block */\n"]
}
```
was base64 encoded similar to #1022 for a URI:
```
data:application/json;base64,eyJ2ZXJzaW9uIjozLCJzb3VyY2VSb290IjoiIiwic291cmNlcyI6WyJpbi5zY3NzIl0sIm5hbWVzIjpbXSwibWFwcGluZ3MiOiJBQUFBIiwiZmlsZSI6Im91dC5jc3MiLCJzb3VyY2VzQ29udGVudCI6WyIvKiBjb21tZW50IGJsb2NrICovXG4iXX0=
```
the source map loads correctly.

stof:
The URL must not be allowed to terminate the comment.

Closes #1351

### DIFF
--- a/lib/src/executable/compile_stylesheet.dart
+++ b/lib/src/executable/compile_stylesheet.dart
@@ -174,6 +174,8 @@ String _writeSourceMap(
     url = p.toUri(p.relative(sourceMapPath, from: p.dirname(destination)));
   }
 
+  var escapedUrl = url.toString().replaceAll("*/", '%2A/');
+
   return (options.style == OutputStyle.compressed ? '' : '\n\n') +
-      '/*# sourceMappingURL=$url */';
+      '/*# sourceMappingURL=$escapedUrl */';
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 * Fix an issue where source locations tracked through variable references could
   potentially become incorrect.
+* Fix a bug where a loud comment in the source can break the source map when
+  embedding the sources, when using the command-line interface or the legacy JS
+  API.
 
 ## 1.51.0
 
--- a/lib/src/node/legacy.dart
+++ b/lib/src/node/legacy.dart
@@ -405,7 +405,8 @@ RenderResult _newRenderResult(
           : p.toUri(outFile == null
               ? sourceMapPath
               : p.relative(sourceMapPath, from: p.dirname(outFile)));
-      css += "\n\n/*# sourceMappingURL=$url */";
+      var escapedUrl = url.toString().replaceAll("*/", '%2A/');
+      css += "\n\n/*# sourceMappingURL=$escapedUrl */";
     }
   }
 


stof:
@jathak anything missing here ?

lib/src/executable/compile_stylesheet.dart
```diff
@@ -174,6 +174,8 @@ String _writeSourceMap(
     url = p.toUri(p.relative(sourceMapPath, from: p.dirname(destination)));
   }
 
+  var escapedUrl = url.toString().replaceAll("*/", "*\\/");
```
jathak, line 177: This should be URL encoded, not backslash escaped (same below)

jathak (commented):
Sorry for the delay! Missed this one when I was going through the list of PRs that needed my review

lib/src/executable/compile_stylesheet.dart
stof: the issue is not about escaping for usage in a URL. It is about escaping it for usage in a CSS comment. This applies *CSS* escaping

jathak: After looking into it, I realize that URL encoding won't work with data URLs (since the percent encoding then ends up in the resulting contents), but I'm not sure backslash escaping will work either.

jathak: Actually, nevermind. Percent encoding does work, I was just doing it wrong. `data:,/*comment%2A/` correctly resolves to `/*comment*/`, so the answer here is probably `var escapedUrl = url.toString().replaceAll('*/', '%2A/');`

stof:
@jathak updated

PR #1681, opened 2022-04-23
Fix the tracking of variable nodes for scopes

stof:


nex3:
Can you give some more context here? Is this causing behavioral problems? Should it have a spec added for it?

stof:
The only effect I could think of is a bad sourcemap or a bad error location, as that's all nodes are used for.
I haven't tried to produce a broken behavior based on this. But the current code expects the lists of scopes in `variables` and `variableNodes` to be in sync, while `variableNodes` never leaves scopes. So it is likely to find a wrong span for reporting errors if a variable has the same name than a variable in a previous scope at the same depth.

PR #1682, opened 2022-04-24
Fix the string representations of the Sass AST

stof:
Those are used when providing recommendations in warnings or errors, so it is better if they produce valid code.


jathak (approved):
Thanks!

PR #1695, opened 2022-05-18
Add support for arbitrary modifiers after @import

nex3:
See sass/sass#3285
See sass/sass#3290
See sass/sass-spec#1787

### DIFF
--- a/lib/src/ast/sass/supports_condition/anything.dart
+++ b/lib/src/ast/sass/supports_condition/anything.dart
@@ -21,5 +21,8 @@ class SupportsAnything implements SupportsCondition {
 
   SupportsAnything(this.contents, this.span);
 
+  Interpolation toInterpolation() =>
+      Interpolation.concat(['(', contents, ')'], span);
+
   String toString() => "($contents)";
 }
--- a/lib/src/ast/sass/supports_condition/anything.dart
+++ b/lib/src/ast/sass/supports_condition/anything.dart
@@ -21,8 +21,5 @@ class SupportsAnything implements SupportsCondition {
 
   SupportsAnything(this.contents, this.span);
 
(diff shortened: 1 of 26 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -1135,54 +1135,100 @@ abstract class StylesheetParser extends Parser {
...
+            buffer.addInterpolation(_interpolatedDeclarationValue(
+                allowEmpty: true, allowSemicolon: true));
+            buffer.writeCharCode($rparen);
+          }
+
+          scanner.expectChar($rparen);
+          whitespace();
+        } else {
+          whitespace();
+          if (scanner.scanChar($comma)) {
```
stof, line 1176: This support for comma looks incompatible with using the comma as separator between multiple ImportArgument.

lib/src/ast/sass/supports_condition/anything.dart
```diff
@@ -21,5 +21,8 @@ class SupportsAnything implements SupportsCondition {
 
   SupportsAnything(this.contents, this.span);
 
+  Interpolation toInterpolation() =>
+      Interpolation.concat(['(', contents, ')'], span);
```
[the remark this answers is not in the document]
nex3: Ah, good catch. Instead of converting these to interpolation, I've added an expression-level AST node just for this.

lib/src/parse/stylesheet.dart
nex3: This actually exists to match the previous behavior: because `MediaQueryList` can contain commas at nearly any point, you could already write `@import "..." foo, bar, baz`. This just makes it more explicit.

jathak (approved):

PR #1723, opened 2022-06-16
Add support for `var()` with an empty fallback argument

nex3:
Closes sass/sass#3245
See sass/sass-spec#1794

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -117,7 +117,7 @@ abstract class StylesheetParser extends Parser {
         return arguments;
       });
 
-  Expression parseExpression() => _parseSingleProduction(expression);
+  Expression parseExpression() => _parseSingleProduction(_expression);
 
   VariableDeclaration parseVariableDeclaration() =>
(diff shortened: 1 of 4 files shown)


lib/src/parse/stylesheet.dart
```diff
@@ -1688,7 +1702,7 @@ abstract class StylesheetParser extends Parser {
   /// still be a valid expression. When it returns `true`, this returns the
   /// expression.
   @protected
```
stof, line 1704: shouldn't `@protected` is removed here ? This method is now private

lib/src/parse/stylesheet.dart
```diff
@@ -2141,7 +2155,7 @@ abstract class StylesheetParser extends Parser {
   ///
   /// If [singleEquals] is true, this will allow the Microsoft-style `=`
   /// operator at the top level.
-  Expression _expressionUntilComma({bool singleEquals = false}) => expression(
+  Expression expressionUntilComma({bool singleEquals = false}) => _expression(
```
stof, line 2158: shouldn't `@protected` be added on this method now ?

PR #1736, opened 2022-06-28
Fix sass types for TS 4.7+ Node16/NodeNext module resolution

### LINKED ISSUE #1714: TypeScript types not accessible with new TS 4.7 Node16/next resolution
As of TypeScript 4.7, the `exports` field can be respected when the new resolution (`node16` or `nodenext`) is used. The Sass npm package uses `exports` but does not set the `types` condition. This results in `d.ts` resolution errors like the following:

```
Could not find a declaration file for module 'sass'. '/X/node_modules/sass/sass.default.dart.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/sass` if it exists or add a new declaration (.d.ts) file containing `declare module 'sass';`

3 import sass from 'sass';
```

One might wonder why this happens when the `sass/package.json` sets the `types` field. This field is ignored when the `exports` are configured and the `default` condition is pointing to the `.js` file, preventing `.d.ts` from being discovered.

The fix is to add another `types` condition that points to the index `.d.ts` file. This seems to require changing `dart_cli_pkg` to either add this per config-setting, or to merge the `exports` from the user-provided "base package.json` file.

Goodwine:
Fixes #1714

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -17,10 +17,20 @@
     "node": ">=12.0.0"
   },
   "dependencies": {
+    "@types/node": "^18.0.0",
     "chokidar": ">=3.0.0 <4.0.0",
     "immutable": "^4.0.0",
     "source-map-js": ">=0.6.2 <2.0.0"
   },
-  "keywords": ["style", "scss", "sass", "preprocessor", "css"],
-  "types": "types/index.d.ts"
+  "keywords": [
+    "style",
+    "scss",
+    "sass",
+    "preprocessor",
+    "css"
+  ],
+  "types": "types/index.d.ts",
+  "exports": {
+    "types": "./types/index.d.ts"
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.54.0
+
+### JS API
+
+* Fix Sass npm package types for TS 4.7+ Node16 and NodeNext module resolution.
+
 ## 1.53.0
 
 * Add support for calling `var()` with an empty second argument, such as
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.53.0
+version: 1.54.0-dev
 description: A Sass implementation in Dart.
 homepage: https://github.com/sass/dart-sass
 
@@ -34,7 +34,7 @@ dependencies:
 dev_dependencies:
   analyzer: ^3.0.0
   archive: ^3.1.2
-  cli_pkg: ^2.1.0
+  cli_pkg: ^2.1.4
   crypto: ^3.0.0
   dart_style: ^2.0.0
   dartdoc: ^5.0.0


package/package.json
```diff
@@ -21,6 +21,18 @@
     "immutable": "^4.0.0",
     "source-map-js": ">=0.6.2 <2.0.0"
   },
-  "keywords": ["style", "scss", "sass", "preprocessor", "css"],
-  "types": "types/index.d.ts"
+  "devDependencies": {
+    "@types/node": "^18.0.0"
+  },
```
nex3, line 26: `devDependencies` doesn't make sense—those aren't used by packages that depend on this package, and this `package.json` isn't ever used for local development. I _think_ this should just be normal `dependencies`.

nex3 (changes requested):

package/package.json
ntkme: @nex3 @Goodwine I don't think this should be added at as `dependencies`, because it is the user's choice to use TS or JS. When using pure JS, we shouldn't need any types to be installed as a dependency.If this is affecting the TS compilation of this project, it should indeed be a `devDependencies`.

Goodwine: Wouldn't the `@types/node` package be downloaded regardless of whether it being declared as `dependencies` vs `devDependencies`? or do devDependencies get skipped if `sass` gets declared as non-dev dependency?

nex3: > do devDependencies get skipped if sass gets declared as non-dev dependency?

This is correct, unless I'm severely misunderstanding. Sometimes it's necessary to include a dependency that's not used by some users because it's necessary for others.

nex3 (approved):

package/package.json
devversion: FWIW: Node packages (even when shipping `.d.ts`) usually never add a dependency to satisfy TypeScript type-checking for users. Instead:

* Users either have to set up `@types/node` themselves, or the library authors place a directive implying the necessary type like: `/// <reference types="node" />` into their `.d.ts`
* Or, users most of the time just disable type-checking of external libraries. e..g by setting `skipLibCheck: true`.

Also just to make this clear: The `devDependencies` will never be installed for users, only for the authors of the package when e.g. `yarn`/`npm` is executed inside the folder containing the `package.json`.

TL;DR: Adding `@types/node` as a dependency is pretty uncommon. It should be removed. It's also not a big deal though and doesn't impact anyone negatively.

devversion:
Thanks for fixing this! I just added a little comment above (which you may want to consider, or not)

package/package.json
nex3: If including `@types/node` doesn't impact anyone negatively, and means that users can freely use the TS types without needing to manually install additional dependencies, why _not_ include it?

ntkme: Well, this may become an addition to the reasons we get bloated node_modules.

Also, having “@types/node” lock to version 18 is not really a good idea given that end users might want to install other versions (e.g. 16) if that is the version of node they are using.

devversion: Mainly for good practice of not increasing the dependency tree for people not using the TS types, which is very likely the majority here. This is peanuts in size, but a general question of best practice.

You also don't have a peer dependency/ or dependency on TypeScript itself, making guarantees with what version your types work 🧐)

I don't care personally, but this is uncommon. Angular for example doesn't do it.

nex3: > Also, having `@types/node` lock to version 18 as part of `dependencies` is not really a good idea given that end users might want to install other versions (e.g. 16) if that is the version of node they are using.

This is a good point—we should look into making the version range here as wide as possible while still including the APIs we use (I think only `Buffer`?).

> You also don't have a peer dependency/ or dependency on TypeScript itself, making guarantees with what version your types work

Ideally, we'd be able to specify a dependency of the form "*if* the downstream user has TypeScript installed, *then* it must have this version", but unfortunately peer dependencies will produce warnings even if the downstream user doesn't have TypeScript.

PR #1737, opened 2022-07-06
Remove @types/node from dependencies

ntkme:
Based on comments from https://github.com/sass/dart-sass/pull/1736.

It seems to be common to use `"@types/node": "*"` when `@types/node` is required.

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -17,7 +17,6 @@
     "node": ">=12.0.0"
   },
   "dependencies": {
-    "@types/node": "^18.0.0",
     "chokidar": ">=3.0.0 <4.0.0",
     "immutable": "^4.0.0",
     "source-map-js": ">=0.6.2 <2.0.0"


Goodwine:
Interestingly when using a test package with:
```json
  "dependencies": {
    "sass": "file:../<PATH_TO_DART_SASS>/build/npm"
  },
```
And then compiling `import * as sass from 'sass'` using `NodeNext` module/moduleResolution I get the same typing error regardless of whether the sass' package dependency 🤔 

in other words it makes no difference whether the dependency is there or not.. perhaps it'd be better to remove it all together as suggested?
The reason I think is because `@types/node` exists within `node_modules/sass/node_modules` and TS requires the typings to exist under `node_modules/<>` or something like that 🤔

ntkme:
@Goodwine I would prefer to remove it all together as well as it is more common for end users to install it themselves anyways.

Goodwine (approved):

Issue #1716, opened 2022-06-11
Allow `@forward` multiple times without `with ()`


KaelWD:
Reference: https://github.com/vuetifyjs/vuetify/issues/14446

```scss
// main.scss
@use 'library' with (
  $var: 'value',
);

// library.scss
@forward 'component-a';
@forward 'component-b';

// component-a.scss
@forward './variables';

// component-b.scss
@forward './variables';

// _variables.scss
$var: 'default' !default;
```

`with` is only used in main.scss, so the second forward can just be ignored instead of throwing

```
Error: This module was already loaded, so it can't be configured using "with".
  ┌──> component-b.scss
1 │   @forward './variables'
  │   ^^^^^^^^^^^^^^^^^^^^^^ new load
  ╵
  ┌──> component-a.scss
1 │   @forward './variables'
  │   ━━━━━━━━━━━━━━━━━━━━━━ original load
  ╵
  ┌──> main.scss
1 │ ┌ @use 'library' with (
2 │ │   $var: 'value',
3 │ │ );
  │ └─^ configuration
  ╵
  component-b.scss 1:1  @forward
  library.scss 2:1      @use
  main.scss 1:1         root stylesheet
```

nex3:
I can't reproduce this. What version of Sass are you using?

Goodwine:
If this is from vuetify, they seem to be using `1.32.13`, I did a quick test and found that this exact code sample does indeed fail at that 1-year-old version and it actually fails for `1.33.0` too.

However this code snippet does not fail anymore from `1.34.0` onwards

KaelWD:
We're only using the new module system in v3 (`next` branch), v2 is stuck on sass 1.32 because the division changes aren't compatible with the way consumers were overriding variables (`@import` in prependData).

Here's a minimal reproduction of it with sass 1.52.3 and no other libraries: https://github.com/KaelWD/sass-forward-twice
Seems to require the second variable that isn't defined in the first file, if you comment out `main.scss:3` and `_index.scss:4-5` it doesn't error. 
Trying to override a non-existent variable with this setup also throws `This module was already loaded` instead of the expected `This variable was not declared with !default in the @used module`

Goodwine:
After debugging and discussing with @nex3 we found this behavior to be "working as specified" but we agreed the desired behavior from this usecase is correct and we'll first need to change the spec, and _then_ implement the fix

I'll mark this as blocked for now until that change in spec happens

Goodwine:
[...]

PR #1739, opened 2022-07-14
Allow a module loaded multiple times by the same configuration

### LINKED ISSUE #1712: Got range error
Hello.

When I updated the dart-sass from 1.52.1 to 1.52.2, I got the error below.
`Error: RangeError: Invalid value: Not in inclusive range 0..203: -1`

Could you have a look? Thanks!

This is the part of ERROR Summary.

```
  - name: Error
  - nodeAnnotation: [undefined]
  - nodeName: SassCompiler
  - originalErrorMessage: Error: RangeError: Invalid value: Not in inclusive range 0..203: -1
  - stack: Error: RangeError: Invalid value: Not in inclusive range 0..203: -1
    at Object.wrapException (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:1234:17)
    at Interceptor.lastIndexOf$2 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:25924:17)
    at _SerializeVisitor0._serialize0$_isTrailingComment$2 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:91405:50)
    at _SerializeVisitor0._serialize0$_visitChildren$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:91353:19)
    at _SerializeVisitor0.visitCssStyleRule$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:90622:13)
    at ModifiableCssStyleRule0.accept$1$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:92659:22)
    at ModifiableCssStyleRule0.accept$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:92662:19)
    at _SerializeVisitor0.visitCssStylesheet$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:90536:19)
    at ModifiableCssStylesheet0.accept$1$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:92709:22)
    at ModifiableCssStylesheet0.accept$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:92712:19)
    at Object.serialize0 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:21669:12)
    at /Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:15367:35
    at _wrapJsFunctionForAsync_closure.$protected (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:3726:15)
    at _wrapJsFunctionForAsync_closure.call$2 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:27983:12)
    at _awaitOnObject_closure.call$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:27971:32)
    at _RootZone.runUnary$2$2 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:29638:18)
    at _Future__propagateToListeners_handleValueCallback.call$0 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:28452:51)
    at Object._Future__propagateToListeners (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:3897:93)
    at _Future._completeWithValue$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:28311:9)
    at _AsyncAwaitCompleter.complete$1 (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:27958:14)
    at Object._asyncReturn (/Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:3698:17)
    at /Users/yusukegoto/src/immo/immo-platform/node_modules/sass/sass.dart.js:72844:24
```

Edited log file
[dart-sass-range-error.log](https://github.com/sass/dart-sass/files/8844426/dart-sass-range-error.log)

Goodwine:
Fixes https://github.com/sass/dart-sass/issues/1716

See https://github.com/sass/sass-spec/pull/1796

Blocked on
- https://github.com/sass/sass/pull/3357

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
    `random()` currently ignores units. A future version will no longer ignore
   units.
 
+* Don't throw an error when the same module is `@forward`ed multiple times
+  through a configured module.
+
 ## 1.54.4
 
 * Improve error messages when passing incorrect units that are also
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -28,20 +28,21 @@ class Configuration {
   final Map<String, ConfiguredValue> _values;
 
   /// Creates an implicit configuration with the given [values].
-  Configuration.implicit(this._values);
+  Configuration.implicit(this._values, {String? opaqueId})
+      : this.opaqueId = opaqueId ?? Xid.string();
 
   /// Uniquely identifying ID for a configuration lasting across the execution
   /// context.
   ///
   /// Implicit configurations will always have different IDs.
-  int get opaqueId => Xid.string().hashCode;
+  final String opaqueId;
 
   /// The empty configuration, which indicates that the module has not been
   /// configured.
   ///
   /// Empty configurations are always considered implicit, since they are
   /// ignored if the module has already been loaded.
-  const Configuration.empty() : _values = const {};
+  factory Configuration.empty() => Configuration.implicit({});
 
   bool get isEmpty => values.isEmpty;
 
@@ -52,7 +53,7 @@ class Configuration {
 
   /// Creates a new configuration from this one based on a `@forward` rule.
   Configuration throughForward(ForwardRule forward) {
-    if (isEmpty) return const Configuration.empty();
+    if (isEmpty) return Configuration.empty();
     var newValues = _values;
 
     // Only allow variables that are visible through the `@forward` to be
@@ -69,12 +70,13 @@ class Configuration {
     } else if (hiddenVariables != null && hiddenVariables.isNotEmpty) {
       newValues = LimitedMapView.blocklist(newValues, hiddenVariables);
     }
-    return _withValues(newValues);
+    return _withValues(newValues, opaqueId: opaqueId);
   }
 
   /// Returns a copy of [this] with the given [values] map.
-  Configuration _withValues(Map<String, ConfiguredValue> values) =>
-      Configuration.implicit(values);
+  Configuration _withValues(Map<String, ConfiguredValue> values,
+          {String? opaqueId}) =>
+      Configuration.implicit(values, opaqueId: opaqueId);
 
   String toString() =>
       "(" +
@@ -95,15 +97,13 @@ class ExplicitConfiguration extends Configuration {
   /// The node whose span indicates where the configuration was declared.
   final AstNode nodeWithSpan;
 
-  ExplicitConfiguration(Map<String, ConfiguredValue> values, this.nodeWithSpan)
-      : super.implicit(values);
-
-  /// ID for a [configuration] created with an explicit `@use ... with` uniquely
-  /// identified by its [nodeWithSpan] property.
-  @override
-  int get opaqueId => nodeWithSpan.span.toString().hashCode;
+  ExplicitConfiguration(Map<String, ConfiguredValue> values, this.nodeWithSpan,
+      {String? opaqueId})
+      : super.implicit(values, opaqueId: opaqueId);
 
   /// Returns a copy of [this] with the given [values] map.
-  Configuration _withValues(Map<String, ConfiguredValue> values) =>
-      ExplicitConfiguration(values, nodeWithSpan);
+  @override
+  Configuration _withValues(Map<String, ConfiguredValue> values,
+          {String? opaqueId}) =>
+      ExplicitConfiguration(values, nodeWithSpan, opaqueId: opaqueId);
 }
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -29,6 +29,12 @@ class Configuration {
   /// Creates an implicit configuration with the given [values].
   Configuration.implicit(this._values);
 
+  /// Uniquely identifying ID for a configuration lasting across the execution
+  /// context.
+  ///
+  /// Implicit configurations will always have different IDs.
+  int get opaqueId => identityHashCode(this);
+
   /// The empty configuration, which indicates that the module has not been
   /// configured.
   ///
@@ -90,6 +96,11 @@ class ExplicitConfiguration extends Configuration {
 
   ExplicitConfiguration(Map<String, ConfiguredValue> values, this.nodeWithSpan)
       : super.implicit(values);
+
+  /// ID for a [configuration] created with an explicit `@use ... with` uniquely
+  /// identified by its [nodeWithSpan] properties.
+  @override
+  int get opaqueId => nodeWithSpan.span.toString().hashCode;
 
   /// Returns a copy of [this] with the given [values] map.
   Configuration _withValues(Map<String, ConfiguredValue> values) =>
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -99,7 +99,6 @@ class Configuration {
   /// [_originalConfiguration] reference.
   Configuration._(this._values, this.__originalConfiguration);
 
-  @override
   String toString() =>
       "(" +
       values.entries
@@ -133,7 +132,6 @@ class ExplicitConfiguration extends Configuration {
   /// Returns a copy of `this` with the given [values] map.
   ///
   /// The copy will have the same [_originalConfiguration] as `this` config.
-  @override
   Configuration _withValues(Map<String, ConfiguredValue> values) =>
       ExplicitConfiguration._(values, nodeWithSpan, _originalConfiguration);
 }
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -3,7 +3,7 @@
 // https://opensource.org/licenses/MIT.
 
 import 'dart:collection';
-import 'package:xid/xid.dart';
+import 'package:uuid/uuid.dart';
 
 import 'ast/node.dart';
 import 'ast/sass.dart';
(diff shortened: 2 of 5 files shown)


lib/src/visitor/async_evaluate.dart
```diff
@@ -140,6 +140,9 @@ class _EvaluateVisitor
   /// All modules that have been loaded and evaluated so far.
   final _modules = <Uri, Module>{};
 
+  /// Configuration ids seen by a module by URI.
+  final _moduleConfigurationIds = <Uri, int>{};
```
nex3, line 144: Rather than explicitly translating configurations to integers, why not just store the `Configuration` object itself as the value here?

lib/src/configuration.dart
```diff
@@ -29,6 +29,12 @@ class Configuration {
   /// Creates an implicit configuration with the given [values].
   Configuration.implicit(this._values);
 
+  /// Uniquely identifying ID for a configuration lasting across the execution
+  /// context.
+  ///
+  /// Implicit configurations will always have different IDs.
+  int get opaqueId => identityHashCode(this);
```
nex3, line 36: Hash codes aren't guaranteed to be globally unique, although it may not be relevant; see below.

lib/src/configuration.dart
```diff
@@ -91,6 +97,11 @@ class ExplicitConfiguration extends Configuration {
   ExplicitConfiguration(Map<String, ConfiguredValue> values, this.nodeWithSpan)
       : super.implicit(values);
 
+  /// ID for a [configuration] created with an explicit `@use ... with` uniquely
+  /// identified by its [nodeWithSpan] properties.
+  @override
+  int get opaqueId => nodeWithSpan.span.toString().hashCode;
```
nex3, line 103: Are there situations where we have multiple distinct `Configuration` objects from the same node that we want to consider equivalent?

lib/src/configuration.dart
Goodwine: Per my understanding the `identityHashCode` returns the `Object.hashCode` regardless of the object overriding it, and it should generate a different one for each instance even if they have the exact same values 🤔

lib/src/configuration.dart
Goodwine: When testing this every time a module was multiple times with "the same config" it was actually behaving as if the `Configuration` object was a completely different instance/reference (as pointed above this might have been my misunderstanding on how `identityHashcode` works, but.. assuming my understanding _was_ correct then this is the reason I was hoping to consider this value equivalent

But I'll double check this :)

lib/src/visitor/async_evaluate.dart
Goodwine: Upon further testing, indeed the instances are different, I could still store the original Configuration and just compare their IDs later (or actually implement an `==` override), thoughts?

lib/src/configuration.dart
Goodwine: double checked and that's indeed the behavior, the instances are different even though it's the same config

lib/src/configuration.dart
nex3: It may be the case that `identityHashCode` produces unique integers for each object today, but the contract of a hash code is only good enough to make it work for a hash table, which means that collisions are allowed as long as they're sufficiently unlikely. The documentation for [`Object.hashCode`](https://api.dart.dev/stable/dart-core/Object/hashCode.html) even explicitly says:

> Objects that are not equal are allowed to have the same hash code.

lib/src/visitor/async_evaluate.dart
nex3: I guess it makes sense that object equality isn't sufficient, since we create derivative configurations for `@forward` rules. Let's nail down the spec before we continue here so we at least know the semantics we're trying to fit.

lib/src/configuration.dart
Goodwine: I was trying to avoid importing a new dependency since dart doesn't have a way of doing it (flutter does), but I ended up using `xid`

lib/src/visitor/async_evaluate.dart
Goodwine: Updated PR following suggestions on spec to get an idea how that looks like

lib/src/configuration.dart
```diff
@@ -3,6 +3,7 @@
 // https://opensource.org/licenses/MIT.
 
 import 'dart:collection';
+import 'package:uuid/uuid.dart';
```
nex3, line 6: I don't think using a full UUID here is necessary. UUIDs are mostly valuable for propagating the notion of "opaque ID" across network or process boundaries. Within a given process, you can call `Object()` and get something that will be `==` to only itself. This will also allow `Configuration.empty()` to remain const.

lib/src/configuration.dart
```diff
@@ -27,14 +28,21 @@ class Configuration {
   final Map<String, ConfiguredValue> _values;
 
   /// Creates an implicit configuration with the given [values].
-  Configuration.implicit(this._values);
+  Configuration.implicit(this._values, {String? opaqueId})
```
nex3, line 31: Rather than passing in the opaque ID itself, it might be better to make the field private, pass in `originalConfiguration`, and expose it publicly as a `bool sameOriginal(Configuration other)`. That way you encapsulate the details of exactly how we track originality, and expose only the core logic the caller needs to specify "this is the same as that".

lib/src/visitor/async_evaluate.dart
```diff
@@ -141,6 +141,9 @@ class _EvaluateVisitor
   /// All modules that have been loaded and evaluated so far.
   final _modules = <Uri, Module>{};
 
+  /// Configuration seen by a module URI.
```
nex3, line 144: "seen by" is unclear. Maybe "used to load"? It's probably also worth clarifying that these are the *first* configurations used to load each module.

nex3 (changes requested):

lib/src/configuration.dart
Goodwine: I stayed away from that to be more within an "opaque ID" value which in my mind looked like a string, but that was also because how the current set of constructors are organized.

I reorganized the constructors and which allowed removing the public references to the original config be hidden

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 * Properly consider `b > c` to be a superselector of `a > b > c`, and similarly
   for other combinators.
 
+* Allow loading the same module more than once from the same module
+  configuration.
```
nex3, line 10: This is a bit confusing because users might expect "the same configuration" to mean "structurally equivalent" rather than the hierarchical-identity scheme we use here. From a user's perspective, this is more like "Don't throw an error when the same module is `@forward`ed multiple times through a configured module".

lib/src/configuration.dart
```diff
@@ -27,14 +26,32 @@ class Configuration {
...
+  /// Returns whether [this] and [that] [Configuration]s have the same
+  /// [_originalConfiguration].
+  ///
+  /// An implicit configuration will always return `false` because it was not
+  /// created through another configuration.
+  ///
+  /// [ExplicitConfiguration]s will and configurations created [throughForward]
+  /// will be considered to have the same original config if they were created
+  /// as a copy from the same base configuration.
+  bool sameOriginal(Configuration that) =>
```
nex3, line 43: It seems like this should be reflexive—that is, a configuration should be said to have the same original configuration as itself. Maybe `.implicit()` and `.empty()` should set `_originalConfiguration` to `this` instead of null? That would allow you to remove some complexity with the `withValues()`/`copy()` method, too.

nex3 (changes requested):

lib/src/configuration.dart
Goodwine: Yup, I actually tried a few ways to achieve "this" earlier but a constructor initializer list can't access `this`, and if I declare it in the constructor body I must declare the originalConfiguration as `late final` which disallows `const` to be used in `Configuration.empty()`

Another alternative is renaming back to `opaqueId` and either assign `Object()` or `SomeNewClass()` value to it

nex3: Good point. I'd use a private getter backed by a field here:

```dart
  /// The configuration from which this was modified with `@forward ... with`.
  Configuration get _originalConfiguration => __originalConfiguration ?? this;

  /// The backing value for [_originalConfiguration].
  ///
  /// This is null if [_originalConfiguration] refers to this configuration,
  /// since `this` can't be assigned to a final field.
  final Configuration? __originalConfiguration;
```

lib/src/configuration.dart
```diff
@@ -65,10 +89,17 @@ class Configuration {
...
+  /// The copy will have the same [_originalConfiguration] as `this` config.
   Configuration _withValues(Map<String, ConfiguredValue> values) =>
-      Configuration.implicit(values);
+      Configuration._(values, _originalConfiguration);
 
+  /// Creates a [Configuration] with the given [_values] map and an
+  /// [_originalConfiguration] reference.
+  Configuration._(this._values, this.__originalConfiguration);
+
+  @override
```
nex3, line 102: The Sass codebase doesn't use `@override`.

nex3 (approved):

Issue #1727, opened 2022-06-22
Deprecate then disable bogus selector combinators


nex3:
See https://github.com/sass/sass/issues/3340

- [ ] Deprecate
- [ ] Disable

PR #1740, opened 2022-07-14
Deprecate bogus combinators

nex3:
See sass/sass#3340
See sass/dart-sass#1727
See https://github.com/sass/sass-spec/pull/1803

Blocked on https://github.com/sass/dart-sass/pull/1741

### DIFF
--- a/lib/src/ast/selector.dart
+++ b/lib/src/ast/selector.dart
@@ -4,12 +4,20 @@
 
 import 'package:meta/meta.dart';
 
+import '../evaluation_context.dart';
+import '../visitor/any_selector.dart';
 import '../visitor/interface/selector.dart';
 import '../visitor/serialize.dart';
+import 'selector/complex.dart';
+import 'selector/list.dart';
+import 'selector/placeholder.dart';
+import 'selector/pseudo.dart';
 
 export 'selector/attribute.dart';
 export 'selector/class.dart';
+export 'selector/combinator.dart';
 export 'selector/complex.dart';
+export 'selector/complex_component.dart';
 export 'selector/compound.dart';
 export 'selector/id.dart';
 export 'selector/list.dart';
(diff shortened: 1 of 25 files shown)


lib/src/ast/selector/combinator.dart
```diff
@@ -0,0 +1,31 @@
...
+// https://opensource.org/licenses/MIT.
+
+import 'package:meta/meta.dart';
+
+/// A combinator that defines the relationship between selectors in a
+/// [ComplexSelector].
+///
+/// {@category Selector}
+@sealed
+class Combinator {
```
jathak, line 12: Not sure if it's worth doing in this PR, but would it make sense to use an [enhanced enum](https://dart.dev/guides/language/language-tour#declaring-enhanced-enums) for this? (they were added in Dart 2.17).

lib/src/visitor/any_selector.dart
```diff
@@ -0,0 +1,39 @@
...
+/// A visitor that visits each selector in a Sass selector AST and returns
+/// `true` if any of the individual methods return `true`.
+///
+/// Each method returns `false` by default.
+@internal
+abstract class AnySelectorVisitor implements SelectorVisitor<bool> {
+  const AnySelectorVisitor();
+
+  bool visitComplexSelector(ComplexSelector complex) => complex.components
+      .any((component) => visitCompoundSelector(component.selector));
```
jathak, line 19: Is there a reason why this calls `visitCompoundSelector` directly instead of `component.selector.accept(this)`?

pubspec.yaml
```diff
@@ -49,3 +49,9 @@ dev_dependencies:
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
   yaml: ^3.1.0
+
+dependency_overrides:
+  source_span:
+    git:
+      url: https://github.com/dart-lang/source_span
```
jathak, line 56: Should this PR also be marked as blocked on dart-lang/source_span#86?

jathak (approved):

lib/src/ast/selector.dart
```diff
@@ -32,11 +41,131 @@ export 'selector/universal.dart';
...
+              component.combinators.length > 1 ||
+              component.selector.accept(this));
+    }
+  }
+
+  bool visitPseudoSelector(PseudoSelector pseudo) {
+    var selector = pseudo.selector;
+    if (selector == null) return false;
+
+    // The CSS spec specifically allows leading combinators in `:has()`.
```
stof, line 154: The CSS spec allows 1 leading combinator.

Issue #1728, opened 2022-06-22
Add support for full media query conditions


nex3:
See https://github.com/sass/sass/issues/2538

* [x] Deprecations (https://github.com/sass/dart-sass/pull/1749)
* [ ] Final (https://github.com/sass/dart-sass/pull/1822)

PR #1749, opened 2022-07-20
Add partial support for Media Queries Level 4

nex3:
See sass/sass#2538
See #1728
See sass/sass-spec#1807

### DIFF
--- a/lib/src/ast/css/media_query.dart
+++ b/lib/src/ast/css/media_query.dart
@@ -15,14 +15,24 @@ class CssMediaQuery {
 
   /// The media type, for example "screen" or "print".
   ///
-  /// This may be `null`. If so, [features] will not be empty.
+  /// This may be `null`. If so, [conditions] will not be empty.
   final String? type;
 
-  /// Feature queries, including parentheses.
-  final List<String> features;
+  /// Whether [conditions] is a conjunction or a disjunction.
+  ///
+  /// In other words, if this is `true this query matches when _all_
+  /// [conditions] are met, and if it's `false` this query matches when _any_
+  /// condition in [conditions] is met.
+  ///
+  /// If this is [false], [modifier] and [type] will both be `null`.
+  final bool conjunction;
 
-  /// Whether this media query only specifies features.
-  bool get isCondition => modifier == null && type == null;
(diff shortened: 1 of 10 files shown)


lib/src/parse/media_query.dart
```diff
@@ -29,52 +30,93 @@ class MediaQueryParser extends Parser {
...
+      // For example, "@media screen {"
+      return CssMediaQuery.type(identifier1);
+    }
 
-      if (equalsIgnoreCase(identifier2, "and")) {
-        // For example, "@media screen and ..."
-        type = identifier1;
+    var identifier2 = identifier();
+
+    if (equalsIgnoreCase(identifier2, "and")) {
```
Goodwine, line 70: Sorry, it took me a while to review because I was confused about this part, I was reading the syntax for [`<media-query>`](https://www.w3.org/TR/mediaqueries-4/#typedef-media-query) and I didn't realize it had two lines 😅 (the second showing `[ not | only ]? <media-type> [ and <media-condition-without-or> ]?`)

I didn't realize one could have `and` like that but not `or` :)

Goodwine (approved):

lib/src/ast/css/media_query.dart
```diff
@@ -36,47 +46,67 @@ class CssMediaQuery {
...
+  CssMediaQuery.condition(Iterable<String> conditions, {bool? conjunction})
       : modifier = null,
         type = null,
-        features = List.unmodifiable(features);
+        conjunction = conjunction ?? true,
+        conditions = List.unmodifiable(conditions) {
+    if (this.conditions.length > 1 && conjunction == null) {
+      throw ArgumentError(
+          "If conditions is longer than one element, conjunction may not be "
+          "null.");
```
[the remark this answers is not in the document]
nex3: It's not the typical style for ArgumentErrors, I think, and hewing to a common style is probably more valuable overall.

lib/src/parse/media_query.dart
nex3: Yeah, it has to balance remaining compatible with the legacy syntax even in places where it's confusing with supporting the new syntax.

PR #1770, opened 2022-08-03
Release 1.54.2

nex3:
This allows us to cut another release which will include
sass/dart-sass-embedded#103.


jathak (approved):

ntkme:
Looks like we might have a flake during tests: https://github.com/sass/dart-sass/runs/7662705588?check_suite_focus=true

PR #1777, opened 2022-08-15
Transform all enum-like classes to dart enums

### LINKED ISSUE #1746: Use Dart enums
Now that Dart 1.17 support [enhanced enums](https://dart.dev/guides/language/language-tour#declaring-enhanced-enums), our reason for avoiding native Dart enums are obsolete. We should go through and update our existing pseudo-enums to real ones.

Goodwine:
Fixes #1746

- Upgrades dart sdk to 2.17 in pubspecs to support enums
- Transforms all enum-like classes to dart enums
- Fixes new linter warnings around `new Foo` in comments

### DIFF
--- a/lib/src/ast/css/media_query.dart
+++ b/lib/src/ast/css/media_query.dart
@@ -204,9 +204,15 @@ class CssMediaQuery {
 /// This is either the singleton values [empty] or [unrepresentable], or an
 /// instance of [MediaQuerySuccessfulMergeResult].
 class MediaQueryMergeResult {
-  MediaQueryUnsuccessfulMergeResult? unsuccessful;
+  final MediaQueryUnsuccessfulMergeResult? unsuccessful;
 
-  MediaQueryMergeResult.unsuccessful(this.unsuccessful);
+  /// Creates a [MediaQueryMergeResult] with an unsuccessful result type.
+  const MediaQueryMergeResult.unsuccessful(
+      MediaQueryUnsuccessfulMergeResult this.unsuccessful);
+
+  /// Creates a successful [MediaQueryMergeResult] meant to be casted as
+  /// a [MediaQuerySuccessfulMergeResult].
+  const MediaQueryMergeResult._() : unsuccessful = null;
 }
 
 /// Non-successful merge results.
@@ -225,5 +231,5 @@ class MediaQuerySuccessfulMergeResult extends MediaQueryMergeResult {
   /// The merged media query.
   final CssMediaQuery query;
 
-  MediaQuerySuccessfulMergeResult._(this.query) : super.unsuccessful(null);
+  MediaQuerySuccessfulMergeResult._(this.query) : super._();
 }
--- a/lib/src/ast/css/media_query.dart
+++ b/lib/src/ast/css/media_query.dart
@@ -75,7 +75,8 @@ class CssMediaQuery {
   /// of both inputs.
   MediaQueryMergeResult merge(CssMediaQuery other) {
     if (!conjunction || !other.conjunction) {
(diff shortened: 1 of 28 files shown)


lib/src/ast/css/media_query.dart
```diff
@@ -194,31 +203,33 @@ class CssMediaQuery {
 ///
 /// This is either the singleton values [empty] or [unrepresentable], or an
 /// instance of [MediaQuerySuccessfulMergeResult].
-abstract class MediaQueryMergeResult {
-  /// A singleton value indicating that there are no contexts that match both
-  /// input queries.
-  static const empty = _SingletonCssMediaQueryMergeResult("empty");
+class MediaQueryMergeResult {
```
nex3, line 206: I'd rather keep this as-is rather than split it into a nested class if new enums can't support extra data on a subtype. Same with `ColorFormat`.

lib/src/callable/async.dart
```diff
@@ -36,7 +36,7 @@ abstract class AsyncCallable {
   /// The argument declaration is parsed from [arguments], which should not
   /// include parentheses. Throws a [SassFormatException] if parsing fails.
   ///
-  /// See [new Callable] for more details.
+  /// See [Callable] for more details.
```
nex3, line 39: This `new` is doing real work here—it indicates that we're trying to link specifically to the default constructor, *not* to the class's general documentation. Is there another recommended way to indicate that?

lib/src/ast/sass/expression/binary_operation.dart
```diff
@@ -88,51 +88,48 @@ class BinaryOperationExpression implements Expression {
 /// A binary operator constant.
 ///
 /// {@category AST}
-@sealed
-class BinaryOperator {
+enum BinaryOperator {
   /// The Microsoft equals operator, `=`.
```
nex3, line 92: Can you run Dartdoc and make sure the option documentation still shows up properly?

nex3 (changes requested):

lib/src/ast/sass/expression/binary_operation.dart
Goodwine: Ah it seems the "new way" is `[MyClass.new]` or `[MyClass.myNamedConstructorOrFactory]`

lib/src/callable/async.dart
Goodwine: Ah it seems the "new way" is [MyClass.new] or [MyClass.myNamedConstructorOrFactory]

lib/src/ast/css/media_query.dart
Goodwine: yup, those two were a bit difficult to wrangle

lib/src/ast/sass/expression/binary_operation.dart
Goodwine: browsing through these, they look correct, they are now nested under an "Enums" subsection but otherwise symbols and everything looks better actually because the "values" appear more upfront compared to static properties, although I'm not sure if you meant Operator instead of Option 🤔

lib/src/functions/math.dart
```diff
@@ -317,7 +317,7 @@ num _fuzzyRoundIfZero(num number) {
   return number.isNegative ? -0.0 : 0;
 }
 
-/// Returns a [Callable] named [name] that transforms a number's value
+/// Returns a [Callable.new] named [name] that transforms a number's value
```
nex3, line 320: This one should probably just be `[Callable]`, since it's just being used to describe the return type.

nex3 (approved):

Issue #1756, opened 2022-07-22
Show deprecation warning when using math.random() with units


Goodwine:
See https://github.com/sass/sass/issues/1890

- [ ] Deprecation
- [ ] Final

Goodwine:
Currently blocked on pending feedback

Goodwine:
deprecation added, now waiting on dart 2.0.0

PR #1779, opened 2022-08-16
Deprecate math.random() when $limit has units

Goodwine:
https://github.com/sass/dart-sass/issues/1756

See https://github.com/sass/sass-spec/pull/1815

### DIFF
--- a/lib/src/functions/math.dart
+++ b/lib/src/functions/math.dart
@@ -297,10 +297,10 @@ final _randomFunction = _function("random", r"$limit: null", (arguments) {
       "future release.\n"
       "\n"
       "Recommendation: "
-      "math.random(\$limit / 1${limit.unitString}) * 1${limit.unitString}\n"
+      "math.random(math.div(\$limit, 1${limit.unitString})) * 1${limit.unitString}\n"
       "\n"
       "To preserve current behavior: "
-      "math.random(\$limit / 1${limit.unitString})",
+      "math.random(math.div(\$limit, 1${limit.unitString}))",
     );
   }
 
--- a/lib/src/functions/math.dart
+++ b/lib/src/functions/math.dart
@@ -289,11 +289,23 @@ final _random = math.Random();
 
 final _randomFunction = _function("random", r"$limit: null", (arguments) {
   if (arguments[0] == sassNull) return SassNumber(_random.nextDouble());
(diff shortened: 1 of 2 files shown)


lib/src/functions/math.dart
```diff
@@ -289,11 +289,23 @@ final _random = math.Random();
...
-  if (limit < 1) {
+  var limit = arguments[0].assertNumber("limit");
+
+  if (limit.hasUnits) {
+    warn(
+      "math.random() will no longer ignore \$limit units ($limit) in a "
+      "future release.\n"
+      "\n"
+      "If you meant to preserve units: "
+      "math.random($limit) * 1${limit.unitString}",
```
nex3, line 300: Deprecation suggestions should generally be written so that they'll silence the warning, provide the desired behavior, *and* work after the breaking change. In this case, this will still produce a warning (since they're passing in the same limit) and it'll break once the unit is preserved (because the multiplication will add an extra unit).

A better suggestion might be: `math.random(\$limit / 1${limit.unitString}) * 1${limit.unitString})`.

nex3, line 300: It's generally a good idea to use parameter names rather than values when providing suggestions for built-in functions, since users are probably not just passing in literal values to begin with. (So `\$limit` rather than `$limit`.)

lib/src/functions/math.dart
```diff
@@ -289,11 +289,23 @@ final _random = math.Random();
...
-  var limit = arguments[0].assertNumber("limit").assertInt("limit");
-  if (limit < 1) {
+  var limit = arguments[0].assertNumber("limit");
+
+  if (limit.hasUnits) {
+    warn(
+      "math.random() will no longer ignore \$limit units ($limit) in a "
+      "future release.\n"
+      "\n"
+      "If you meant to preserve units: "
```
nex3, line 299: It's probably also worth adding a suggestion for preserving the existing behavior even after teh deprecation.

nex3 (changes requested):

lib/src/functions/math.dart
Goodwine: yeah I was thinking something similar when looking at the tests 😅

nex3 (changes requested):

nex3 (approved):

PR #1784, opened 2022-08-19
Convert visitor classes to mixins

nex3:


pkg/sass_api/CHANGELOG.md
```diff
@@ -1,7 +1,23 @@
 ## 3.0.0
 
-* Replace the `minSpecificity` and `maxSpecificity` fields on `ComplexSelector`,
-  `CompoundSelector`, and `SimpleSelector` with a single `specificity` field.
+* **Breaking change:** Convert all visitor superclasses into mixins. This
+  includes `RecursiveAstVisitor`, `RecursiveSelectorVisitor`,
+  `RecursiveStatementVisitor`, and `StatementSearchVisitor`. This has several
+  effects;
+
+  * You must use `on` to mix in visitors rather than `extends`.
```
jathak, line 8:
```suggestion
  * You must use `with` to mix in visitors rather than `extends`.
```

jathak (approved):
Is there something blocking a release right now? I'll need to update the migrator in order to bring this in internally, but if we're waiting on something externally, I could always just do an internal patch for the time being.

PR #1795, opened 2022-09-02
Upgrade dev dependencies to the latest version

stof:


stof:
Changes done in 2nd to 6th commits are about fixing errors reporting by the new analyzer rules.

nex3:
Can we disable the string composition lint? I think that's pretty universally making the code look worse.

stof:
@nex3 updated with the lint rule disabled. I also updated the `sass-analysis` dependency to use lints 2.0 for consistency.

nex3:
What happened to the two disables you removed from the analysis config?

stof:
@nex3 as explained in the commit message removing them, those rules were part of the pedantic config but are not part of the `lints` recommended config. So they were ignoring rule that were not enabled in the first place.

nex3 (approved):

nex3:
Thanks!

PR #1801, opened 2022-09-14
Fix bug in JS MultiSpan

### LINKED ISSUE #1790: MultiSpan does not fullfill documented API interface
The `MultiSpan` introduced in Sass 1.54.0 does not provide access to the properties documented for [`SourceSpan`](https://sass-lang.com/documentation/js-api/interfaces/SourceSpan). Although it [looks like](https://github.com/sass/dart-sass/blob/1.54.5/lib/src/util/multi_span.dart#L31-L34) the properties should be accessible (I can only take a guess here because I don't know Dart), they are `undefined` in the JavaScript API.

Minimal Node.js code for reproducing the problem (tested with latest Sass version, 1.54.5):

```js
let sass = require('sass')

sass.compileString(`* > { --foo: bar }`, {
  logger: {
    warn(_message, { span }) {
      // `span.start` should yield a `SourceLocation` according to documentation, gets `undefined` instead
      console.log(span.start)
    }
  }
})
```

jathak:
Fixes #1790.

See https://github.com/sass/sass-spec/pull/1825

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.55.1
+
+### JS API
+
+* Fix a bug in which certain warning spans would not have their properties
+  accessible by the JS API.
+
 ## 1.55.0
 
 ### Dart API
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,14 @@
-## 1.55.1
-
-### JS API
-
-* Fix a bug in which certain warning spans would not have their properties
-  accessible by the JS API.
-
 ## 1.55.0
 
 ### Dart API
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.55.1
```
nex3, line 1: 1.55.0 hasn't been released yet

nex3 (approved):

CHANGELOG.md
jathak: Should I release this as 1.55.0 then, or should I make it a -dev version while we wait for more changes?

Issue #1550, opened 2021-11-08
Improve error message when a custom importer returns a Buffer instead of a string


Genuifx:
when using `@import ` to import customer CSS subset file such as `@import "a.wxss"`,  the dart-sass compiler always throws the error `Error: Can't find stylesheet to import.`, even I manually read file contents and pass to done Functions.

Version:
- sass@1.43.4
- node@12.18.1

**reproduce code:**

`a.scss`

```
@import "./index.skeleton.wxss";
```

`index.skeleton.wxss`  CSS-subset file.

```
// some CSS compat rules
```

process scripts: 

```js
import sass from 'sass';
import path from 'path';
import fs from 'fs';
import Fiber from 'fibers';

...some code 

let ret = sass.renderSync({
    ...configs,
    data,
    file: filepath,
    filber: Fiber,
    // includePaths: [filepath],
    importer(url, prev, done) {
        // debugger;
        let file = path.resolve(path.dirname(prev), url);
        let ext = path.extname('url')

        if (ext === '.wxss') {
            done({
                contents: fs.readFileSync(file)
            })
        } 

        return file;
    }
})

```

PapaNappa:
Your code says `path.extname('url')` instead of `path.extname(url)`. `url` is probably meant to be passed as the variable, not the string.

Genuifx:
using `node-fibers` is work.
but `node-fibers` is not compat with node 16.

> Update [April 13th, 2021] -- Fibers is not compatible with nodejs v16.0.0 or later. Unfortunately, v8 commit dacc2fee0f is a breaking change and workarounds are non-trivial.

without enabling fiber option. importer seems not work
```
sass.renderSync({
      ...configs,
      importer(url, prev) {
          let file = path.resolve(path.dirname(prev), url);
          let ext = path.extname(url)

          if (ext === '.wxss') {
              console.log('wxss\n', file)

              return {
                  contents: fs.readFileSync(file)
              }
          } 
          return {file};
      }
  })
```

and get Error: `Error: NoSuchMethodError: method not found: 'charCodeAt' (receiver.charCodeAt is not a function)`

@PapaNappa

jathak:
It looks like the underlying issue is that the `contents` expects a string, but `fs.readFileSync(file)` returns a `Buffer`. You need to either pass an encoding to `readFileSync` (e.g. `fs.readFileSync(file, 'utf8')`) or call `toString()` on the result.

We could probably improve the error message though.

[...]

PR #1816, opened 2022-10-28
JS API: Validate that importer result 'contents' is a `string` and improve ArgumentError output

Goodwine:
See: https://github.com/sass/sass-spec/pull/1834

Example output:
```
sass.Exception [Error]: Invalid argument (contents): must be a string but was: Buffer.: Instance of 'NativeUint8List'
  ╷
1 │ @import './foo.wxss';
  │         ^^^^^^^^^^^^
  ╵
  - 1:9  root stylesheet
  - [...]
```

Before using `toString()` for Argument error:
```
sass.Exception [Error]: must be a string but was: Buffer
  ╷
1 │ @import './foo.wxss';
  │         ^^^^^^^^^^^^
  ╵
  - 1:9  root stylesheet
  - [...]
```
Original error:
```
sass.Exception [Error]: NoSuchMethodError: method not found: 'charCodeAt' (receiver.charCodeAt is not a function)
  ╷
1 │ @import './foo.wxss';
  │         ^^^^^^^^^^^^
  ╵
  - 1:9  root stylesheet
  - [...]
```

Fixes https://github.com/sass/dart-sass/issues/1550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@
 ### JS API
 
 * Importer results now validate whether `contents` is actually a string type.
+
 * Importer result argument errors are now rendered correctly.
 
 ## 1.55.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,11 @@
 * Emit a deprecation warning when passing a `sassIndex` with units to
   `Value.sassIndexToListIndex()`. This will be an error in Dart Sass 2.0.0.
 
+### JS API
+
+* Importer results now validate whether `contents` is actually a string type.
+* Importer result argument errors are now rendered correctly.
+
 ## 1.55.0
 
 * **Potentially breaking bug fix:** Sass numbers are now universally stored as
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,11 @@
 * Emit a deprecation warning when passing a `sassIndex` with units to
   `Value.sassIndexToListIndex()`. This will be an error in Dart Sass 2.0.0.
 
+### JS API
+
+* Importer results now validate whether `contents` is actually a string type.
+* Importer result argument errors are now rendered correctly.
```
nex3, line 34: Nit: We usually include a blank line between bullet points to help with plain-text readability.

nex3 (changes requested):
Can you add a test for this to `sass-spec` as well?

Goodwine:
> Can you add a test for this to sass-spec as well?

I wasn't sure where were the tests actually, but I found them now 🙂

Added tests and updated to also handle legacy importers which I realized just now via testing that it wasn't handled (yay tests)

nex3 (approved):

Issue #1776, opened 2022-08-13
Implement strict function units


nex3:
See https://github.com/sass/sass/issues/3374

- [ ] Deprecation
- [ ] Removal

PR #1817, opened 2022-10-28
Implement deprecations for strict function units

nex3:
See #1776
See sass/sass#3374
See https://github.com/sass/sass-spec/pull/1830


lib/src/functions/color.dart
```diff
@@ -656,12 +666,13 @@ void _checkAngle(SassNumber angle, String name) {
 
     var actualUnit = angle.numeratorUnits.first;
     message
-      ..writeln("To preserve current behavior: \$$name * 1deg/1$actualUnit")
+      ..writeln(
+          "To preserve current behavior: calc(\$$name * " "1deg/1$actualUnit)")
```
Goodwine, line 670: Nit: combine both strings into one

Goodwine (approved):

Issue #1363, opened 2021-06-15
Add a --future-deprecations flag


jathak:
This would be somewhat similar to #633, except instead of opting-in to early errors for specific active deprecations, it would opt-in to early warnings for all known future deprecations.

- [x] CLI support
- [x] Dart API support
- [x] JS API support
- [x] Documentation

PR #1820, opened 2022-10-28
Add --fatal-deprecations and --future-deprecations

jathak:
Partial fix for #633 and #1363.

This adds a new `Deprecation` enum to track the different deprecations that are ongoing or upcoming and allow users to select certain deprecations to treat as fatal (or future deprecations to begin warning for now).

This adds support to the CLI and the Dart API. It does not add support for the JS API yet, but I'd like to land this first so the underlying logic isn't blocked on the language change process.


bin/sass.dart
```diff
@@ -52,8 +53,15 @@ Future<void> main(List<String> args) async {
...
-        ImportCache(loadPaths: options.loadPaths, logger: options.logger));
+    var graph = StylesheetGraph(ImportCache(
+        loadPaths: options.loadPaths,
+        logger: DeprecationHandlingLogger(options.logger,
+            fatalDeprecations: options.fatalDeprecations,
+            futureDeprecations: options.futureDeprecations,
+            // This logger may be reused by multiple parses, so we don't want
+            // warnings from a previous parse to cause subsequent warnings to
+            // be skipped.
+            limitRepetition: false)));
```
nex3, line 64: I have a couple questions here:

* Previously, the `Options` class was responsible for setting up the logger according to the options passed. Why not do the same with `DeprecationHandlingLogger`?

* If we don't limit repetitions here, under what circumstances _will_ we limit repetitions on the CLI?

lib/sass.dart
```diff
@@ -105,7 +107,9 @@ CompileResult compileToResult(String path,
         bool quietDeps = false,
         bool verbose = false,
         bool sourceMap = false,
-        bool charset = true}) =>
+        bool charset = true,
+        Set<Deprecation> fatalDeprecations = const {},
+        Set<Deprecation> futureDeprecations = const {}}) =>
```
nex3, line 112: Non-boolean parameters should generally have `null` defaults, so that we can change default behavior later without changing the public signature.

lib/src/deprecation.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright 2022 Google LLC. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:collection/collection.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+/// Represents a deprecated feature in the language.
```
nex3, line 8: Nit: Remove "Represents", since an enum is always a representation of something

lib/src/deprecation.dart
```diff
@@ -0,0 +1,98 @@
...
+      deprecatedIn: '1.56.0',
+      description: 'Passing invalid units to built-in functions.'),
+
+  /// Deprecation for `@import` rules.
+  import('import', deprecatedIn: null, description: '@import rules.'),
+
+  /// Used for deprecations of an unknown type.
+  ///
+  /// We set its deprecated-in version to 1000.0.0 so that it won't be made
+  /// fatal by passing a Sass version to --fatal-deprecation.
```
nex3, line 60: I think it would be clearer to set this to `null` and have an explicit boolean field for `futureDeprecation`.

lib/src/deprecation.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  /// Deprecation for passing invalid units to certain built-in functions.
+  functionUnits('function-units',
+      deprecatedIn: '1.56.0',
+      description: 'Passing invalid units to built-in functions.'),
+
+  /// Deprecation for `@import` rules.
+  import('import', deprecatedIn: null, description: '@import rules.'),
+
+  /// Used for deprecations of an unknown type.
```
nex3, line 57: If this is only going to be used by deprecations from user-authored functions in practice, it might make more sense to name and document it for that explicitly.

lib/src/deprecation.dart
```diff
@@ -0,0 +1,98 @@
...
+  ///
+  /// This is used to refer to the deprecation on the command line.
+  final String id;
+
+  /// The Dart Sass version this feature was first deprecated in.
+  ///
+  /// For deprecations that have existed in all versions of Dart Sass, this
+  /// should be 0.0.0. For deprecations that are not yet active, this should be
+  /// null.
+  final String? deprecatedIn;
```
nex3, line 73: Maybe make this a `pub_semver` `Version` object?

lib/src/executable/options.dart
```diff
@@ -78,6 +79,34 @@ class ExecutableOptions {
...
+      ..addFlag('quiet', abbr: 'q', help: "Don't print warnings.")
+      ..addFlag('quiet-deps',
+          help: "Don't print compiler warnings from dependencies.\n"
+              "Stylesheets imported through load paths count as dependencies.")
+      ..addFlag('verbose',
+          help: "Print all deprecation warnings even when they're repetitive.")
+      ..addMultiOption('fatal-deprecation',
+          help: 'Deprecations to treat as errors. You may also pass a Sass\n'
+              'version to include any behavior deprecated in or before it.\n'
+              'See https://sass-lang.com/documentation/breaking-changes for \n'
```
nex3, line 93: We should update this documentation so that each breaking change includes an example of how to make it fatal as well

lib/src/executable/options.dart
```diff
@@ -78,6 +79,34 @@ class ExecutableOptions {
...
+              'a complete list.',
+          allowedHelp: {
+            for (var deprecation in Deprecation.values)
+              if (deprecation.deprecatedIn != null &&
+                  deprecation.description != null)
+                deprecation.id: deprecation.description!,
+          })
+      ..addMultiOption('future-deprecation',
+          help: 'Opt-in to a deprecation early.',
```
nex3, line 102: Nit: "Opt in"

"Opt-in" with a hyphen is an adjective; the verb form always uses a space.

lib/src/logger.dart
```diff
@@ -28,12 +30,32 @@ abstract class Logger {
...
+/// making a breaking API change.
+///
+/// {@category Compile}
+extension WarnForDeprecation on Logger {
+  /// Emits a deprecation warning for [deprecation] with the given [message].
+  void warnForDeprecation(Deprecation deprecation, String message,
+      {FileSpan? span, Trace? trace}) {
+    var self = this;
+    if (self is DeprecationHandlingLogger) {
+      self.handleDeprecationWarning(deprecation, message, span, trace);
```
nex3, line 52: It might be clearer to call the method on `DeprecationHandlingLogger` `warnForDeprecation()`. The call won't be ambiguous here because the static type of `self` in this branch is `DeprecationHandlingLogger`.

lib/src/logger/deprecation_handling.dart
```diff
@@ -0,0 +1,103 @@
...
+  /// If it's a future deprecation that hasn't been opted into or its a
+  /// deprecation that's already been warned for [_maxReptitions] times and
+  /// [limitRepetitions] is true, the warning is dropped.
+  ///
+  /// Otherwise, this is passed on to [warn].
+  void handleDeprecationWarning(
+      Deprecation deprecation, String message, FileSpan? span, Trace? trace) {
+    if (fatalDeprecations.contains(deprecation)) {
+      message += '\n\nThis is only an error because of '
+          '--fatal-deprecation=$deprecation.\n'
```
nex3, line 62: The CLI flag format may not be accurate for library users.

lib/src/deprecation.dart
```diff
@@ -0,0 +1,98 @@
...
+
+/// Represents a deprecated feature in the language.
+enum Deprecation {
+  /// Deprecation for passing a string to `call` instead of `get-function`.
+  callString('call-string',
+      deprecatedIn: '0.0.0',
+      description: 'Passing a string directly to meta.call().'),
+
+  /// Deprecation for `@elseif`.
+  elseIf('elseif', deprecatedIn: '1.3.2', description: '@elseif.'),
```
nex3, line 16: Nit: I think the `i` should be lowercase here to match the at-rule.

nex3 (changes requested):
Can you also file issues to track adding this to the JS API and documenting it on the Sass website?

bin/sass.dart
jathak: This wrapped logger is only used during parsing. `options.logger` is wrapped by a `DeprecationHandlingLogger` again during the actual compilation (this time, with `limitRepetition` set to true if `verbose` is false).

This matches the current behavior with the `TerseLogger` that `DeprecationHandlingLogger` replaces (repetitions aren't limited for deprecation warnings during parsing). The need to reset the state of `DeprecationHandlingLogger` is also why we don't just have `options.logger` produce it.

We could change `options.logger` to just produce a single `DeprecationHandlingLogger`, but we'd then need some way for `StylesheetGraph` and the compiler to reset its state after each run. Do you think that makes more sense instead of wrapping the base logger multiple times?

lib/src/deprecation.dart
jathak: Done

lib/src/deprecation.dart
jathak: Done

lib/src/deprecation.dart
jathak: Done

CHANGELOG.md
```diff
@@ -1,3 +1,34 @@
...
+  Sass `@import` rules, which are not yet deprecated, but will be in the future.
+
+### Dart API
+
+* New `Deprecation` enum, which contains the different current and future
+  deprecations used by the new CLI flags.
+
+* The optional `deprecation` boolean parameter of the `warn` function is now
+  deprecated. Use `deprecationType` to pass the specific `Deprecation` being
+  warned for instead.
```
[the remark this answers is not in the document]
jathak: I was thinking they would pass `Deprecation.userAuthored`, but do you think it would make more sense to just leave the boolean parameter un-deprecated? (since right now if they pass `deprecation: true` with no `deprecationType`, it defaults to `Deprecation.userAuthored`)

Or could it be confusing to have both `deprecation` and `deprecationType` un-deprecated, since then you could potentially call it with a deprecation type but explicitly pass `deprecation: false`? (in which case the boolean parameter would be ignored)

Would making `deprecation` nullable but un-deprecated work?

If we do leave the `deprecation` un-deprecated in `EvaluationContext.warn`, should we also do the same for `Logger.warn`? If we do that, should we just leave the `warnForDeprecation` extension for `Logger` as private? (since custom functions should usually be just using `Deprecation.userAuthored` anyway)

lib/sass.dart
jathak: Done

lib/src/deprecation.dart
jathak: Done

lib/src/executable/options.dart
jathak: Done

lib/src/logger.dart
jathak: Done

lib/src/logger/deprecation_handling.dart
jathak: Updated this to be more generic. (Or would it make more sense to try and detect CLI vs API use and update the message accordingly?)

lib/src/executable/options.dart
jathak: Working on a sass-site PR now

lib/src/deprecation.dart
nex3: Oh, I see... real enums strike again. I think the cleanest option given that might be to convert this to a normal class with `static final` members instead of a literal enum. That's more work to declare, but nicer on the consuming end, which I think it a worthy trade-off.

CHANGELOG.md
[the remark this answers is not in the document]
nex3: What if we just have two separate functions for both `EvaluationContext` _and_ `Logger`, one for public consumption that takes a boolean `deprecation` flag and one that's private to Sass which takes a `Deprecation`?

bin/sass.dart
nex3: No, I think this is fine. It might be worth documenting more of what you just said in the comments, though.

lib/src/logger/deprecation_handling.dart
nex3: I think making it generic is fine.

lib/src/deprecation.dart
```diff
@@ -68,16 +66,23 @@ enum Deprecation {
...
   /// A description of this deprecation that will be displayed in the CLI usage.
   ///
   /// If this is null, the given deprecation will not be listed.
   final String? description;
 
-  const Deprecation(this.id, {required this.deprecatedIn, this.description});
+  /// Whether this deprecation will occur in the future.
+  ///
+  /// If this is true, `deprecatedIn` should be null, since we do not yet know
+  /// what version of Dart Sass this deprecation will be live in.
```
nex3, line 81: If you make this a real class, you can also verify this in a constructor `assert()`.

lib/src/deprecation.dart
jathak: Added a getter for the parsed Version and made the underlying string private, as discussed.

lib/src/deprecation.dart
jathak: Accomplished this in the enum form by having separate constructors for regular and future deprecations.

CHANGELOG.md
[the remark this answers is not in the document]
jathak: Done

lib/src/executable/options.dart
jathak: https://github.com/sass/sass-site/pull/702

nex3 (approved):

PR #1829, opened 2022-11-17
Add a factory method for creating host callable

ntkme:
Expose a new method so that we can have consistent function signature parsing between dart-sass and dart-sass-embedded.

https://github.com/sass/sass-spec/pull/1847


lib/src/callable.dart
```diff
@@ -117,4 +120,14 @@ abstract class Callable extends AsyncCallable {
   factory Callable.function(String name, String arguments,
           Value callback(List<Value> arguments)) =>
       BuiltInCallable.function(name, arguments, callback);
+
+  /// Creates a host callable with a single [signature] and a single [callback].
```
nex3, line 124: "host callable" isn't really a meaningful term in the context of the Dart Sass API. I think it would make more sense to name/document this after its function, like `fromSignature()`.

lib/src/node/compile.dart
```diff
@@ -236,42 +230,35 @@ List<AsyncCallable> _parseFunctions(Object? functions, {bool asynch = false}) {
...
-    } on SassFormatException catch (error, stackTrace) {
-      throwWithTrace(
-          SassFormatException(
-              'Invalid signature "$signature": ${error.message}', error.span),
-          stackTrace);
-    }
-
     if (!asynch) {
-      result.add(BuiltInCallable.parsed(tuple.item1, tuple.item2, (arguments) {
+      Callable? callable;
```
nex3, line 234:
```suggestion
      late Callable callable;
```

This will allow you to declare the value lazily without needing explicit null assertions (instead, the language bakes in "this was assigned" assertions at each access point).

lib/src/util/function_signature.dart
```diff
@@ -0,0 +1,28 @@
...
+import '../parse/scss.dart';
+import '../utils.dart';
+
+/// Parses a function signature of the format allowed by Node Sass's functions
+/// option and returns its name and declaration.
+///
+/// If [requireParens] is `false`, this allows parentheses to be omitted.
+///
+/// Throws a [SassFormatException] if parsing fails.
+Tuple2<String, ArgumentDeclaration> parseSignature(String signature,
```
nex3, line 18: Nit: just include this in `lib/src/util.dart`

nex3 (changes requested):

lib/src/util/function_signature.dart
ntkme: In the current structure `lib/src/utils.dart` only have high level helper functions, and all feature specific helpers are in their own files in `lib/src/util/`. - This is why I created a separate file. @nex3 let me know if you still prefer to put this function in `lib/src/utils.dart`.

ntkme: @nex3 Any thoughts?

nex3: Yeah, generally I prefer to only add sub-libraries for *groups* of utility functions. If there's only one function of a given type, it should go in the general utility library.

nex3:
This also seems to be blocked on https://github.com/dart-lang/pub/issues/3687.

PR #1848, opened 2022-12-14
Be strict about whitespace in custom functions

nex3:
See sass/sass-spec#1856

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@
 
 ### Dart API
 
-* **Potentially breaking bug fix**: `Callable.parseSignature()` and
-  `AsyncCallable.parseSignature()` now reject signatures with whitespace between
+* **Potentially breaking bug fix**: `Callable.fromSignature()` and
+  `AsyncCallable.fromSignature()` now reject signatures with whitespace between
   the function name and parentheses.
 
 ## 1.56.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+## 1.56.3
+
+### JavaScript API
+
+* **Potentially breaking bug fix**: Custom functions in both the modern and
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -9,5 +9,8 @@
     "chokidar": ">=3.0.0 <4.0.0",
     "immutable": "^4.0.0",
     "intercept-stdout": "^0.1.2"
+  },
+  "dependencies": {
+    "node-sass": "^8.0.0"
```
ntkme, line 14: I guess this was accidentally added for some local testing?

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+* **Potentially breaking bug fix**: Custom functions in both the modern and
+  legacy API now properly reject signatures with whitespace between the function
+  name and parentheses.
+
+* Custom functions in the legacy API now allow signatures with whitespace before
+  the function name, to match a bug in Node Sass.
+
+### Dart API
+
+* **Potentially breaking bug fix**: `Callable.parseSignature()` and
```
ntkme, line 14: I think this is `fromSignature()`, and same for next line.

lib/src/node/legacy.dart
```diff
@@ -210,7 +210,7 @@ List<AsyncCallable> _parseFunctions(RenderOptions options, DateTime start,
 
     var fiber = options.fiber;
     if (fiber != null) {
-      result.add(Callable.fromSignature(signature, (arguments) {
+      result.add(Callable.fromSignature(signature.trimLeft(), (arguments) {
```
ntkme, line 213: There are two more places below in legacy API, do we need to match all three calls? Not sure what is the scope of node-sass bug we’re trying to match.

ntkme (approved):

jathak (approved):

PR #1853, opened 2022-12-17
Fix dart-sass-embedded release process

ntkme:
Top level pub command is no longer available and it’s causing 1.57.0 release to fail.

@nex3 We need to either manually cut 1.57.0 for dart-sass-embedded or cut 1.57.1 instead.


ntkme:
Just saw #1851. Closing.

PR #1860, opened 2022-12-27
Remove sourcemap comments from Sass sources

Goodwine:
See https://github.com/sass/sass-spec/pull/1862

Closes https://github.com/sass/sass/issues/3448

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: sass
-version: 1.57.2-dev
+version: 1.57.2
 description: A Sass implementation in Dart.
 homepage: https://github.com/sass/dart-sass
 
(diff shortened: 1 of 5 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.57.1
+version: 1.57.2-dev
```
nex3, line 2: Any reason not to just cut a release?

nex3 (approved):

pubspec.yaml
Goodwine: letting other changes come in and release several at once, but I guess that's only needed during a fixit week :)

PR #1868, opened 2023-01-11
Add all CSS length units in known compatibilities

stof:
This allows more validation of wrong computations in first-class calc.

This adds the missing font-relative and viewport-relative units from https://drafts.csswg.org/css-values-4/#font-relative-lengths

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.57.2
+## 1.58.0
 
 * Remove sourcemap comments from Sass sources. The generated sourcemap comment
   for the compiled CSS output remains unaffected.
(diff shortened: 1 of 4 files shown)


nex3:
Which of these are actually implemented in a real browser? That's our metric for whether to add support for a new feature in the spec.

stof:
Based on https://developer.mozilla.org/en-US/docs/Web/CSS/length#browser_compatibility, only `rlh` is not yet implemented (but based on https://bugs.webkit.org/show_bug.cgi?id=246522, it is implemented in Safari TP so will probably be there in the next Safari release). And `cap` is implemented only in Firefox for now.
And looking at that list, I forgot the container query units (which are defined in a different CSS module but have cross-browser support)

stof:
And https://bugs.chromium.org/p/chromium/issues/detail?id=1381037 marks `rlh` (and a few other `r*` units) as implemented in Chromium 2 days ago.

nex3:
Sounds good. Can you add a changelog entry as well?

stof:
@nex3 done.

nex3 (approved):

PR #1880, opened 2023-02-02
Update synchronization tests to verify the entire file

nex3:
This catches cases where the target file is edited directly without
also editing the source file.

Also re-run `pub run grinder`.

### DIFF
--- a/test/double_check_test.dart
+++ b/test/double_check_test.dart
@@ -5,7 +5,6 @@
 import 'dart:io';
 import 'dart:convert';
 
-import 'package:crypto/crypto.dart';
 import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:pubspec_parse/pubspec_parse.dart';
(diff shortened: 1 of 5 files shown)


test/double_check_test.dart
```diff
@@ -21,16 +21,11 @@ void main() {
...
-        var match = checksumPattern.firstMatch(target);
-        if (match == null) fail(message);
-
-        var source = File(sourcePath).readAsBytesSync();
-        var expectedHash = sha1.convert(source).toString();
-        expect(match[1], equals(expectedHash), reason: message);
+        if (File(targetPath).readAsStringSync() !=
+            synchronize.synchronizeFile(sourcePath)) {
+          fail("$targetPath is out-of-date.\n"
+              "Run `pub run grinder` to update it.");
```
jathak, line 27: The standalone `pub` command no longer exists. This should be `dart run grinder`.

jathak (approved):

PR #1895, opened 2023-02-24
Add support for running in the browser

jerivas:
WIP, but this gets us closer to fixing #25 

- [x] Fix `self`/`window` issues (https://github.com/mbullington/node_preamble.dart/pull/29)
- [x] Generate ESM bundles (https://github.com/google/dart_cli_pkg/pull/118)
- [x] More resilient access to `process.*`
- [x] Throw if trying to access the file system
- [x] Let users compile stylesheets without importers, but improve error message if trying to import a stylesheet without explicit importers
- [x] Throw errors if trying to use `compile()`, `compileAsync()`, `render()`, and `renderSync()` in the browser
- [x] Make sure logging still works in browser contexts
- [x] Add basic tests that run in the browser
- [x] Run JS API tests in the browser. Add browser tests for the JS API in the CI for both this and the `sass/sass-spec` repo. See sass/sass-spec#1902


lib/src/io/node.dart
```diff
@@ -8,14 +8,18 @@ import 'dart:js_util';
...
 import 'package:path/path.dart' as p;
 import 'package:source_span/source_span.dart';
 import 'package:watcher/watcher.dart';
 
 import '../exception.dart';
 import '../node/chokidar.dart';
+import '../logger.dart';
+
+@JS('process')
+external final process;
```
jerivas, line 22: I brought the entire `process` instead of just `process.stdout.isTTY`. Not sure if that's the best approach, but all errors related to `process.env` and `isTTY` are now fixed with the changes in this file

pubspec.yaml
```diff
@@ -49,3 +49,9 @@ dev_dependencies:
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
   yaml: ^3.1.0
+
+dependency_overrides:
+  node_preamble:
+    git:
+      url: https://github.com/mbullington/node_preamble.dart
+      ref: ec11f5a
```
jerivas, line 57: Fixes illegal invocations of `window` after being wrapped in `Object.create()`

jerivas:
@nex3 I'm unable to prevent `sass.compile()` from trying to access the file system, even when I have passed explicit importers. It boils down to this call of `compile` where both branches of the conditional call `FileSystemImporter` or `readFile` directly: https://github.com/sass/dart-sass/blob/754f30c89d8bf994c91664c0d26b516e2cbbdab4/lib/src/compile.dart#L56-L66

I didn't want to go into `FileSystemImporter` and start introducing checks related to node since that's a pure Dart function, so I think we should use the `load` from the import cache here. WDYT?

nex3:
I may have mispoken on the call earlier—`compile{,ToResult}{,Async}` are all fundamentally not going to work on the web. They take a filesystem path as a parameter, not a URL, and there's no real way to make that make sense in a web context. They should all just throw errors indicating that they're not supported and users should use `compileStringToResult{,Async}` instead.

tool/grind.dart
```diff
@@ -44,6 +44,9 @@ void main(List<String> args) {
   pkg.standaloneName.value = "dart-sass";
   pkg.githubUser.fn = () => Platform.environment["GH_USER"];
   pkg.githubPassword.fn = () => Platform.environment["GH_TOKEN"];
+  // Limit the methods exposed as ESM exports to those that don't require
+  // filesystem access
+  pkg.jsEsmExports.value = {'compileString', 'compileStringAsync'};
```
jerivas, line 49: @nex3 with this configuration, `grinder pkg-npm-dev` produces this code in `sass.default.js`:

```javascript
import * as util from "util"
import * as immutable from "immutable"
import * as fs from "fs"
import * as stream from "stream"
import * as _cliPkgModule from "./sass.dart.js";

// Bundlers may try to resolve this file eagerly, in which case
// sass.dart.js won't be able to accurately detect that it's being imported
// rather than required. Fortunately, in that case, exports will be defiend and
// we can load the library that way instead.
const _cliPkgLibrary =
    'load' in _cliPkgModule ? _cliPkgModule : window._cliPkgExports;
delete window._cliPkgExports;
const _cliPkgExports = {};
_cliPkgLibrary.load({util, immutable, fs, stream}, _cliPkgExports);

export const compileString = _cliPkgExports.compileString;
export const compileStringAsync = _cliPkgExports.compileStringAsync;
```

That _almost_ works. I just had to change from `import * as _cliPkgModule` to `import _cliPkgModule`. Otherwise accessing the `load` function is not possible.

jerivas: On Slack, Natalie explained this might be related to the bundler I'm using to test. I tried without bundlers, and this works as expected.

lib/src/io/node.dart
```diff
@@ -76,15 +79,15 @@ Future<String> readStdin() async {
   });
   // Node defaults all buffers to 'utf8'.
   var sink = utf8.decoder.startChunkedConversion(innerSink);
-  process.stdin.on('data', allowInterop(([Object? chunk]) {
+  process?.stdin.on('data', allowInterop(([Object? chunk]) {
```
nex3, line 82: Currently, if this is called in a context without access to `process`, it'll return a future that silently never completes which could make debugging confusing. It's probably better to check whether `process` is available eagerly and throw an error if it isn't. For example:

```dart
// Reassigning to a variable is necessary for null analysis to work for now.
var process_ = process;
if (process_ == null) {
  throw UnsupportedError("readStdin() is only supported on Node.js");
}

// ...
```

We should probably do something similar for all the Node-specific methods here. That way, if we end up accidentally invoking them from browser-available contexts later on down the line, we'll get helpful error messages.

nex3 (changes requested):

lib/src/io/node.dart
jerivas: Fixed in 2dcf8cdebf48242408025fc9297e68b8c41f7deb, and also added checks to all functions

lib/src/io/node.dart
```diff
@@ -187,30 +191,28 @@ T _systemErrorToFileSystemException<T>(T callback()) {
...
+final stderr = Stderr(process?.stderr ?? createWritable(WritableOptions()));
 
 /// We can't use [process.stdout.isTTY] from `node_interop` because of
 /// pulyaevskiy/node-interop#93: it declares `isTTY` as always non-nullably
 /// available, but in practice it's undefined if stdout isn't a TTY.
-@JS('process.stdout.isTTY')
-external bool? get isTTY;
-
-bool get hasTerminal => isTTY == true;
+// ignore: invalid_null_aware_operator
```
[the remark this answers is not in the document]
jerivas: This is to work around the issue described in the comment above this line. Without the `ignore`, Dart Analyzer complains about `stdout?.isTTY` because it assumes `isTTY` is always available

lib/src/io/node.dart
```diff
@@ -187,30 +191,28 @@ T _systemErrorToFileSystemException<T>(T callback()) {
...
+bool get hasTerminal => process?.stdout?.isTTY == true;
 
-bool get isWindows => process.platform == 'win32';
+bool get isWindows => process?.platform == 'win32';
 
-bool get isMacOS => process.platform == 'darwin';
+bool get isMacOS => process?.platform == 'darwin';
 
-bool get isNode => true;
+bool get isNode => process != null;
```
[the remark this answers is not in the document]
jerivas: Huh, where would we use `isJS`?

jerivas:
@nex3 looks like the upgrade to clk_pkg 2.4.2 broke [some tests](https://github.com/sass/dart-sass/actions/runs/4420834123):

```
Process `node /Users/runner/work/dart-sass/dart-sass/build/npm/sass.js --interactive --color` exited with exitCode 1. Output:
[e] /Users/runner/work/dart-sass/dart-sass/build/npm/sass.js:4
[e] library.load({
[e]         ^
[e] 
[e] TypeError: library.load is not a function
[e]     at Object.<anonymous> (/Users/runner/work/dart-sass/dart-sass/build/npm/sass.js:4:9)
[e]     at Module._compile (node:internal/modules/cjs/loader:1254:14)
[e]     at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)
[e]     at Module.load (node:internal/modules/cjs/loader:1117:32)
[e]     at Module._load (node:internal/modules/cjs/loader:958:12)
[e]     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
[e]     at node:internal/main/run_main_module:23:47
[e] 
[e] Node.js v18.14.2
```

The call to `load()` fails when `cli_pkg` outputs this section due to the presence of `jsEsmExports`.

https://github.com/google/dart_cli_pkg/blob/187fe8c093a8dfdf253ff37a9a58f23e956b041a/lib/src/npm.dart#L461-L473

.github/workflows/ci.yml
```diff
@@ -167,6 +167,27 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: browser-actions/setup-chrome@v1
+    - uses: dart-lang/setup-dart@v1.2
+      with:
+        sdk: ${{ matrix.dart-channel }}
+    - run: dart pub get
+    - run: dart run grinder before-test
+    - run: google-chrome --version
+    - run: dart run test --platform chrome -j 2 -t browser -r expanded
```
jerivas, line 188: For some reason [all tests are running in Chrome](https://github.com/sass/dart-sass/actions/runs/4444257847/jobs/7802292764?pr=1895), not only those tagged with `browser`. Am I missing something?

test/browser_test.dart
```diff
@@ -0,0 +1,38 @@
...
+  test('renderSync() is not available', () {
+    expect(sass['renderSync'], isNull);
+  });
+  test('compileString() produces output', () {
+    var result = sass.callMethod('compileString', ['foo {bar: baz}']);
+    expect(result['css'], equals('foo {\n  bar: baz;\n}'));
+  });
+  test('compileStringAsync() produces output', () async {
+    var result = sass.callMethod('compileStringAsync', ['foo {bar: baz}']);
+    result = await promiseToFuture<JsObject>(result as Promise);
```
jerivas, line 35: This is failing with `TypeError: Instance of 'JsObject': type 'JsObject' is not a subtype of type 'Promise'`. Not sure what to do

jerivas (commented):
I added some basic browsers tests. Most of them are already passing, but I'm having trouble witha  few things

lib/src/io/node.dart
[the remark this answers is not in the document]
jerivas: See the Map class in https://pub.dev/packages/js

.github/workflows/ci.yml
jerivas: Don't use tags, use @TestOn and separate vm from browsers tests

test/browser_test.dart
jerivas: Let's declare the API for the `compile*` methods and call them directly instead of `callMethod`. Additionally figure out if we can get the spec tests to run in the browser, if that doesn't happen let's test a limited surface of the API in this repo (including basic compilation and forbidden methods)

test/browser_test.dart
```diff
@@ -0,0 +1,51 @@
...
+@JS()
+external Sass get sass;
+
+@JS()
+class Sass {
+  external NodeCompileResult compileString(String text,
+      [CompileStringOptions? options]);
+  external Promise compileStringAsync(String text,
+      [CompileStringOptions? options]);
+}
```
jerivas, line 20: @nex3 I'm using `@JS()` instead of `context['sass']`, but because of the imports I brought in to type it, I'm getting the following error:

```
lib/src/node/url.dart:11:7:
Error: Non-static JS interop class 'JSUrl' conflicts with natively supported class 'Url' in 'dart:html'.
class JSUrl {
      ^
Error: Compilation failed.
```

I tried going over the codebase and using `Url` from `dart:html`, but it's missing the constructor, which I assume is why we have our own class in the repo. Not sure where to go next 😅

.github/workflows/ci.yml
jgerigmeyer: Fixed in https://github.com/oddbird/dart-sass/commit/e238290c1738e761d3ebba5ae40016a754fa5268

test/browser_test.dart
nex3: Do you need `dart:html` at all here? In theory the browser support should just be a subset of Node support rather than containing anything browser-specific. Or does `dart:html` get brought in by the test framework itself?

jerivas: I'm not including it directly in the test, but it seems to only happen during test runs. I think we're hitting this [bug](https://github.com/dart-lang/sdk/issues/42200), specifically:v

> The users noticed that the API is hidden, so they tried using @JS to access it directly. This creates an issue because dart2js associates the dart:html interceptor with those types, so their JS-interop interfaces don't work

nex3: Actually, it's possible you can get away with using the `dart:html` bindings even in Node if you just use `import 'dart:html' show Url`. That's probably worth at least trying before you start working around this.

jgerigmeyer: Hm, I added `@anonymous` in https://github.com/oddbird/dart-sass/commit/7e51c9b326a1f5e6df71cd4e959bc85359ee84ed, but didn't mess with `jsEval()` and the tests seem to be passing. @nex3 Does that make sense?

pubspec.yaml
```diff
@@ -35,12 +35,12 @@ dependencies:
 dev_dependencies:
   analyzer: ^4.7.0
   archive: ^3.1.2
-  cli_pkg: ^2.4.2
+  cli_pkg: ^2.4.3
```
jgerigmeyer, line 38: This update is causing CI to fail on Node v14. I submitted an [issue](https://github.com/google/dart_cli_pkg/issues/124) and [PR to fix it](https://github.com/google/dart_cli_pkg/pull/125).

test/browser_test.dart
```diff
@@ -0,0 +1,47 @@
...
+  test('compile() is not available', () {
+    expect(() => (sass as dynamic).compile, throwsA(isA<NoSuchMethodError>()));
+  });
+  test('render() is not available', () {
+    expect(() => (sass as dynamic).render, throwsA(isA<NoSuchMethodError>()));
+  });
+  test('renderSync() is not available', () {
+    expect(
+        () => (sass as dynamic).renderSync, throwsA(isA<NoSuchMethodError>()));
+  });
```
jerivas, line 37: It looks like these tests don't work. I wanted to assert that these methods are not included in the ESM export, but even when the export includes the methods these will pass. It looks like I'm just asserting the `Sass` type defined above.

@nex3 that makes me wonder: do we want to completely omit these methods from the ESM export, or include them but have them raise an error? For the former, how can I test for their absence here? And for the latter, what should the error say?

test/browser_test.dart
nex3: If it works, it works! JS interop is an eternal mystery.

test/browser_test.dart
nex3: I think throwing a runtime error is better. It should say something like "<method name> is only supported on Node.js."

pubspec.yaml
jgerigmeyer: Fixed in https://github.com/oddbird/dart-sass/commit/38f1f5ffdd0f42131692ca78a120636a72b3ecaa

ntkme:
> Throw errors if trying to call compileString* without explicit importers

We already have `Throw if trying to access the file system` on the list, and we can just throw an error when import is attempted instead of at the time of the compile call. E.g. I should be able to compileString `a: {b: c}` without explicit importers as it doesn't need one.

lib/src/io/node.dart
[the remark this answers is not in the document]
jgerigmeyer: @jerivas We didn't discuss this `isJS` request -- is that something you're planning to add?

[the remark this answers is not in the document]
jgerigmeyer: - Add a new `isJS` getter which will be true for both Node and browser contexts, hardcoded to `false` in `vm.dart`.
- Browser contexts can check for `isJS && !isNode`, or we can add a new `isBrowser`
- Existing uses of `isNode` can probably be replaced with `isJS`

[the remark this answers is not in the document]
jerivas: Added in bbc867244c9c40a52f3210302891864cde0fc208

test/browser_test.dart
jerivas: Need to make JS API tests run in the browser, both in this repo and in `sass/sass-spec`

test/browser_test.dart
jerivas: Done in 716f785c1891be3da06b3388cb318d0fbf2078ed and bbc867244c9c40a52f3210302891864cde0fc208

test/browser_test.dart
```diff
@@ -0,0 +1,97 @@
...
+            (err as dynamic).message ==
+            'The compile() method is only available in Node.js.')));
+  });
+
+  test('render() is not available', () {
+    expect(
+        () => sass.render(RenderOptions(), (error, result) {}),
+        throwsA(predicate((err) =>
+            (err as dynamic).message ==
+            'The render() method is only available in Node.js.')));
```
jerivas, line 54: This is the only test failing so far with:

```
Expected: throws satisfies function
    Actual: <Closure 'main__closure5'>
     Which: threw JSObject:<TypeError: Instance of 'main___closure': type 'UnknownJavaScriptObject' is not a subtype of type '(Object?, RenderResult?) => void'>
```

I think the noop callback I'm passing here is failing type checks here, but I'm not sure how to fix it. Would appreciate any pointers @jgerigmeyer @nex3

jerivas:
@nex3 re-requested review to get your thoughts on the currently completed items in the checklist. That being said, we can wait until we complete all of them to review everything at once

lib/src/io/node.dart
```diff
@@ -76,15 +94,15 @@ Future<String> readStdin() async {
   });
   // Node defaults all buffers to 'utf8'.
   var sink = utf8.decoder.startChunkedConversion(innerSink);
-  process.stdin.on('data', allowInterop(([Object? chunk]) {
+  process?.stdin.on('data', allowInterop(([Object? chunk]) {
```
nex3, line 97: You can avoid nullability issues here by replacing the `isNode` check at the beginning with

```dart
var process_ = process;
if (process_ == null) {
  throw UnsupportedError("readStdin() is only supported on Node.js");
}
```

and then just referring to `process_` here, which the Dart compiler should understand as non-nullable.

lib/src/io/node.dart
```diff
@@ -187,32 +223,44 @@ T _systemErrorToFileSystemException<T>(T callback()) {
...
 }
 
-final stderr = Stderr(process.stderr);
+Stderr get stderr {
+  var stderr_ = process?.stderr;
+  if (stderr_ == null) {
+    throw UnsupportedError("stderr is only supported on Node.js");
+  }
+  return Stderr(stderr_);
+}
```
nex3, line 232: This will create a new `Stderr` object every time it's referenced, rather than initializing it once and re-using it. You can put complex logic in a one-time initializer using an IIFE:

```suggestion
final Stderr stderr = () {
  var stderr_ = process?.stderr;
  if (stderr_ == null) {
    throw UnsupportedError("stderr is only supported on Node.js");
  }
  return Stderr(stderr_);
}();
```

lib/src/io/node.dart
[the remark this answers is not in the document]
nex3: Shouldn't need this anymore I think

lib/src/io/node.dart
```diff
@@ -187,32 +223,44 @@ T _systemErrorToFileSystemException<T>(T callback()) {
...
+bool get isJS => true;
 
-bool get isNode => true;
+bool get isNode => process != null;
 
 // Node seems to support ANSI escapes on all terminals.
 bool get supportsAnsiEscapes => hasTerminal;
 
-String get currentPath => process.cwd();
+String get currentPath => process?.cwd() ?? '/';
```
nex3, line 251: This looks unused; you can delete it from here and the other two IO files.

lib/src/io/interface.dart
```diff
@@ -37,7 +37,10 @@ bool get isMacOS => throw '';
 /// Returns whether or not stdout is connected to an interactive terminal.
 bool get hasTerminal => throw '';
 
-/// Whether we're running as Node.JS.
+/// Whether we're running as JS (browser or Node).
+bool get isJS => throw '';
+
+/// Whether we're running as Node.JS (not browser).
```
nex3, line 43:
```suggestion
/// Whether we're running as Node.JS (not browser or Dart VM).
```

test/browser_test.dart
nex3: You need to call [`allowInterop()`](https://api.dart.dev/stable/2.9.0/dart-js/allowInterop.html) on every Dart callback you pass to JS.

dart_test.yaml
```diff
@@ -1,3 +1,8 @@
 tags:
   # Indicates tests that require Node.js.
   node:
+
+override_platforms:
+  chrome:
+    settings:
+      arguments: --no-sandbox
```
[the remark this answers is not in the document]
jgerigmeyer: @jerivas I'm not sure this is actually needed? Seems like I can run the tests without it...

tool/grind.dart
```diff
@@ -44,6 +46,17 @@ void main(List<String> args) {
   pkg.standaloneName.value = "dart-sass";
   pkg.githubUser.fn = () => Platform.environment["GH_USER"];
   pkg.githubPassword.fn = () => Platform.environment["GH_TOKEN"];
+  // Limit the methods exposed as ESM exports to those that don't require
+  // filesystem access
```
jgerigmeyer, line 50: Is this still accurate?

lib/src/io/node.dart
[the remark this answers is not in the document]
jgerigmeyer: Removed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

test/browser_test.dart
jgerigmeyer: Fixed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

lib/src/io/node.dart
jgerigmeyer: Fixed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

lib/src/io/node.dart
jgerigmeyer: Fixed in Fixed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

lib/src/io/node.dart
jgerigmeyer: Removed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

tool/grind.dart
jgerigmeyer: Removed in https://github.com/oddbird/dart-sass/commit/1937affd6fa9fc3965c72222f323f56fb53352d7

lib/src/visitor/async_evaluate.dart
```diff
@@ -1601,6 +1602,9 @@ class _EvaluateVisitor
       _importSpan = span;
 
       var importCache = _importCache;
+      if (isBrowser && importCache == null && _importer == null) {
+        throw "Custom importers are required to `@use` or `@import` when compiling in the browser.";
+      }
```
jgerigmeyer, line 1607: @nex3 I think I got a bit carried away in https://github.com/oddbird/dart-sass/commit/7feffba46b25157c293cd18891650cccba47cc0b, making sure that both `importCache` and `_importer` would be null here. I think that's required for `_importer`, but wouldn't be necessary if there's a way to check whether `importCache` has any `_importers` defined? I'd be happy for your feedback on this. 🙏

ntkme:
From a design perspective, it might be better to having a default `BrowserImporter` that literally just throw an error message when import is attempted. Use it to replace `NoOpImporter` or `FileSystemImporter` under browser environment. - This way you can avoid the need of making `importer` parameter `nullable` and perhaps save a few additional `null` checks.

lib/src/visitor/async_evaluate.dart
jgerigmeyer: > From a design perspective, it might be better to having a default `BrowserImporter` that literally just throw an error message when import is attempted. Use it to replace `NoOpImporter` or `FileSystemImporter` under browser environment. - This way you can avoid the need of making `importer` parameter `nullable` and perhaps save a few additional `null` checks.

@ntkme What do you think about this approach? https://github.com/sass/dart-sass/pull/1895/files/49182bdddad87cb7ca5d14862ea76a028d4e52e6..18d9c985b9823e25f0e21bd041d16398db73dc3e

I retained `NoOpImporter` for the base importer, but then added a default `BrowserImporter` that throws an error if no other `importers` are defined. That way users can provide a custom `importer` or `importers` if they want, but if both are missing then it will throw an error. One downside I see is that if a user-supplied `importer` does not find the imported stylesheet and the request falls through to `importers`, I think it will now throw the "Custom importers are required" error instead of a more helpful "Can't find stylesheet to import" error.

ntkme: I'm not that familiar with logic around this, maybe something like `nodeImporter ? NoOpImporter() : BrowserImporter()`? E.g. If user defined an importer, you can use `NoOp` so that it does not fall-over and throw, if user did not define the importer, we pick our `BrowserImporter` to throw.

jgerigmeyer: Cleaned up a bit and resolved the mentioned issue: https://github.com/sass/dart-sass/pull/1895/files/49182bdddad87cb7ca5d14862ea76a028d4e52e6..aa3afabda16fe95cb899fdef98c3308811318069

@nex3 @ntkme What do you think of this solution?

ntkme: Much simpler. LGTM

lib/src/async_import_cache.dart
```diff
@@ -122,6 +126,12 @@ class AsyncImportCache {
       {AsyncImporter? baseImporter,
       Uri? baseUrl,
       bool forImport = false}) async {
+    if (isBrowser &&
+        (baseImporter == null || baseImporter is NoOpImporter) &&
+        _importers.isEmpty) {
+      throw "Custom importers are required to `@use` or `@import` when compiling in the browser.";
```
nex3, line 132: Nit: I'd say something like "to load stylesheets" here, in case users get confused and thing that other structures like `@forward` or `meta.load-css()` are allowed.

lib/src/visitor/async_evaluate.dart
nex3: Looks good to me as well!

lib/src/async_import_cache.dart
jerivas: Fixed in c3b373eeac4953f592dc8ff0c13771131ccbe4ec

tool/grind.dart
```diff
@@ -54,6 +54,7 @@ void main(List<String> args) {
     'info',
     'render',
     'renderSync',
+    'Exception',
```
jgerigmeyer, line 57: @jerivas I assume it's okay to expose this? The JS API Spec tests were using it.

.github/workflows/ci.yml
```diff
@@ -103,6 +103,52 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - uses: browser-actions/setup-chrome@v1
+      - uses: dart-lang/setup-dart@v1.2
+        with:
+          sdk: ${{ matrix.dart-channel }}
+      - run: dart pub get
+
+      - name: Check out sass-spec
+        uses: sass/clone-linked-repo@v1
+        # TODO update this to use "sass/sass-spec" directly
```
jgerigmeyer, line 125: Remember to update this before merging.

nex3: If you send a PR to the official sass-spec repo, you can link to it in the PR description here and this PR will automatically run against that one.

lib/src/io/node.dart
```diff
@@ -27,20 +30,33 @@ class FileSystemException {
...
+    }
+  }
 
   void writeln([Object? object]) {
-    _stderr.write("${object ?? ''}\n");
+    if (_stderr == null) {
+      console.error("${object ?? ''}\n");
+    } else {
+      _stderr?.write("${object ?? ''}\n");
+    }
```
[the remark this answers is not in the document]
nex3: I think a better solution than stubbing out stderr in the browser might be to have a separate `BrowserLogger` that directly calls `console.error()`.

lib/src/io/node.dart
```diff
@@ -27,20 +30,33 @@ class FileSystemException {
...
 class Stderr {
-  final Writable _stderr;
+  final Writable? _stderr;
 
   Stderr(this._stderr);
 
-  void write(Object object) => _stderr.write(object.toString());
+  void write(Object object) {
+    if (_stderr == null) {
+      console.error(object.toString());
```
nex3, line 39: Note that `stderr.write()` doesn't append a newline, while `console.error()` does. If we _do_ decide we need to fake stderr in the browser, we'll need to build up a buffer here and only emit it once we hit an actual newline.

nex3 (commented):
Other than the logger stuff, this is looking very good!

.github/workflows/ci.yml
jgerigmeyer: Fixed.

lib/src/io/node.dart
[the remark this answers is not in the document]
jgerigmeyer: @nex3 What do you think of the revised approach in https://github.com/sass/dart-sass/pull/1895/commits/d30dd424bd7a0dd973048dadd0585bd6850ad2bf?

Screenshot 2023-05-15 at 3 01 24 PM

lib/src/logger.dart
```diff
@@ -21,7 +23,12 @@ abstract class Logger {
 
   /// Creates a logger that prints warnings to standard error, with terminal
   /// colors if [color] is `true` (default `false`).
-  const factory Logger.stderr({bool color}) = StderrLogger;
+  factory Logger.stderr({bool color = false}) {
```
nex3, line 26: Unfortunately, converting a factory constructor from const to non-const is a breaking change in Dart.  This means that instead of a separate `BrowserLogger` class, you'll need to use `if (isBrowser)` in `StderrLogger` itself. Sorry I didn't catch this earlier.

nex3 (approved):
A few relatively small comments, but overall looks great!

.github/workflows/ci.yml
```diff
@@ -218,6 +218,54 @@ jobs:
         run: npm run js-api-spec -- --sassPackage ../embedded-host-node --sassSassRepo ../language
         working-directory: sass-spec
 
+  sass_spec_js_browser:
+    name: "JS API Tests | Pure JS | Dart ${{ matrix.dart_channel }} | Browser"
```
nex3, line 222: To make this a little easier to differentiate in the web UI, let's make it

```suggestion
    name: "JS API Tests | Browser | Dart ${{ matrix.dart_channel }}"
```

since "browser" implies "pure JS" anyway.

.github/util/initialize/action.yml
```diff
@@ -10,7 +10,7 @@ runs:
   steps:
     - uses: dart-lang/setup-dart@v1
       with:
-        sdk: "${{ inputs.sdk }}"
+        sdk: "${{ inputs.dart-sdk }}"
```
jgerigmeyer, line 13: @nex3 I think this is a typo in the `main` branch.

lib/src/logger/stderr.dart
```diff
@@ -9,8 +9,9 @@ import 'package:stack_trace/stack_trace.dart';
 import '../io.dart';
 import '../logger.dart';
 import '../utils.dart';
+import '../node/console_stub.dart' if (dart.library.js) '../node/console.dart';
```
jgerigmeyer, line 12: @nex3 Is there a better way to do this that still avoids `Dart library 'dart:js_util' is not available on this platform` errors?

lib/src/node/console_stub.dart
```diff
@@ -0,0 +1,12 @@
...
+// https://opensource.org/licenses/MIT.
+
+abstract class Console {
+  external void log(String data);
+  external void info(String data);
+  external void warn(String data);
+  external void error(String data);
+}
+
+Console get console => throw '';
```
jgerigmeyer, line 12: This seems a bit silly, but it doesn't seem like you can conditionally import a package without also having a fallback package.

jgerigmeyer (commented):
@nex3 Ready for another round of review (https://github.com/sass/dart-sass/pull/1895/commits/564a4ce372a0a02fc4681c285dcf2b58e4dcea81).

lib/src/logger/stderr.dart
nex3: Unfortunately no :(. The Dart VM still needs to know how to typecheck the APIs even if it won't actually run them in practice.

I think, on retrospect, that the cleanest approach here is probably to hide even more of the implementation and just expose a `printError()` function. I don't want to make you undo all the refactors so I've just gone and made the change myself.

nex3 (approved):

lib/src/logger/stderr.dart
jgerigmeyer: @nex3 Makes sense, and LGTM! 🚀

PR #1903, opened 2023-03-03
Track original source spans for selectors

### LINKED ISSUE #1783: Add spans to the Selector AST
Currently the selector AST used by Sass and exposed by `sass_api` doesn't expose source span information. This information is tricky to determine, because selectors are interpolated in the source file. However, it should be possible to map the source spans from the parsed string back to the original file by tracking where the source offsets of the interpolated sections begin and end, and thereby get accurate spans for the selectors.

This would allow us to improve a number of error/deprecation messages, and make the Kythe indexer implementation cleaner.

nex3:
Closes #1783

See sass/sass-spec#1887


lib/src/interpolation_map.dart
```diff
@@ -0,0 +1,180 @@
...
+
+import 'package:charcode/charcode.dart';
+import 'package:source_span/source_span.dart';
+import 'package:string_scanner/string_scanner.dart';
+
+import 'ast/sass.dart';
+import 'util/character.dart';
+
+/// A class that can map locations in a string generated from an [Interpolation]
+/// to the original source code in the [interpolation].
```
jathak, line 15: I think this reference is causing the docs to fail

lib/src/interpolation_map.dart
```diff
@@ -0,0 +1,180 @@
...
+  /// Each of these indicates the location in the generated string that
+  /// corresponds to the end of the component at the same index of
+  /// [_interpolation.contents]. Its length is always one less than
+  /// [_interpolation.contents] because the last element always ends the string.
+  final List<SourceLocation> _targetLocations;
+
+  /// Creates a new interpolation map that maps the given [targetLocations] in
+  /// the generated string to the contents of the interpolation.
+  ///
+  /// Each [targetLocation] at index `i` corresponds to the character in the
```
jathak, line 31: Also this one

jathak (approved):

PR #1906, opened 2023-03-10
Fix macOS arm64 build and release a new version

jathak:


Goodwine (approved):

ntkme:
@jathak @Goodwine We need to also revert the same change in dart-sass-embedded.

PR #1919, opened 2023-03-21
Update behavior for sass/sass#3497

nex3:
See https://github.com/sass/sass-spec/pull/1889

### DIFF
--- a/lib/src/value/color.dart
+++ b/lib/src/value/color.dart
@@ -123,7 +123,7 @@ class SassColor extends Value {
       case ColorSpace.oklab:
       case ColorSpace.lch:
       case ColorSpace.oklch:
-        return fuzzyEquals(channel0, 0);
+        return fuzzyEquals(channel0, 0) || fuzzyEquals(channel0, 100);
 
       default:
         return false;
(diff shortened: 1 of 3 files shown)


lib/src/value/color.dart
```diff
@@ -156,11 +156,13 @@ class SassColor extends Value {
...
       case ColorSpace.oklab:
-        return fuzzyEquals(channel0, 0);
+        return fuzzyEquals(channel0, 0) || fuzzyEquals(channel0, 100);
 
       case ColorSpace.lch:
       case ColorSpace.oklch:
-        return fuzzyEquals(channel0, 0) || fuzzyEquals(channel1, 0);
+        return fuzzyEquals(channel0, 0) ||
+            fuzzyEquals(channel0, 100) ||
+            fuzzyEquals(channel1, 0);
```
Goodwine, line 165: Being unfamiliar with channels and power, should `channel1` be compared against 100 too?

Goodwine: Nitpicking a bit: I'm confused by the function called `isChannel2Powerless` where `channel0` and `channel1` are checked, but not `channel2`. I'm assuming that's intended, but I don't understand why xD

Goodwine (approved):
LGTM but I have a couple questions

lib/src/value/color.dart
nex3: See https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#powerless-components. For example, for an LCH color, the hue is powerless if the chroma is zero because the color is grayscale and so hue doesn't matter anymore.

PR #1922, opened 2023-03-23
Add support for constants in calculations

### LINKED ISSUE #1900: Add support for calc constants
See https://github.com/sass/sass/issues/3258

nex3:
Closes #1900
See sass/sass#3258
See sass/sass-spec#1890

### DIFF
--- a/test/cli/shared/repl.dart
+++ b/test/cli/shared/repl.dart
@@ -302,13 +302,15 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
     group("and colorizes", () {
       test("an error in the source text", () async {
         var sass = await runSass(["--interactive", "--color"]);
-        sass.stdin.writeln("max(2, 1 + blue)");
+        sass.stdin.writeln("@use 'sass:math'");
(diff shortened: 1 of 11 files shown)


test/cli/shared/repl.dart
```diff
@@ -195,12 +195,14 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
 
     test("a runtime error", () async {
       var sass = await runSass(["--interactive"]);
-      sass.stdin.writeln("max(2, 1 + blue)");
+      sass.stdin.writeln("@use 'sass:math'");
```
Goodwine, line 198: Just to confirm, this change is not required as part of the PR, and it is just a fly-by cleanup, right?

Goodwine (approved):

test/cli/shared/repl.dart
nex3: No, it's required—we actually consider `max(2, 1 + blue)` to be valid CSS now because `max()` is a calculation context and `blue` is an identifier.

PR #1926, opened 2023-03-27
Allow global Sass function colliding with CSS native functions to use CSS variables

Goodwine:
Many Sass functions are available globally even without loading their module. Some of these are also valid CSS native functions. Sass performs validations which disallow the use of CSS variables because the arguments are asserted a given type of value. For these collisions allow the use of CSS variables and in such cases assume the entire function call is meant to be the CSS native function rather than the global Sass function.

Fixes https://github.com/sass/sass/issues/3507

See https://github.com/sass/sass-spec/pull/1892

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 * Deprecate the use of multiple `!global` or `!default` flags on the same
   variable. This deprecation is named `duplicate-var-flags`.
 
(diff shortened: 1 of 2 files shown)


nex3:
This needs a language proposal before the implementation can land.

nex3 (approved):
Please add a CHANGELOG entry, then LGTM

PR #1933, opened 2023-04-11
Don't separate out media queries after one has bubbled

### LINKED ISSUE #777: nested at-rules breaks at-rule merging
I recently moved from node-sass 4.12.0 to (dart-)sass 1.22.7

Upon switching, I noticed that some at-rules end up being repeated. I've tracked down the issue where upon using a nested at-rule, it breaks subsequent rules.

My project that shows the issue is here: https://github.com/clshortfuse/materialdesignweb/blob/3964f7d72f2d6753d9691ed70c8b4fc60cb52dd7/components/layout/_spec.scss

I've a smaller version of the issue available here:

https://www.sassmeister.com/gist/a66eb997749a1c088dd290c30bfd0003

On a related sidenote, I actually had rewritten a good chunk of cssnano's merge-rules plugin to do this in post-processing, and it's been commited to master, but cssnano hasn't received a version bump yet.

https://github.com/cssnano/cssnano/pull/722/files

nex3:
Closes #777
See sass/sass-spec#1898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
   native CSS `filter` functions. This is in addition to number values which were
   already allowed.
 
+* Fix a cosmetic bug where an outer rule could be duplicated after nesting was
+  resolved, instead of re-using a shared rule.
+
 ## 1.61.0
 
 * **Potentially breaking change:** Drop support for End-of-Life Node.js 12.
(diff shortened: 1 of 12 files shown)


Goodwine (approved):
Needs the synchronized file, but LGTM

PR #1938, opened 2023-04-17
Fix :has(+ &)

### LINKED ISSUE #1937: Broken :has with modifier compilation
I'm on v1.62.0 (latest), and the following:

```scss
.foo {
  &:has(+ &) {
    display: none;
  }
}
```

Is being compiled to this:

```css
.foo:has(.foo) {
  display: none;
}
```

While instead it should be compiled to this:

```css
.foo:has(+ .foo) {
  display: none;
}
```

Basically the `+` inside the `:has` disappeared.

nex3:
Closes #1937
See sass/sass-spec#1900

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.62.1
+
+* Fix a bug where `:has(+ &)` and related constructs would drop the leading
+  combinator.
+
 ## 1.62.0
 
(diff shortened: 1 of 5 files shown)


tbondwilkinson (approved):
I'm assuming that the spec change is the test for this functionality, and there is no dart-sass unit test to update?

nex3:
Yep, that's right.

PR #1948, opened 2023-05-02
Include pubspec.lock, for reproducible builds

lelgenio:
I'm requesting this to lower the burden on package maintainers from having individual copies of this file.
This may also help pervent issue caused by accidental API changes on dependencies.

Looking at the git history it seems like the decision to ignore pubspec.lock was simply inherited from a template gitignore, since the first commit  cf28c7ed200691828e34aade7ba1629e2ebb4757.


nex3:
Because Dart Sass is distributed as a Dart library as well as a binary release, it's expected to work with _all_ compatible version constellations of its dependencies. We don't include a lockfile because it can lead to us accidentally testing on only those specific package versions, even though our users will potentially be using different versions.

Issue #1889, opened 2023-02-17
Distribute native dart sass as npm package


7rulnik:
It would be nice to have the option to install and manage native dart sass like a normal node.js dependency via npm/yarn/pnpm.
Currently, it's possible to download it and adjust path variable or install it using brew, but with these options, it's hard to sync and manage versions in teams with different people.

I guess we can make something like esbuild doing here https://github.com/evanw/esbuild/blob/main/lib/npm/node-install.ts
Or like swc https://github.com/swc-project/swc/blob/main/postinstall.js

nex3:
If you're already using Node, why not use the Node.js API? The `sass-embedded` package already uses the Dart VM, so if you need speed that's a good option.

7rulnik:
@nex3 I want to use it via CLI. As I understand there is no option for it in sass-embedded

nex3:
This is something we're open to including as part of Embedded Dart Sass and the Node.js embedded host, although we're unlikely to have bandwidth to work on it ourselves. Here's what would need to happen for this to work:

* We'd need to expose the CLI `main()` method through the `sass_api` package entrypoint.
* We'd need to provide a flag on the Embedded Dart Sass executable (`--cli` maybe) to make it run in CLI mode instead of as an embedded compiler.
* We'd need to add an executable to the `sass-embedded` package that invokes this CLI.

ntkme:
We can upload dart-sass binary just like the `optionalDependencies` of dart-sass-embedded, and then have a "dart-sass" npm meta package which only declares the `optionalDependencies`. This would allow us to provide a dart-only package without any JS API support, which I think it what this issue is asking for.

7rulnik:
And naive wrapper implementation could be like this:

```js
#!/usr/bin/env node

const path = require('path')
const foo = require.resolve(`sass-embedded-${process.platform}-${process.arch}/package.json`)
const pathToBinary =  path.resolve(foo, '../binary/sass')

require('child_process').execFileSync(
  pathToBinary,
  process.argv.slice(2),
  { stdio: 'inherit' },
);
```

But I think that main problem here — release process. At least for me, because I don't famillar with dart at all

nex3:
[...]

PR #1955, opened 2023-05-10
Merge the Dart Sass embedded compiler repo into the Dart Sass repo

nex3:
See #1889
See sass/embedded-host-node#218

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,48 @@ on:
   pull_request:
 
 jobs:
+  format:
+    name: Code formatting
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dart-lang/setup-dart@v1
+    - run: dart format --fix .
+    - run: git diff --exit-code
+
+  static_analysis:
+    name: Static analysis
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dart-lang/setup-dart@v1
+      - run: dart pub get
+
+      - uses: ./.github/util/build-protobuf
+        with: {github-token: "${{ github.token }}"}
+
+      - name: Analyze Dart
+        run: dart analyze --fatal-warnings ./
+
+  dartdoc:
+    name: Dartdoc
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dart-lang/setup-dart@v1
+      - run: dart pub get
+      - name: dartdoc sass
+        run: dart run dartdoc --quiet --no-generate-docs
+          --errors ambiguous-doc-reference,broken-link,deprecated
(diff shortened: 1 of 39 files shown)


.github/workflows/ci.yml
```diff
@@ -103,6 +152,72 @@ jobs:
         run: npm run js-api-spec -- --sassSassRepo ../language --sassPackage ../build/npm
         working-directory: sass-spec
 
+  # The versions should be kept up-to-date with the latest LTS Node releases.
+  # They next need to be rotated October 2021. See
```
jathak, line 156: Should we rotate these now, or save that for a separate PR?

pubspec.lock
jathak, line 1: Is there a reason why we're now checking this in?

jathak (approved):

.github/workflows/ci.yml
```diff
@@ -469,58 +576,46 @@ jobs:
...
-      - uses: dart-lang/setup-dart@v1
-      - uses: frenck/action-setup-yq@v1
-        with: {version: v4.30.5} # frenck/action-setup-yq#35
 
       - name: Get version
         id: version
-        run: |
-          echo "sass=${GITHUB_REF##*/}" | tee --append $GITHUB_OUTPUT
-          echo "sass_api=$(curl --fail --silent --show-error --location https://raw.githubusercontent.com/sass/dart-sass/${GITHUB_REF##*/}/pkg/sass_api/pubspec.yaml | yq .version)" | tee --append $GITHUB_OUTPUT
+        run: echo "::set-output name=version::${GITHUB_REF##*/}"
```
ntkme, line 592:
```suggestion
        run: echo "version=${GITHUB_REF##*/}" | tee --append $GITHUB_OUTPUT
```

`::set-output` is deprecated:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

ntkme (changes requested):

.github/workflows/ci.yml
ntkme: Looks like we may have a few of these. Maybe we should do this separately in a different PR.

.github/workflows/ci.yml
nex3: Separate PR, I think.

pubspec.lock
nex3: Oops, got caught in the merge.

.github/workflows/ci.yml
nex3: Yeah, I'd prefer to save this for a separate PR.

ntkme:
@nex3 Are we going to release 1.63.0 as is or are we waiting for some other changes to land? I think we should cut a release now as we already archived dart-sass-embedded repo.

nex3:
We're going to release it as soon as we implement the other embedded protocol changes. I want to make sure all the breaking changes are rolled into a single release.

PR #1972, opened 2023-05-19
Document browser support

nex3:


jathak:
I think the name and description of this PR is wrong. This is for documenting running in a browser, right?

nex3:
Yeah, I think my editor used a cached description and I didn't notice.

jathak (approved):

PR #1974, opened 2023-05-23
Upgrade the analyzer package

nex3:


Goodwine (approved):
Are the changes on `synchronize.dart`  intended for this PR? 

I imagine these are changes due to new findings in the updated analyzer, but at the same time some changes look unrelated, so just making sure that's the case

PR #1985, opened 2023-06-02
Properly close resources

ntkme:
@nex3 This PR fixes the issue where the process would hang even after closing stdio from parent process.


ntkme:
Any compilation's behavior after stdin get closed is unpredictable, that it may get stuck if there is pending request to host as it can never receive the response. Therefore to make the behavior consistent, this implementation defines closing stdin as an indication of requesting compiler shutdown - all inflight and queued requests are effectively cancelled or ignored.

PR #1987, opened 2023-06-06
Release 1.63.0

nex3:


ntkme:
@nex3 We should add a change log entry under "Embedded Sass" section for embedded protocol 2.0.

nex3:
Good call, thanks! We should also update the embedded protocol's changelog.

jathak (approved):

PR #1988, opened 2023-06-07
Expose calculations in JS API

jerivas:
https://github.com/sass/sass/pull/3622
https://github.com/sass/sass-spec/pull/1918
https://github.com/sass/embedded-host-node/pull/237


lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,144 @@
...
+  var jsClass =
+      createJSClass('sass.SassCalculation', (Object self, [Object? _]) {
+    jsThrow(JsError("new sass.SassCalculation() isn't allowed"));
+  });
+
+  // Static methods
+  setProperty(jsClass, 'calc', allowInteropNamed('calc', calc));
+  setProperty(jsClass, 'min', allowInteropNamed('min', min));
+  setProperty(jsClass, 'max', allowInteropNamed('max', max));
+  setProperty(jsClass, 'clamp', allowInteropNamed('clamp', clamp));
```
jerivas, line 74: Is this the right way to add static methods?

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,144 @@
...
+          '`value` or `max` is undefined and neither `min` nor `value` is a SassString that begins with "var("'));
+    }
+    [min, value, max].whereNotNull().forEach(isCalculationValue);
+    return SassCalculation.unsimplified(
+        'clamp', [min, value, max].whereNotNull());
+  }
+
+  var jsClass =
+      createJSClass('sass.SassCalculation', (Object self, [Object? _]) {
+    jsThrow(JsError("new sass.SassCalculation() isn't allowed"));
```
jerivas, line 67: From the dart definition, I understand that direct instantiation is not allowed. Is that correct?

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,144 @@
...
+    'assertCalculation': (SassCalculation self, [String? name]) => self,
+  });
+
+  jsClass.defineGetters({
+    'name': (SassCalculation self) => self.name,
+    'arguments': (SassCalculation self) => self.arguments,
+  });
+
+  getJSClass(SassCalculation.unsimplified('calc', [SassNumber(1)]))
+      .injectSuperclass(jsClass);
```
jerivas, line 86: I got this approach from [`color.dart`](https://github.com/oddbird/dart-sass/blob/f5ce0906e7013eace2800f601870cb80193de712/lib/src/node/value/color.dart#L67), is it correct?

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,144 @@
...
+}();
+
+/// The JavaScript CalculationOperator class
+final JSClass calculationOperatorClass = () {
+  var jsClass = getJSClass(CalculationOperator.plus).superclass;
+  setProperty(jsClass, 'plus', CalculationOperator.plus);
+  setProperty(jsClass, 'minus', CalculationOperator.minus);
+  setProperty(jsClass, 'times', CalculationOperator.times);
+  setProperty(jsClass, 'dividedBy', CalculationOperator.dividedBy);
+  return jsClass;
```
jerivas, line 97: Is this the right approach to expose a Dart enum?

jerivas (commented):
@nex3 I'm getting a bunch of call stack errors, it sounds like I'm somehow messing up the methods and getters https://github.com/sass/dart-sass/actions/runs/5205889223/jobs/9391830839?pr=1988

lib/src/node/value/calculation.dart
nex3: That's right.

nex3 (commented):
> @nex3 I'm getting a bunch of call stack errors, it sounds like I'm somehow messing up the methods and getters https://github.com/sass/dart-sass/actions/runs/5205889223/jobs/9391830839?pr=1988

It's very frustrating that this doesn't actually include the stack trace! My suggestion would be to create a little test.js file that loads the compiled JS from `build/npm` and directly invokes the functions that are overflowing, and see if that gives more clarity on what the actual recursive call stack is.

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,128 @@
...
+        inMinMax: false, simplify: false);
+  });
+
+  jsClass.defineMethods({
+    'equals': (CalculationOperation self, Object other) => self == other,
+    'hashCode': (CalculationOperation self) => self.hashCode,
+  });
+
+  jsClass.defineGetters({
+    'operator': (CalculationOperation self) => self.operator.operator,
```
jerivas, line 102: ...unfortunately the call stack error is back for `CalculationOperation.operator`. Because in JS we want this to be a string and not the enum it is in Dart, we have to define this getter explicitly to get the nested `operator.operator`. Trying to access this property is what is causing the current test failures:

```
Uncaught RangeError: Maximum call stack size exceeded
    at Rti._generalAsCheckImplementation [as _as] (/workspaces/dart-sass/build/npm/sass.dart.js:2682:34)
    at _callDartFunctionFastCaptureThis (/workspaces/dart-sass/build/npm/sass.dart.js:7468:26)
    at CalculationOperation0.<anonymous> (/workspaces/dart-sass/build/npm/sass.dart.js:7455:18)
    at calculationOperationClass__closure2.call$1 (/workspaces/dart-sass/build/npm/sass.dart.js:82775:51)
    at Object.Primitives_applyFunction (/workspaces/dart-sass/build/npm/sass.dart.js:1202:30)
    at Object.Function_apply (/workspaces/dart-sass/build/npm/sass.dart.js:5464:16)
    at _callDartFunctionFastCaptureThis (/workspaces/dart-sass/build/npm/sass.dart.js:7472:16)
```

I visited each of those and couldn't find anything that's obviously wrong. This is `sass.dart.js:82775` for reference

```js
A.calculationOperationClass__closure2.prototype = {
  call$1($self) {
    return type$.CalculationOperation._as($self).operator.operator;
  },
  $signature: 372
};
```

In general, this seems to be a problem with redefining getters in the JS class that are already defined in the Dart class... but it's a pattern that's been used extensively in other files like [`color.dart`](https://github.com/oddbird/dart-sass/blob/13b4f337d7e57856512f8d13cc9bae04ca9e54c7/lib/src/node/value/color.dart#L56-L64), so I don't know why it's failing here 😕

jerivas: Thanks, that did the trick

jerivas:
@nex3  I'm working on the changes to `Options.functions`, and from my understanding we need to simplify return values as explained [here](https://github.com/sass/sass/blob/main/accepted/calculation-api.d.ts.md#functions): 

> Return a copy of `result.internal` with all calculations it transitively contains (including the return value itself if it's a calculation) replaced with the result of simplifying those calculations.

It sounds like we need to modify [`_parseFunctions`](https://github.com/oddbird/dart-sass/blob/71aa9a2149116f14f5ee65ab85d6d3d75869eeaa/lib/src/node/compile.dart#L233) to simplify the result if it's a `SassCalculation` instance, something like this:

```dart
if (result is Value) {
  if (result is SassCalculation)
    return SassCalculation._simplify(result);
  return result;
}
```

But `_simplify` is a private static method. Should we make it public? Am I approaching this correctly?

nex3:
@jerivas Since we _don't_ want `_simplify` to be used by users outside the Sass package (since they can never get ahold of an unsimplified calculation anyway), it may be simpler to just call `SassCalculation.calc()` instead, which will in practice simplify a calculation.

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,144 @@
...
+          minString = min.text;
+        } else if (min is CalculationInterpolation) {
+          minString = min.value;
+        } else {
+          jsThrow(JsError(
+              '`value` and `max` are both undefined and `min` is not a SassString or CalculationInterpolation'));
+        }
+        final values = minString
+            .split(',')
+            .map((v) => num.tryParse(v.trim()).andThen((v) => SassNumber(v)))
```
jerivas, line 65: @nex I'm trying to parse the `min` argument of `clamp()` as a comma-separated list of `SassNumber`s. This approach works well with something like `1, 2, 3`, but fails if the numbers contain units: `1px, 2rem, 3ch`. I can extract the units for `SassNumber` with a regex, but wanted to check with you if there's something in the codebase that already does this

jgerigmeyer: @jerivas I think you mean @nex3 👍🏻

lib/src/node/compile.dart
```diff
@@ -239,6 +265,7 @@ List<AsyncCallable> _parseFunctions(Object? functions, {bool asynch = false}) {
       late Callable callable;
       callable = Callable.fromSignature(signature, (arguments) {
         var result = (callback as Function)(toJSArray(arguments));
+        result = simplify(result);
```
nex3, line 268: Style nit:

```suggestion
        var result = simplify((callback as Function)(toJSArray(arguments)));
```

Generally good to avoid reassigning variables if possible.

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,134 @@
...
+      var argList = jsToDartList(arguments).cast<Object>();
+      argList.forEach(assertCalculationValue);
+      return SassCalculation.unsimplified('max', argList);
+    },
+    'clamp': (Object min, [Object? value, Object? max]) {
+      if (value == null) {
+        assertUnquotedStringOrInterpolation(min);
+      } else if (max == null) {
+        assertUnquotedStringOrInterpolation(min);
+        assertUnquotedStringOrInterpolation(value);
```
nex3, line 69: If `max` is null, only one of these needs to be an unquoted string or interpolation, not both. (The idea is to allow things like `clamp(#{1px, 2px}, 3px)`.)

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,134 @@
...
+import 'package:collection/collection.dart';
+import 'package:node_interop/js.dart';
+import 'package:sass/src/node/immutable.dart';
+import 'package:sass/src/node/utils.dart';
+
+import '../../value.dart';
+import '../reflection.dart';
+
+/// Check that [arg] is a valid argument to a calculation function.
+void assertCalculationValue(Object arg, {bool checkUnquoted = true}) {
```
nex3, line 14: I'm updating the spec so that `CalculationOperation` will check for unquoted strings, which means you can get rid of `checkUnquoted` here _and_ the `CalculationOperation` check below.

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,134 @@
...
+        inMinMax: false, simplify: false);
+  });
+
+  jsClass.defineMethods({
+    'equals': (CalculationOperation self, Object other) => self == other,
+    'hashCode': (CalculationOperation self) => self.hashCode,
+  });
+
+  jsClass.defineGetters({
+    'operator': (CalculationOperation self) => self.operator.operator,
```
nex3, line 112: Shouldn't this expose `left` and `right` as well?

lib/src/node/value/calculation.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  getJSClass(SassCalculation.operateInternal(
+          CalculationOperator.plus, SassNumber(1), SassNumber(1),
+          inMinMax: false, simplify: false))
+      .injectSuperclass(jsClass);
+  return jsClass;
+}();
+
+/// The JavaScript CalculationInterpolation class
+final JSClass calculationInterpolationClass = () {
```
nex3, line 123: Shouldn't this expose `value` as a getter?

nex3 (changes requested):

lib/src/node/compile.dart
```diff
@@ -225,6 +225,32 @@ Importer _parseImporter(Object? importer) {
...
+/// Implements the simplification algorithm for custom function return values.
+/// {@link https://github.com/sass/sass/blob/main/spec/types/calculation.md#simplifying-a-calculationvalue}
+dynamic simplify(dynamic value) {
+  if (value is SassCalculation) {
+    var simplifiedArgs = value.arguments.map(simplify).toList().cast<Object>();
+    if (value.name == 'calc') {
+      return simplifiedArgs[0];
+    }
+    if (value.name == 'clamp') {
+      return simplify(Function.apply(SassCalculation.clamp, simplifiedArgs));
```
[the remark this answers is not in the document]
jerivas: This and the other review comments are gold! Really appreciate the time you're taking to make the code more idiomatic

lib/src/node/value/calculation.dart
jerivas: It seems like normal class properties are transparently available in JS (the tests verify this). I can implement them as getters manually if you prefer to call them out explicitly

lib/src/node/value/calculation.dart
jerivas: Same as above, it is available for free

lib/src/node/compile.dart
```diff
@@ -227,29 +227,25 @@ Importer _parseImporter(Object? importer) {
...
-  }
-  return value;
-}
+Object simplify(Object value) => switch (value) {
+      SassCalculation() => switch ((
+          // Match against...
+          value.name, // ...the calculation name
+          value.arguments.map(simplify).toList() // ...and simplified arguments
+        )) {
+          ('calc', [var first, ...]) => first,
```
nex3, line 236: As with `clamp`, this should be

```suggestion
          ('calc', [var first]) => first,
```

and then have a fallback case that throws an error for more arguments.

lib/src/node/compile.dart
```diff
@@ -227,29 +227,25 @@ Importer _parseImporter(Object? importer) {
...
+          ('clamp', [var min, var value, var max]) =>
+            SassCalculation.clamp(min, value, max),
+          ('clamp', _) =>
+            throw ArgumentError('clamp() requires exactly 3 arguments.'),
+          ('min', var args) => SassCalculation.min(args),
+          ('max', var args) => SassCalculation.max(args),
+          (var name, var args) => SassCalculation.unsimplified(name, args)
```
nex3, line 243: I think this should throw an error. We don't allow arbitrary calculation names.

lib/src/node/value/calculation.dart
```diff
@@ -62,11 +53,14 @@ final JSClass calculationClass = () {
       return SassCalculation.unsimplified('max', argList);
     },
     'clamp': (Object min, [Object? value, Object? max]) {
-      if (value == null) {
-        assertUnquotedStringOrInterpolation(min);
-      } else if (max == null) {
-        assertUnquotedStringOrInterpolation(min);
-        assertUnquotedStringOrInterpolation(value);
+      if ((value == null && !isValidClampArg(min)) ||
+          (max == null) && !([min, value]).any(isValidClampArg)) {
```
nex3, line 57: I think the parens are slightly wrong here:

```suggestion
          (max == null && ![min, value].any(isValidClampArg))) {
```

lib/src/node/value/calculation.dart
nex3: I think it's a safer bet to be explicit, yeah. I don't think dart2js provides strong stability guarantees for the JS API it generates.

lib/src/node/value/calculation.dart
```diff
@@ -11,33 +11,24 @@ import '../../value.dart';
...
         'SassNumber, SassString, SassCalculation, CalculationOperation, '
         'CalculationInterpolation'));
   }
-  if (checkUnquoted && arg is SassString && arg.hasQuotes) {
-    jsThrow(JsError('Argument must be unquoted SassString'));
-  }
-  if (arg is CalculationOperation) {
-    assertCalculationValue(arg.left, checkUnquoted: checkUnquoted);
-    assertCalculationValue(arg.right, checkUnquoted: checkUnquoted);
+  if (arg is SassString && arg.hasQuotes) {
```
nex3, line 24: I've been trying out a more pattern-matching style lately:

```suggestion
  if (arg case SassString(hasQuotes: true)) {
```

It's a little longer here, but only because the repeated variable name was so short, and I think it's a bit easier to understand.

nex3 (changes requested):

lib/src/node/compile.dart
```diff
@@ -260,18 +263,24 @@ List<AsyncCallable> _parseFunctions(Object? functions, {bool asynch = false}) {
...
-        } else {
-          throw 'Invalid return value for custom function '
-              '"${callable.name}": $result is not a sass.Value.';
         }
+        if (result is Value) {
+          var simplified = simplify(result);
+          if (simplified is! Value) {
+            throw 'Custom function "${callable.name}" '
+                'returned an object that cannot be simplified to a '
+                'sass.Value.';
```
nex3, line 278: As it stands, I think this throw is unreachable. If you call `simplify(Value)` it will always return a `Value`. This suggests to me that we should actually have two different functions:

```dart
Value _simplifyValue(Value value);
Object _simplifyCalcArg(Object object);
```

which call each other recursively as necessary.

lib/src/node/compile.dart
```diff
@@ -225,6 +225,31 @@ Importer _parseImporter(Object? importer) {
   }
 }
 
+/// Implements the simplification algorithm for custom function return values.
+/// {@link https://github.com/sass/sass/blob/main/spec/types/calculation.md#simplifying-a-calculationvalue}
+Object simplify(Object value) => switch (value) {
```
nex3, line 230: This should be private (indicated by a leading `_`).

lib/src/node/value/calculation.dart
```diff
@@ -11,24 +11,23 @@ import '../../value.dart';
...
-
-/// Check that [arg] is an unquoted string or interpolation
-bool isValidClampArg(Object? arg) => ((arg is CalculationInterpolation) ||
-    (arg is SassString && !arg.hasQuotes));
+void assertCalculationValue(Object arg) => switch (arg) {
+      (SassNumber() ||
+            SassString(hasQuotes: false) ||
+            SassCalculation() ||
+            CalculationOperation() ||
+            CalculationInterpolation()) =>
```
nex3, line 19: Nit: you don't need parentheses here.

lib/src/node/value/calculation.dart
```diff
@@ -11,24 +11,23 @@ import '../../value.dart';
...
-  }
-  if (arg is SassString && arg.hasQuotes) {
-    jsThrow(JsError('Argument `$arg` must be unquoted SassString'));
-  }
-}
-
-/// Check that [arg] is an unquoted string or interpolation
-bool isValidClampArg(Object? arg) => ((arg is CalculationInterpolation) ||
-    (arg is SassString && !arg.hasQuotes));
+void assertCalculationValue(Object arg) => switch (arg) {
```
nex3, line 14: These helper functions should also be private.

lib/src/node/value/calculation.dart
```diff
@@ -54,7 +53,7 @@ final JSClass calculationClass = () {
     },
     'clamp': (Object min, [Object? value, Object? max]) {
       if ((value == null && !isValidClampArg(min)) ||
-          (max == null) && !([min, value]).any(isValidClampArg)) {
+          (max == null) && !([min, value].any(isValidClampArg))) {
```
nex3, line 56: The parentheses here still aren't right:

```suggestion
          (max == null && ![min, value].any(isValidClampArg))) {
```

You only want parens wher eit's necessary to clarify boolean operation order. There's no need for them around individual expressions that aren't boolean operations.

lib/src/value/calculation.dart
```diff
@@ -341,15 +342,23 @@ class CalculationOperation {
   ///
   /// This is either a [SassNumber], a [SassCalculation], an unquoted
   /// [SassString], a [CalculationOperation], or a [CalculationInterpolation].
-  final Object left;
+  Object get left {
+    return _left;
+  }
+
+  final Object _left;
```
nex3, line 349: Style nit: use expression-bodied methods when possible, especially when wrapping fields.

```suggestion
  Object get left => _left;
  final Object _left;
```

nex3 (changes requested):

nex3 (approved):

PR #1994, opened 2023-06-07
Fix deploy configuration

nex3:


.github/workflows/ci.yml
```diff
@@ -576,7 +576,7 @@ jobs:
       - run: dart pub get
 
       - name: Deploy
-        run: dart run grinder pkg-chocolatey-deploy
+        run: dart run grinder protobuf pkg-chocolatey-deploy
```
ntkme, line 579: I think we need to run `./.github/util/initialize`. Otherwise we'll be missing `buf`.

ntkme (changes requested):

jathak (approved):

PR #1997, opened 2023-06-08
Use the correct pub credential location for deploying sub-packages

nex3:


jathak (approved):

ntkme:
As we are looking to fix release issues, another issue happening right now is that the self-hosted mac arm64 running seems to be stuck again. This seems to happen every once a while when we cut new releases. Not sure if anything can be done.

nex3:
Yeah, there are two issues there. First, it seems like the long-running process that's supposed to connect to GitHub Actions just... stops. Second, my login password periodically stops working for unclear reasons and I need to contact support to get it fixed. I have very little visibility into either of these issues (and frankly not a lot of experience managing servers in general), so I'm not sure what can be done.

Issue #2008, opened 2023-06-09
Add a default export for Node.js backwards compatibility


nex3:
Node.js allows ESM modules to `import` CJS files as though they used a default export, which means there's some amount of existing ESM usage of the `sass` package via `import sass from 'sass'`. However, with the browser support changes that landed in 1.63.0, we started providing explicit ESM exports that only work with our desired syntax of `import * as sass from 'sass'`. We should:

- [x] Re-add a default export for backwards-compatibility but have it emit deprecation warnings (https://github.com/sass/dart-sass/pull/2009)
- In Dart Sass 2.0.0, drop this default export
  - [x] Dart Sass (https://github.com/sass/dart-sass/pull/2717)
  - [x] Node.js Embedded Host (https://github.com/sass/embedded-host-node/pull/414/)

PR #2009, opened 2023-06-09
Re-enable support for default exports

nex3:
See #2008


code-forger (approved):
👍 

We just stumbled on this issue, pleased to see there is already a fix in the works

PR #2013, opened 2023-06-11
Improve `sass --embedded` performance

### LINKED ISSUE #2004: `sass --embedded` race condition in deactivating isolates
There is a race condition due to the deactivation of isolation (free up a wire id) being an async callback.

In a normal flow this is what should happen.
1. Host send compilation request id=1
2. Compiler activate isolate id=1
3. Compiler send compilation response id=1
4. Compiler deactivate isolate id=1
5. Host send compilation request id=1
6. ...

However, in the test I discovered that because the deactivation happens after sending response on a event loop rather than immediately after, there is a race condition:

1. Host send compilation request id=1
2. Compiler activate isolate id=1
3. Compiler send compilation response id=1
4. Host send compilation request id=1
6. Compiler deactivate isolate id=1
7. ???

It is completely legit for host to reuse id=1 as soon as it receives the response. However, a new request with the same id may land before the hook to deactivate isolate finishes, which causes host side lockup because it will never get a response.

ntkme:
This PR improves the performance by doing the following changes.

- ~~The wire packets of Record type `(int, Uint8List)` are decoded as streaming.~~
- Each isolate channel and its dispatcher is persisted rather than recreated for each compilation.
- ~~The communication between main isolate uses Record type `(bool, Uint8List)`, therefore allow a boolean signal to be passed to deactivate the channel at the correct timing, to fix the race condition.~~
- Use Mailbox via ffi to provide synchronized communication between isolates.

Performance summary:

- For lots of tiny files
  - Single-threaded performance has been improved a lot and getting closer to old single isolate performance.
  - Multi-threaded performance beats all previous versions and even beats multi-processes.
- For high number of round trips (function calls, importers, etc)
  - It beats all previous versions in all running modes (single-threaded, multi-threaded or multi-processes).
- For large files like bootstrap
  - Almost no performance difference in all running modes as this test is CPU & I/O bounded. Multi-processes mode is fastest for any version because multi-threaded mode is bottlenecked on I/O.


ntkme:
More or less the same on 100x bootstrap test. 

However, it has about 5% improvement on 10000x tiny input test.

```
$ ruby bm.rb 10000 tiny.scss
sass-embedded	1.63.3
       user     system      total        real
1 thread   0.269173   0.227431   0.496604 (  2.293781)
2 threads  0.314830   0.282349   0.597179 (  1.481784)
3 threads  0.321951   0.329817   0.651768 (  1.131190)
4 threads  0.322707   0.376493   0.699200 (  0.961794)
5 threads  0.336276   0.448218   0.784494 (  0.929259)
6 threads  0.352964   0.529659   0.882623 (  0.939316)
7 threads  0.361448   0.568346   0.929794 (  0.947324)
8 threads  0.359855   0.593809   0.953664 (  0.937949)
9 threads  0.361307   0.587833   0.949140 (  0.936673)
1 compiler   0.277534   0.224289   0.501823 (  2.313627)
2 compilers  0.332037   0.306910   0.638947 (  1.555841)
3 compilers  0.348038   0.380420   0.728458 (  1.186149)
4 compilers  0.366459   0.480182   0.846641 (  1.050849)
5 compilers  0.406660   0.684456   1.091116 (  1.072650)
6 compilers  0.423023   0.765263   1.188286 (  1.094254)
7 compilers  0.431131   0.822768   1.253899 (  1.109062)
8 compilers  0.441975   0.884685   1.326660 (  1.104897)
9 compilers  0.444171   0.878887   1.323058 (  1.095092)

$ ruby bm.rb 10000 tiny.scss
sass-embedded	1.63.4-dev
       user     system      total        real
1 thread   0.258727   0.220546   0.479273 (  2.166353)
2 threads  0.316143   0.289076   0.605219 (  1.404291)
3 threads  0.320524   0.334269   0.654793 (  1.074253)
4 threads  0.322629   0.384850   0.707479 (  0.933074)
5 threads  0.331725   0.443616   0.775341 (  0.885327)
6 threads  0.349339   0.530988   0.880327 (  0.895577)
7 threads  0.358472   0.566946   0.925418 (  0.905533)
8 threads  0.353131   0.566190   0.919321 (  0.885103)
9 threads  0.358826   0.590530   0.949356 (  0.894480)
1 compiler   0.261235   0.213976   0.475211 (  2.161243)
2 compilers  0.326474   0.311382   0.637856 (  1.477508)
3 compilers  0.344202   0.375815   0.720017 (  1.123474)
4 compilers  0.368854   0.512808   0.881662 (  1.026562)
5 compilers  0.390013   0.633235   1.023248 (  1.012801)
6 compilers  0.416028   0.757909   1.173937 (  1.051294)
7 compilers  0.415711   0.804037   1.219748 (  1.056922)
8 compilers  0.419840   0.805798   1.225638 (  1.054973)
9 compilers  0.433085   0.821438   1.254523 (  1.052989)
```

```
$ ruby bm.rb 100 node_modules/bootstrap/scss/bootstrap.scss
sass-embedded	1.63.3
       user     system      total        real
1 thread   0.023727   0.019990   0.043717 ( 15.556963)
2 threads  0.019926   0.018016   0.037942 (  8.486535)
3 threads  0.018927   0.016742   0.035669 (  6.095970)
4 threads  0.020328   0.019960   0.040288 (  4.956011)
5 threads  0.020069   0.017474   0.037543 (  4.351237)
6 threads  0.019909   0.017632   0.037541 (  3.806167)
7 threads  0.017738   0.019674   0.037412 (  3.656391)
8 threads  0.020224   0.023835   0.044059 (  3.434102)
9 threads  0.034716   0.033350   0.068066 (  3.711418)
1 compiler   0.021641   0.019480   0.041121 ( 15.581757)
2 compilers  0.019360   0.023874   0.043234 (  8.219086)
3 compilers  0.018385   0.022074   0.040459 (  5.627711)
4 compilers  0.021455   0.028490   0.049945 (  4.304132)
5 compilers  0.020231   0.028030   0.048261 (  3.473951)
6 compilers  0.019264   0.025563   0.044827 (  2.820670)
7 compilers  0.021347   0.037436   0.058783 (  2.526336)
8 compilers  0.022316   0.040450   0.062766 (  2.234550)
9 compilers  0.034136   0.044984   0.079120 (  2.194601)

$ ruby bm.rb 100 node_modules/bootstrap/scss/bootstrap.scss
sass-embedded	1.63.4-dev
       user     system      total        real
1 thread   0.022353   0.019819   0.042172 ( 15.560986)
2 threads  0.020251   0.018085   0.038336 (  8.467656)
3 threads  0.018742   0.016116   0.034858 (  6.099678)
4 threads  0.019197   0.019472   0.038669 (  4.988630)
5 threads  0.018519   0.017958   0.036477 (  4.385172)
6 threads  0.019444   0.018109   0.037553 (  3.882296)
7 threads  0.018454   0.019287   0.037741 (  3.643622)
8 threads  0.019348   0.021418   0.040766 (  3.419871)
9 threads  0.033607   0.031562   0.065169 (  3.717373)
1 compiler   0.020352   0.019002   0.039354 ( 15.700850)
2 compilers  0.017561   0.023552   0.041113 (  8.254263)
3 compilers  0.017576   0.024063   0.041639 (  5.627922)
4 compilers  0.020734   0.027967   0.048701 (  4.312268)
5 compilers  0.019619   0.031230   0.050849 (  3.531205)
6 compilers  0.017892   0.029569   0.047461 (  2.820040)
7 compilers  0.021994   0.031977   0.053971 (  2.521595)
8 compilers  0.021029   0.036081   0.057110 (  2.222935)
9 compilers  0.038061   0.045175   0.083236 (  2.238084)
```

ntkme:
```
$ ruby bm.rb 50000 tiny.scss
sass-embedded	1.62.1
       user     system      total        real
1 thread   1.148843   1.060483   2.209326 (  9.792060)
10 threads  1.376341   1.636628   3.012969 (  7.427155)
10 compilers  1.922713   3.331672   5.254385 (  4.511086)

$ ruby bm.rb 50000 tiny.scss
sass-embedded	1.63.3
       user     system      total        real
1 thread   1.335180   1.143176   2.478356 ( 11.433008)
10 threads  1.851761   3.071678   4.923439 (  4.728727)
10 compilers  2.194051   4.142164   6.336215 (  5.224874)

$ ruby bm.rb 50000 tiny.scss
sass-embedded	1.63.4-dev
       user     system      total        real
1 thread   1.261559   1.041179   2.302738 ( 10.772167)
10 threads  1.792097   2.895756   4.687853 (  4.343920)
10 compilers  2.083639   3.742469   5.826108 (  4.952839)
```

ntkme:
```
$ ruby fn.rb 10000 10
sass-embedded	1.62.1
       user     system      total        real
1 thread   2.938312   0.980359   3.918671 ( 10.200864)
10 threads  3.085656   1.045484   4.131140 (  9.898954)
10 compilers  4.315805   5.599141   9.914946 (  7.788891)

$ ruby fn.rb 10000 10     
sass-embedded	1.63.3
       user     system      total        real
1 thread   2.798330   1.154743   3.953073 ( 11.849905)
10 threads  3.903682   4.670239   8.573921 (  7.125081)
10 compilers  4.505083   6.536008  11.041091 (  8.586000)

$ ruby fn.rb 10000 10                                                                                                       
sass-embedded	1.63.4-dev
       user     system      total        real
1 thread   2.667328   1.033751   3.701079 (  9.446387)
10 threads  3.812873   4.425946   8.238819 (  6.642242)
10 compilers  4.281066   5.672533   9.953599 (  7.533604)
```

nex3:
I would like to avoid using FFI for this until there's a Dart-Team-owned package available for it. I don't feel confident in my ability to own and maintain that code.

ntkme:
@mraleph I took your Mailbox implementation and made it work with the current dart-sass. However, we would like to see Dart Team to own the FFI part and ship it as a package that we can consume. Would you or someone from Dart SDK team be able to help us? Thanks in advance!

mraleph:
@ntkme yes, we are going to provide the package as agreed. I was planning to take a look at this in the next month.

ntkme:
@nex3 I will split this up into smaller PRs.

mraleph:
@ntkme we have published https://pub.dev/packages/native_synchronization

ntkme:
@nex3 Welcome back. This PR is now ready for review.

nex3:
Thanks for the ping. It may be a few days before I can get around to this, but it's on my radar.

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -18,34 +20,38 @@ import 'util/explicit_close_transformer.dart';
...
+  final Mailbox mailbox;
+
+  /// The compilationId.
+  int id = 0;
+
+  /// The PoolResource.
+  PoolResource resource;
+
+  _Lease(this.mailbox, this.id, this.resource);
+}
```
nex3, line 36: I'm not generally a fan of using mutable state to implicitly pass messages around a program. If we need to track information for a given child isolate, I'd rather keep that logic encapsulated in a class where mutations are done explicitly through method calls.

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -102,84 +106,67 @@ class IsolateDispatcher {
...
     var receivePort = ReceivePort();
-    var future = Isolate.spawn(_isolateMain, receivePort.sendPort);
-    _allIsolates.add(future);
+    var future =
+        Isolate.spawn(_isolateMain, (mailbox.asSendable, receivePort.sendPort));
+    _isolates.add(future);
     await future;
 
-    var channel = IsolateChannel<_InitialMessage?>.connectReceive(receivePort)
+    var channel = IsolateChannel<Uint8List?>.connectReceive(receivePort)
```
nex3, line 144: Do we still need a two-way channel here? Could we just use the `Mailbox` to handle all the message sends, including those that start new compilations?

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -102,84 +106,67 @@ class IsolateDispatcher {
...
+        _mailboxes.remove(lease.id);
+        _inactive.add(lease);
+        lease.resource.release();
+      }
+      _channel.sink.add(Uint8List.sublistView(message, 1));
+    }, onError: (Object error, StackTrace stackTrace) {
+      _handleError(error, stackTrace);
+    }, onDone: () {
+      try {
+        mailbox.put(Uint8List(0));
```
nex3, line 165: Why do we need to send an explicit "close" message when we've already received a close from the isolate?

nex3 (changes requested):

lib/src/embedded/dispatcher.dart
```diff
@@ -30,83 +30,65 @@ final _outboundRequestId = 0;
...
-            throw paramsError("VersionRequest must have compilation ID 0.");
-
           case InboundMessage_Message.compileRequest:
-            if (_compiling) {
-              throw paramsError(
-                  "A CompileRequest with compilation ID $_compilationId is "
-                  "already active.");
+            var request = message.compileRequest;
+            var response = _compile(request);
+            if (!_asyncError) {
```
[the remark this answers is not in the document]
ntkme: This is because we want to capture the `ProtocolError` thrown when parsing the host response. `_compile` has the issue where it will capture the `ProtocolError` and convert it to `SassException`. So this check is to prevent it from sending a `CompileFailure` after sending a `ProtocolError`.

I also thought about let `_compile` simply throw `ProtocolError` but that would require change to the non-embedded part of the code base, and introduce `ProtocolError` to the general error handling code.

lib/src/embedded/isolate_dispatcher.dart
ntkme: Unfortunately we cannot use mailbox. Mailbox currently only support one message at a time therefore we cannot have multiple isolate sending message to a single mailbox.

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -70,13 +72,15 @@ class IsolateDispatcher {
...
-              .add(messageBuffer);
-          return;
+          var mailbox =
+              (_mailboxes[compilationId] ?? await _getMailbox(compilationId));
+          try {
+            mailbox.put(packet);
+            return;
+          } on StateError catch (_) {
+            throw paramsError(
+                "InboundMessage with compilation ID $compilationId is out of order.");
```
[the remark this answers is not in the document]
ntkme: This is a limit of mailbox can only hold one message at a time, meaning any speculative message send from the compiler will cause issue.

Imagine the following sequence:
1. Host send compile request
2. Compiler send function call request
3. Host send function call response
4. Compiler send compiler response

If the mailbox is a synchronized queue, then it will be possible to write a test where:

1. Host send compile request
2. Host send function call response (buffered in the queue)
3. Compiler send function call request (response consumed from queue)
4. Compiler send compiler response

This kind of speculative messaging is not going to work due to mailbox is limited to single message.

lib/src/embedded/isolate_dispatcher.dart
ntkme: To reuse the isolate in the most efficient way, the logic is basically:

``` dart
while {
  mailbox.take();
  ...
} do (!_asyncError);
```

The mutex basically puts the thread to into sleep state until it receives a new message and wakes up. However, the mutex while waiting is in FFI calling native code, and therefore it is not managed by dart lifecycle until re-entering a Dart code frame back from native code. This means while the thread is sleeping, `isolate.kill()` has no effect until the thread wakes up and come back to dart code frame. Therefore, we need to send something to wake up the thread to make sure it will be gracefully terminated on next dart frame.

lib/src/embedded/dispatcher.dart
```diff
@@ -30,83 +30,65 @@ final _outboundRequestId = 0;
...
+
+  /// The sink for sending messages to the host.
+  final StreamSink<Uint8List> _sink;
 
   /// The compilation ID for which this dispatcher is running.
   ///
   /// This is added to outgoing messages but is _not_ parsed from incoming
   /// messages, since that's already handled by the [IsolateDispatcher].
-  final int _compilationId;
+  int _compilationId = 0;
```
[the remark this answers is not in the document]
ntkme: Done.

lib/src/embedded/dispatcher.dart
```diff
@@ -30,83 +30,65 @@ final _outboundRequestId = 0;
...
-  final StreamChannel<Uint8List> _channel;
+  /// The mailbox for receiving messages from the host.
+  final Mailbox _mailbox;
+
+  /// The sink for sending messages to the host.
+  final StreamSink<Uint8List> _sink;
 
   /// The compilation ID for which this dispatcher is running.
   ///
   /// This is added to outgoing messages but is _not_ parsed from incoming
```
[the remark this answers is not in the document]
ntkme: Updated.

lib/src/embedded/dispatcher.dart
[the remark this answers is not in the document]
ntkme: We can try let ProtocolError throw all the way through here: https://github.com/sass/dart-sass/blob/58cbab42e63500f4269bc3f5ed01f7ee311cc4d3/lib/src/visitor/evaluate.dart#L1668-L1675

However, this will introduce ProtocolError to the JS side of the code base, which I'm not sure how to handle in a clean way. Another option is to capture ProtocolError in the `HostCallable` and re-throw it as a different error type e.g. FatalError, and let that particular error fall-through the catch statement all the way.

lib/src/embedded/isolate_dispatcher.dart
ntkme: Here is why we need a mutable wrapper to hold reference:

``` dart
  Future<Mailbox> _getMailbox(int compilationId) async {
    // Acquire or create a wrapper object
    var lease = ...;
    ... 
    // Save new compilationId into wrapper object
    lease.id = compilationId;
    ...
    // Because the isolate and channel objects are persisted, callback is only registered once
    channel.stream.listen((message) {
      ... 
      if (message[0] == 1) {
        // Because callback is only registered once, we cannot use compilationId here as it always reference to the first compilationId any persist isolate has handled. 
        // Therefore we need a mutable wrapper that hold a reference to the "current" compilationId 
        _mailboxes.remove(lease.id);
        _inactive.add(lease);
        // Same for the PoolResource object, we need to hold a reference to the "current" one
        lease.resource.release();
      }
      ...
    }
```

I understand having a mutable wrapper is not ideal, however, this is the most performant options among the alternatives:
1. Having extra lookup maps of `Mailbox -> compilationId` and `compilationId -> PoolResource` - map lookup is slower than dereference.
2. Deserialize the varint at the beginning of the outbound message to get the compilationId. This would be already slower than dereference, and would still require an extra map to track `PoolResource`.

ntkme: I added method `void transfer(int, PoolResource)` to the `_Lease` wrapper to hide the mutation of property.

@nex3 I'm open to other suggestion if you have better idea.

lib/src/embedded/dispatcher.dart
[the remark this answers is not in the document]
ntkme: I just pushed a new commit implementing the idea described above.

lib/src/embedded/isolate_dispatcher.dart
nex3: I've pushed a patch that refactors parts of the lease concept out into a `ReusableIsolate` class that encapsulates all the necessary mutability.

nex3 (changes requested):

lib/src/embedded/isolate_dispatcher.dart
ntkme: The reason I kept `IsolateChannel` to have two way connection is that `ExplicitCloseTransformer` was super handy for cleaning up the sink on either side, so that cleaning up on crash is easier. This can definitely be just a one-way stream.

lib/src/embedded/isolate_dispatcher.dart
ntkme: > although it seems like sending the empty message in the same place as calling isolate.kill() might be clearer

You're right, that we can just send an empty message to break from the loop and the isolate will stop on its own.

lib/src/embedded/dispatcher.dart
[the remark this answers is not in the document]
ntkme: Reverted the commit and update the name and documentation accordingly.

lib/src/embedded/isolate_dispatcher.dart
ntkme: Actually I see that you have already refactored this part. And it looks good to me.

ntkme:
@nex3 Huge thanks for refactoring the reuse logic and remove some left-overs. I've reverted the commit of bubbling up ProtocolError back to original implementation and updated documentation accordingly.

nex3 (approved):

nex3:
Thanks!

PR #2018, opened 2023-06-15
Fix a race condition with re-used compilation isolate IDs

### LINKED ISSUE #2004: `sass --embedded` race condition in deactivating isolates

nex3:
Closes #2004

### DIFF
--- a/lib/src/embedded/dispatcher.dart
+++ b/lib/src/embedded/dispatcher.dart
@@ -322,10 +322,17 @@ class Dispatcher {
     var protobufWriter = CodedBufferWriter();
     message.writeToCodedBufferWriter(protobufWriter);
 
-    var packet =
-        Uint8List(_compilationIdVarint.length + protobufWriter.lengthInBytes);
-    packet.setAll(0, _compilationIdVarint);
-    protobufWriter.writeTo(packet, _compilationIdVarint.length);
+    // Add one additional byte to the beginning to indicate whether or not the
+    // compilation is finished, so the [IsolateDispatcher] knows whether to
+    // treat this isolate as inactive.
+    var packet = Uint8List(
+        1 + _compilationIdVarint.length + protobufWriter.lengthInBytes);
+    packet[0] =
+        message.whichMessage() == OutboundMessage_Message.compileResponse
+            ? 1
+            : 0;
+    packet.setAll(1, _compilationIdVarint);
+    protobufWriter.writeTo(packet, 1 + _compilationIdVarint.length);
     _channel.sink.add(packet);
   }
 }
--- a/lib/src/embedded/isolate_dispatcher.dart
+++ b/lib/src/embedded/isolate_dispatcher.dart
@@ -156,19 +156,28 @@ class IsolateDispatcher {
     var receivePort = ReceivePort();
     isolate.sink.add((receivePort.sendPort, compilationId));
 
-    var channel = IsolateChannel<Uint8List?>.connectReceive(receivePort)
-        .transform(const ExplicitCloseTransformer());
-    channel.stream.listen(_channel.sink.add,
+    var channel = IsolateChannel<Uint8List>.connectReceive(receivePort);
+    channel.stream.listen(
+        (message) {
+          _channel.sink.add(Uint8List.sublistView(message, 1));
+
+          // The first byte of messages from isolates indicates whether the
+          // entire compilation is finished. Sending this as part of the message
+          // buffer rather than a separate message avoids a race condition where
+          // the host might send a new compilation request with the same ID as
+          // one that just finished before the [IsolateDispatcher] receives word
+          // that the isolate with that ID is done. See sass/dart-sass#2004.
+          if (message[0] == 1) {
+            channel.sink.close();
+            _activeIsolates.remove(compilationId);
+            _inactiveIsolates.add(isolate);
+            resource.release();
+          }
+        },
         onError: (Object error, StackTrace stackTrace) =>
             _handleError(error, stackTrace),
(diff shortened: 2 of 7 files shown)


lib/src/embedded/dispatcher.dart
```diff
@@ -322,10 +322,17 @@ class Dispatcher {
     var protobufWriter = CodedBufferWriter();
     message.writeToCodedBufferWriter(protobufWriter);
 
-    var packet =
-        Uint8List(_compilationIdVarint.length + protobufWriter.lengthInBytes);
-    packet.setAll(0, _compilationIdVarint);
-    protobufWriter.writeTo(packet, _compilationIdVarint.length);
+    // Add one additional byte to the beginning to indicate whether or not the
```
Goodwine, line 325: It feels error-prone to push a byte into the bytearray as a completion signal for the Isolate because then the Isolate has to remember about it and remove it before sending it to the host.

Would it make sense to wrap the bytearray with another object like a custom class or a tuple to make this signal more obvious? This would require changing the `IsolateDispatcher`'s input channel to `IsolateChannel<MyWrapper>`, but the output channel would remain the same 🤔

ntkme: Dart 3 Record would be a good fit for this. E.g. `IsolateChannel<(int, Uint8List)>`

nex3: I considered a record-based approach, but I decided against it for a couple reasons. First, I don't like the asymmetry between sending and receiving, especially given that the `StreamChannel` API requires symmetrical types. Passing an `int` that doubles as the compilation ID is a clever way around that, but ultimately moving the compilation ID out of the initial message doesn't really buy us anything—we've already got a pretty good way of providing the ID to the dispatcher isolate that doesn't require it to handle cases where it unexpectedly changes over time.

There's also the question of efficiency. I haven't measured, but my assumption is that (even as a value type) instantiating and passing records is going to be somewhat more overhead than expanding the binary buffer that already exists.

Ultimately, I'm not too worried about requiring the isolate dispatcher to narrow the buffer. Two classes that have a tightly coupled protocol is a relatively common pattern in message-passing scenarios like this.

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -156,19 +156,28 @@ class IsolateDispatcher {
     var receivePort = ReceivePort();
     isolate.sink.add((receivePort.sendPort, compilationId));
 
-    var channel = IsolateChannel<Uint8List?>.connectReceive(receivePort)
-        .transform(const ExplicitCloseTransformer());
-    channel.stream.listen(_channel.sink.add,
+    var channel = IsolateChannel<Uint8List>.connectReceive(receivePort);
+    channel.stream.listen(
+        (message) {
+          _channel.sink.add(Uint8List.sublistView(message, 1));
```
Goodwine, line 162: If we're keeping the first byte as a special thing between the isolate and the main thread, then I think it would be helpful to add a comment immediately before this line about why we're sending a slice of the bytearray.

lib/src/embedded/dispatcher.dart
Goodwine: I agree with the asymmetry concern, and with being OK with the isolate being tightly coupled (my experience working with isolates is pretty limited, but that sounds about right).

I wouldn't think the overhead would be that big, but I also didn't think appending values to a `List<T>` was expensive, and it was. 😝 

While I still think the pattern can be error prone, we do have tests which I think would catch a case where we forget to trim off the first byte.. I guess this is fine

Goodwine (approved):

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -158,19 +158,27 @@ class IsolateDispatcher {
...
+    var channel = IsolateChannel<Uint8List>.connectReceive(receivePort);
+    channel.stream.listen(
+        (message) {
+          // The first byte of messages from isolates indicates whether the
+          // entire compilation is finished. Sending this as part of the message
+          // buffer rather than a separate message avoids a race condition where
+          // the host might send a new compilation request with the same ID as
+          // one that just finished before the [IsolateDispatcher] receives word
+          // that the isolate with that ID is done. See sass/dart-sass#2004.
+          _channel.sink.add(Uint8List.sublistView(message, 1));
```
ntkme, line 170: To be extra safe, this should come after the if branch below. In other words, after isolate is deactivated / resource is released.

ntkme (changes requested):

lib/src/embedded/isolate_dispatcher.dart
nex3: This has to come _before_ `_channel.sink.close()`, or the call to `add()` will fail. I think it's safe enough to rely on Dart's single-threading guarantees here.

PR #2027, opened 2023-06-22
Use a common error handler for embedded protocol errors

ntkme:
This PR refactors out some duplicated code around error handling.

### DIFF
--- a/lib/src/embedded/utils.dart
+++ b/lib/src/embedded/utils.dart
@@ -2,10 +2,12 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import 'dart:io';
 import 'dart:typed_data';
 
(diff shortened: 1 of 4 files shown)


lib/src/embedded/utils.dart
```diff
@@ -134,3 +136,26 @@ final _compilationIdBuilder = VarintBuilder(32, 'compilation ID');
     _compilationIdBuilder.reset();
   }
 }
+
+/// Wraps error object into ProtocolError, writes it to stderr, and sends it to errorHandler function.
+void handleError(Object error, StackTrace stackTrace,
+    void Function(ProtocolError) errorHandler,
```
nex3, line 142: Could we just return the `ProtocolError` instead of passing it to a callback?

nex3 (approved):

nex3 (approved):

Issue #2023, opened 2023-06-20
If parsedAsCustomProperty is true, value must contain a SassString


sysmat:
- When using "@picocss/pico": "^1.5.10" sass I get error
- "sass": "^1.63.4", it works with previous version 
```scss
$primary-500: #E65A00;
$primary-600: #C54100;
$primary-700: #A52D00;

// Import full Pico source code
@import "./node_modules/@picocss/pico/scss/pico.scss";
```
- error
```
Unexpected exception:
Invalid argument(s): If parsedAsCustomProperty is true, value must contain a SassString (was `"Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"` of type SassList).


node_modules\sass\sass.dart.js 1267:17   Object.wrapException
node_modules\sass\sass.dart.js 1281:15   Object.throwExpression
node_modules\sass\sass.dart.js 7919:15   Object.ModifiableCssDeclaration$
node_modules\sass\sass.dart.js 64390:70  _EvaluateVisitor.visitCssDeclaration$1
node_modules\sass\sass.dart.js 37073:22  ModifiableCssDeclaration.accept$1$1
node_modules\sass\sass.dart.js 37076:19  ModifiableCssDeclaration.accept$1
node_modules\sass\sass.dart.js 66207:40  _EvaluateVisitor_visitCssStyleRule__closure.call$0
node_modules\sass\sass.dart.js 64606:25  _EvaluateVisitor._withStyleRule$1$2
node_modules\sass\sass.dart.js 64611:19  _EvaluateVisitor._withStyleRule$2
node_modules\sass\sass.dart.js 66198:10  _EvaluateVisitor_visitCssStyleRule_closure.call$0
```

Goodwine:
I'm unable to repro with the provided info. Can you provide a small repro repository and instructions on how to reproduce this?

What I tried:
```sh
cd $(mktemp --dir)
npm init # accept everything as default
npm install sass @picocss/pico ts-node
```
Setup my `package.json` scripts to:
```json
"scripts": {
  "test": "ts-node index.ts",
  "sass": "sass index.scss"
}
```
I used your input as `index.scss` and created an `index.ts` file with:
```ts
import * as sass from "sass";
const css = sass.compile("./index.scss");
console.log(css.css);
```
Then ran:
```sh
npm run test # compiles with JS API
npm run sass # compiles with CLI
```
Both of those commands compiled everything without any issue.
[...]

PR #2036, opened 2023-07-12
Do not validate a CSS declaration if its key is interpolated

Goodwine:
Specs: https://github.com/sass/sass-spec/pull/1923
Fixes: https://github.com/sass/dart-sass/issues/2023


nex3 (approved):
Please add a changleog entry. Also, the commit name should use "its" (possessive) instead of "it's" (contraction of "it is"). (Yes it's goofy that "'s" is possessive everywhere else in English but not here, I'm sorry about this language.)

PR #2049, opened 2023-07-31
Deprecate explicitly passing null as an alpha value

nex3:
This allows us to reserve null to indicate a missing alpha channel
instead in future versions.

See sass/sass#2831


pkg/sass_api/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 
 * All uses of classes from the `tuple` package have been replaced by record
   types.
+  
+## 7.2.2
+
+* No user-visible changes.
```
jathak, line 13: Is this still necessary?

jathak (approved):

pkg/sass_api/CHANGELOG.md
nex3: Yes, it got accidentally deleted in a previous CL but we did release 7.2.2 so it should have a changelog entry.

PR #2051, opened 2023-08-01
Track missing alpha channels in colors

nex3:
See sass/sass#2831


lib/src/value/color.dart
```diff
@@ -89,6 +89,8 @@ class SassColor extends Value {
   /// indicates a [missing] component.
   ///
   /// [missing]: https://www.w3.org/TR/css-color-4/#missing
+  ///
+  /// @nodoc
```
Goodwine, line 93: I'm curious why the `@nodoc` was added

Goodwine (commented):
I'm guessing no CHANGELOG was added because this is a feature branch, but should there be tests anywhere for these?

lib/src/value/color.dart
nex3: All of the channel-specific getters are intended to be package-private, I just forgot to add `@nodoc` earlier.

nex3:
> I'm guessing no CHANGELOG was added because this is a feature branch, but should there be tests anywhere for these?

We're working on tests separately in the feature branch of the sass-spec repo, but because the tests are fairly holistic we're getting the implementation mostly in-place first.

Goodwine (approved):

PR #2055, opened 2023-08-08
Use wrapJSExceptions() to work around dart-lang/sdk#53105

nex3:


pubspec.yaml
```diff
@@ -38,7 +39,6 @@ dependencies:
 dev_dependencies:
   analyzer: ">=5.13.0 <7.0.0"
   archive: ^3.1.2
-  cli_pkg: ^2.4.4
   crypto: ^3.0.0
   dart_style: ^2.0.0
   dartdoc: ^6.0.0
```
Goodwine, line 44: Looks like you need to update dartdoc too 🤔, otherwise LGTM

Issue #1934, opened 2023-04-13
Support additional calculation functions


nex3:
- [x] Original Implementation (https://github.com/sass/dart-sass/pull/1970)
- [x] Revert (https://github.com/sass/dart-sass/pull/2067)
- [ ] Roll Forward (https://github.com/sass/dart-sass/pull/2076)
- [ ] Breaking Changes

See https://github.com/sass/sass/issues/3504

stof:
@nex3 step 3 is now done here. And weren't breaking changes removed from the last version of the proposal ?

nex3:
There's still the final breaking change for `abs-percent`, which will need to happen in Dart Sass 2.

PR #2067, opened 2023-08-17
Revert new calculation functions

nex3:
See #1934
See sass/sass#3504
See sass/sass-spec#1931


stof:
shouldn't this keep the new CSS functions that don't conflict with Sass functions ? Many Sass math functions are only available namespaced.

nex3:
The issue isn't built-in Sass functions, it's user-defined Sass functions. Any user could define a function named `rem()` (to use the most common example) which would get completely blown up by this change.

PR #2068, opened 2023-08-17
Deprecate user-defined functions with math function names

nex3:
See #1934
See sass/sass#3504
See sass/sass-spec#1932

### DIFF
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -870,6 +870,37 @@ abstract class StylesheetParser extends Parser {
             "not" ||
             "clamp") {
       error("Invalid function name.", scanner.spanFrom(start));
+    } else if (name
+        case "round" ||
+            "mod" ||
+            "rem" ||
+            "sin" ||
+            "cos" ||
+            "tan" ||
+            "asin" ||
+            "acos" ||
+            "atan" ||
+            "atan2" ||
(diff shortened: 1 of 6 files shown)


Goodwine (approved):

lib/src/parse/stylesheet.dart
```diff
@@ -870,6 +870,37 @@ abstract class StylesheetParser extends Parser {
...
+        _ => 'Either rename it or use the math.$name() function.'
+      };
+
+      logger.warnForDeprecation(
+          Deprecation.mathFunctionName,
+          'Naming a Sass function "$name" is deprecated because it conflicts '
+          'with the new\n'
+          'CSS $name() syntax. $suggestion\n'
+          '\n'
+          'More info: https:/sass-lang.com/d/math-fn-name',
```
Goodwine, line 902: The only nit is that this page is not live yet, but I understand it should go live soon

Goodwine: Oh actually on second look, there is a missing `/`
```suggestion
          'More info: https://sass-lang.com/d/math-fn-name',
```

Goodwine (changes requested):

nex3:
Closing this out in favor of the approach described in https://github.com/sass/sass/pull/3661

Issue #1978, opened 2023-05-26
Add Dart 3 class modifiers to public classes


nex3:
Pretty much all the classes we expose publicly in the Dart API should have some form of [class modifier](https://dart.dev/language/class-modifiers), but adding this would be a breaking change, so we should wait until 2.0.0.

nex3:
I think it's fine to convert classes that have the `@sealed` annotation to `sealed` or `final`, but we shouldn't do so for un-annotated classes and we can't safely mark classes as `base` or `interface`.

PR #2073, opened 2023-08-25
implement first class mixins

### LINKED ISSUE #2046: Implement first-class mixins
See https://github.com/sass/sass/issues/626

connorskees:
Tracking issue: https://github.com/sass/sass/issues/626
Protobuf changes: https://github.com/sass/sass/pull/3674
Spec tests: sass/sass-spec#1933
Embedded host: https://github.com/sass/embedded-host-node/pull/253

Closes #2046


lib/src/callable/built_in.dart
```diff
@@ -21,6 +21,8 @@ final class BuiltInCallable implements Callable, AsyncBuiltInCallable {
   /// The overloads declared for this callable.
   final List<(ArgumentDeclaration, Callback)> _overloads;
 
+  bool hasContent = false;
```
connorskees, line 24: Not super happy with adding this property on the main `BuiltInCallable` class, but it is necessary in `meta.accepts-content` to know whether the builtin mixin callable hasContent because of `meta.apply`, which does accept content in some scenarios.

lib/src/functions/meta.dart
```diff
@@ -77,6 +79,18 @@ final local = UnmodifiableListView([
...
+  }),
+  _function("accepts-content", r"$mixin", (arguments) {
+    var mixin = arguments[0].assertMixin("mixin");
+    bool acceptsContent = switch (mixin.callable) {
+      AsyncBuiltInCallable(hasContent: var hasContent) ||
+      BuiltInCallable(hasContent: var hasContent) ||
+      UserDefinedCallable(declaration: MixinRule(hasContent: var hasContent)) =>
+        hasContent,
+      AsyncCallable() =>
+        throw UnsupportedError("Unknown callable type $mixin."),
```
connorskees, line 91: Unsure what should happen here. AFAICT this class has been replaced by `AsyncBuiltInCallable` and is never itself actually used anywhere?

connorskees:
I'm not able to reproduce the CI build failures locally. I assume it's to do with missing auto-generated (and gitignored) files that I have locally but which aren't created during the CI run. Is there something I'm missing updating for the embedded protocol files?

ntkme:
@connorskees If you are making changes to the protobuf definition, you need to submit a PR for https://github.com/sass/sass, and link the PR in the description of this PR.

lib/src/callable/built_in.dart
connorskees: I suppose one alternative is to have `meta.apply` not actually return true for `meta.accepts-content`.

connorskees:
I've reverted the embedded protocol changes for now as adding the protobuf changes to `sass/sass` and linking the issues didn't seem to resolve the build failures.

nex3:
It looks like the CI _is_ picking up the protocol buffer definition changes (at least now), so you may want to roll forward that patch.

lib/src/callable/built_in.dart
nex3: From a strict reading of the spec, the question `meta.accepts-content` asks is "does the body of this mixin have a `@content` rule?" The body of `meta.apply` doesn't exist, so the answer is "no". I think it's safe to just return "no" here as well. (It's also the case that _many_ invocations of `meta.apply` won't actually accept a content block in practice.)

lib/src/functions/meta.dart
nex3: Throwing an `UnsupportedError` is correct. Eventually hopefully https://github.com/dart-lang/language/issues/3113 will be addressed and we'll be able to explicitly declare to the language that the set of callables described above are exhaustive.

lib/src/value.dart
```diff
@@ -177,6 +179,13 @@ abstract class Value {
   SassFunction assertFunction([String? name]) =>
       throw SassScriptException("$this is not a function reference.", name);
 
+  /// Throws a [SassScriptException] if [this] isn't a mixin reference.
+  ///
+  /// If this came from a mixin argument, [name] is the argument name
```
nex3, line 184: This should still be "function argument", since the Dart API doesn't support defining custom mixins.

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,70 @@ final class _EvaluateVisitor
...
+        var invocation = ArgumentInvocation([], {}, callableNode.span,
+            rest: ValueExpression(args, callableNode.span),
+            keywordRest: args.keywords.isEmpty
+                ? null
+                : ValueExpression(
+                    SassMap({
+                      for (var (name, value) in args.keywords.pairs)
+                        SassString(name, quotes: false): value
+                    }),
+                    callableNode.span));
```
nex3, line 557: You don't need to define this manually. The reason we store keywords in the primary rest argument by default is specifically so that you can pass it directly as a rest argument to another callable and have  keyword arguments just work. (This also means that `meta.call()` is doing this unnecessarily as well.)

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,70 @@ final class _EvaluateVisitor
...
+
+        var callable = mixin.assertMixin("mixin").callable;
+        // ignore: unnecessary_type_check
+        if (callable is AsyncCallable) {
```
nex3, line 561: You could fold this check into the `switch` statement below.

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,70 @@ final class _EvaluateVisitor
...
+          switch (callable) {
+            case BuiltInCallable() when _environment.content != null:
+              throw _exception("Mixin doesn't accept a content block.");
```
nex3, line 564: This should have `callableNode.span` as its span.

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,70 @@ final class _EvaluateVisitor
...
+                  callableNode.span,
+                  "invocation",
+                  {callable.declaration.arguments.spanWithName: "declaration"},
+                  _stackTrace(callableNode.span));
+
+            case UserDefinedCallable<AsyncEnvironment>():
+              var contentCallable = _environment.content.andThen((content) =>
+                  UserDefinedCallable(
+                      content as CallableDeclaration, _environment.closure(),
+                      inDependency: _inDependency));
```
nex3, line 584: `_environment.content` is already a `UserDefinedCallable`. Why not just use that as-is? In fact, do you need to redefine this at all, or will the existing `_environment.content` just naturally fall through to the child mixin?

nex3 (changes requested):

lib/src/embedded/mixin_registry.dart
nex3, line 1: Rather than implementing the same basic registry twice, it might be better to generalize this to `OpaqueValueRegistry` and have `protofy()` just return an ID that the caller uses to construct the actual protocol buffer.

lib/src/value/mixin.dart
```diff
@@ -0,0 +1,38 @@
...
+///
+/// {@category Value}
+@sealed
+class SassMixin extends Value {
+  /// The callable that this mixin invokes.
+  ///
+  /// Note that this is typed as an [AsyncCallable] so that it will work with
+  /// both synchronous and asynchronous evaluate visitors, but in practice the
+  /// synchronous evaluate visitor will crash if this isn't a [Callable].
+  final AsyncCallable callable;
```
nex3, line 24: Mark this `@internal` and `:nodoc:`, because there's no public Dart API for interacting with mixin definitions currently.

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,62 @@ final class _EvaluateVisitor
             configuration: configuration,
             namesInErrors: true);
         _assertConfigurationIsEmpty(configuration, nameInError: true);
-      }, url: "sass:meta")
+      }, url: "sass:meta"),
+      BuiltInCallable.mixin("apply", r"$mixin, $args...", (arguments) async {
```
nex3, line 543: Style nit: empty line above

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,62 @@ final class _EvaluateVisitor
...
+        var mixin = arguments[0];
+        var args = arguments[1] as SassArgumentList;
+
+        var callableNode = _callableNode!;
+        var invocation = ArgumentInvocation(
+          [],
+          {},
```
nex3, line 550: Mark these `const` since they're never expected to mutate.

lib/src/functions/meta.dart
```diff
@@ -77,6 +79,15 @@ final local = UnmodifiableListView([
...
         ListSeparator.comma);
+  }),
+  _function("accepts-content", r"$mixin", (arguments) {
+    var mixin = arguments[0].assertMixin("mixin");
+    return SassBoolean(switch (mixin.callable) {
+      AsyncBuiltInCallable() || BuiltInCallable() => false,
+      UserDefinedCallable(declaration: MixinRule(hasContent: var hasContent)) =>
+        hasContent,
+      _ => throw UnsupportedError("Unknown callable type $mixin.")
+    });
```
nex3, line 90: I've changed my mind on how this should behave with built-in mixins—I think it _should_ look for `allowsContent`, for two reasons:

1. This matches the behavior of user-defined mixins, where if _any_ branch references `@content` the entire mixin is considered to "have content".

2. If in the future we add any built-in mixins that _always_ take a content block, we'll definitely want to check `acceptsContent` here for them, and at that point it would be awkward if we had to add a special case or change the behavior for `meta.apply()`.

Could you also update the spec to check explicitly for built-in mixins that can take content blocks?

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,62 @@ final class _EvaluateVisitor
...
+          rest: ValueExpression(args, callableNode.span),
+        );
+
+        var callable = mixin.assertMixin("mixin").callable;
+        var content = _environment.content;
+
+        switch (callable) {
+          case AsyncBuiltInCallable(acceptsContent: false) when content != null:
+            throw _exception(
+                "Mixin doesn't accept a content block.", callableNode.span);
```
nex3, line 561: We should be able to use the same `MultiSpanSassRuntimeException` as the user-defined case, although you may need to make `BuiltInCallback.overloads` public to do so.

lib/src/callable/async_built_in.dart
```diff
@@ -26,6 +26,9 @@ class AsyncBuiltInCallable implements AsyncCallable {
   /// The callback to run when executing this callable.
   final Callback _callback;
 
+  /// Whether this callback could potentially accept an `@content` block.
```
nex3, line 29: "callback" -> "callable"

Add a note that this is only ever true for mixins.

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,62 @@ final class _EvaluateVisitor
...
+                "Mixin doesn't accept a content block.",
+                callableNode.span,
+                "invocation",
+                {callable.declaration.arguments.spanWithName: "declaration"},
+                _stackTrace(callableNode.span));
+
+          case UserDefinedCallable<AsyncEnvironment>():
+            _runUserDefinedCallable(invocation, callable, callableNode,
+                () async {
+              await _environment.withContent(content, () async {
```
nex3, line 582: Do you need to call this? Won't the current content just remain the same otherwise?

lib/src/visitor/async_evaluate.dart
```diff
@@ -512,7 +539,62 @@ final class _EvaluateVisitor
...
+                  }
+                });
+              });
+            });
+
+          case _:
+            throw SassScriptException(
+                "The mixin ${callable.name} is asynchronous.\n"
+                "This is probably caused by a bug in a Sass plugin.");
+        }
```
nex3, line 596: Can you factor this out into an `_applyMixin()` function that can be shared between this and `visitIncludeRule()`?

nex3 (changes requested):

lib/src/visitor/async_evaluate.dart
connorskees: I agree with this, but the dart formatter does not and immediately reverts the formatting change.

connorskees:
Most recent CI failure seems to be because the spec test PR didn't include https://github.com/sass/sass-spec/commit/3301bb5cde9ff1be5892bdce58cd191c6a0105af. This should now be resolved.

connorskees:
The current CI failure seems to be unrelated to the changes in this PR, as I'm seeing it on other PRs as well. CI looks green ignoring this workflow.

image

nex3:
> The current CI failure seems to be unrelated to the changes in this PR, as I'm seeing it on other PRs as well.

This is worked around on the current main branch.

lib/src/visitor/async_evaluate.dart
```diff
@@ -1733,37 +1785,50 @@ final class _EvaluateVisitor
     }
   }
 
-  Future<Value?> visitIncludeRule(IncludeRule node) async {
-    var nodeWithSpan = AstNode.fake(() => node.spanWithoutContent);
-    var mixin = _addExceptionSpan(node,
-        () => _environment.getMixin(node.name, namespace: node.namespace));
+  Future<void> _applyMixin(
```
nex3, line 1788: Nit: Document this.

nex3 (approved):
One more docstring needed, and merge main again to fix CI, but otherwise looks good!

lib/src/visitor/async_evaluate.dart
connorskees: `/// Evaluate a given [mixin] with [arguments] and [contentCallable]` Does this make sense, or would you want something more specific?

nex3: That's fine. Sometimes the role of a documentation comment is just to reassure the reader that yes, this does do what it looks like it does.

lib/src/embedded/opaque_registry.dart
```diff
@@ -0,0 +1,30 @@
...
+final class OpaqueRegistry<T> {
+  /// Instantiations of `T` that have been sent to the host.
+  ///
+  /// The values are located at indexes in the list matching their IDs.
+  final _elementsById = <T>[];
+
+  /// A reverse map from elements to their indexes in [_elementsById].
+  final _idsByElement = <T, int>{};
+
+  /// Converts an element of type `T` to a protocol buffer to send to the host.
```
stof, line 16: this converts it to an id only, not to a protocol buffer value.

lib/src/embedded/opaque_registry.dart
```diff
@@ -0,0 +1,30 @@
...
+  int protofy(T element) {
```
stof, line 17: The name `protofy` is not consistent with the other parts of the embedded compiler. AFAICT, all `protofy*` functions are returning a protocol buffer object, which is not the case here. It might be better to name it differently to avoid confusion.

lib/src/value/mixin.dart
```diff
@@ -0,0 +1,41 @@
...
+import '../visitor/interface/value.dart';
+import '../value.dart';
+
+/// A SassScript mixin reference.
+///
+/// A mixin reference captures a mixin from the local environment so that
+/// it may be passed between modules.
+///
+/// {@category Value}
+@sealed
```
stof, line 17: As this is a new class (and so there is no BC concern on it), should it directly use a Dart 3 modifier (refs #1978) or do you prefer to have them all or none ?

lib/src/visitor/async_evaluate.dart
```diff
@@ -1733,37 +1785,51 @@ final class _EvaluateVisitor
...
-    var nodeWithSpan = AstNode.fake(() => node.spanWithoutContent);
-    var mixin = _addExceptionSpan(node,
-        () => _environment.getMixin(node.name, namespace: node.namespace));
+  /// Evaluate a given [mixin] with [arguments] and [contentCallable]
+  Future<void> _applyMixin(
+      AsyncCallable? mixin,
+      UserDefinedCallable<AsyncEnvironment>? contentCallable,
+      ArgumentInvocation arguments,
+      FileSpan span,
+      FileSpan spanWithoutContent) async {
```
stof, line 1794: Should those spans really be provided as is or should a node be provided to keep them lazy in case there is no error ?

lib/src/value/mixin.dart
nex3: Good point, let's make this properly `final`.

lib/src/visitor/async_evaluate.dart
connorskees: `IncludeRule` isn't possible afaict because `meta.apply` doesn't have access to the rule including it. I made this change in https://github.com/sass/dart-sass/pull/2073/commits/f1c1d5af1cc1408b80fc240c3ba42077029d2e88 where I just use a regular `AstNode`.

connorskees:
Do I need to link to the embedded host PR in this PR description?

nex3:
Yes, you do.

connorskees:
Updated.

ntkme:
> Updated.

You will also need to push a different commit in order for CI to get the updated PR description. E.g. an empty created by `git commit --allow-empty` would work.

lib/src/js/value/mixin.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'package:node_interop/js.dart';
+
+import '../../callable.dart';
+import '../../value.dart';
+import '../reflection.dart';
+import '../utils.dart';
+
+/// The JavaScript `SassMixin` class.
+final JSClass mixinClass = () {
+  var jsClass = createJSClass('sass.SassMixin', () {
```
ntkme, line 14: Not sure if it is related to the browser failure, but I think the argument is missing `Object self` here when looking at all other values.

connorskees:
I added back the assertion to the sass-spec to see if adding that parameter fixes the browser failure.

lib/src/js/value/mixin.dart
```diff
@@ -0,0 +1,22 @@
...
+
+/// The JavaScript `SassMixin` class.
+final JSClass mixinClass = () {
+  var jsClass = createJSClass('sass.SassMixin', (Object self) {
+    jsThrow(JsError(
+        'It is not possible to construct a SassMixin through the JavaScript API'));
+  });
+
+  getJSClass(SassMixin(Callable('f', '', (_) => sassNull)))
+      .injectSuperclass(jsClass);
```
ntkme, line 20: Also this does not make much sense. If I understand it correctly the reason we have this for function is to allow the "JS" type to inherit the prototype of dart-converted-JS type? If so here the argument passed into SassMixin (the dart type) should be some argument necessary to construct an actual object.

@nex3 I don't fully understand how the JS-interop works so please correct me if I'm wrong.

ntkme: Never mind. Looks like Mixin just takes AsyncCallable so this is probably fine.

connorskees:
If possible, I would like for the first class mixin PRs to merge quite soon -- after this week I will have significantly less time to work outside of work.

nex3:
It's not clear to me why the `instanceof` check isn't working in a browser environment. Injecting the superclass _should_ make it respond to `instanceof`. You might need to load it up yourself in a browser and poke around there to see where the disconnect is between what _should_ be happening and what _is_.

connorskees:
Seems it might be a known jest issue? See https://github.com/jestjs/jest/issues/5946 and https://github.com/jestjs/jest/issues/2549 and similar issues on stackoverflow, https://stackoverflow.com/questions/52551035/tobeinstanceofnumber-does-not-work-in-jest

nex3:
Jest does have weird `instanceof` errors, but we don't use Jest in the browser. That also wouldn't explain why the `instanceof` check fails for `SassMixin` but succeeds for other `Value` subtypes.

ntkme:
Another difference I noticed is that the dart `SassFunction` type has `@sealed class SassFunction extends Value` but the dart `SassMixin` type has `final class SassMixin extends Value`. Not sure if this will affect the generated JS code, but I don't have other idea.

connorskees:
The most recent error is interesting: `Error: Right-hand side of 'instanceof' is not an object`. This would imply to me `SassMixin` imported from `'sass'` in the browser is undefined (or something else??).

ntkme:
@connorskees  I think I found the root cause: we're missing ESM export here:
https://github.com/sass/dart-sass/blob/16b85120f53bf3f26f77daa143beb0bcf3eabf90/tool/grind.dart#L49-L65

Node runs the CJS entrypoint but browser runs the ESM entrypoint...

connorskees:
Nice find! Thank you -- that would have been annoying to find otherwise.

nex3:
I wonder if it would be possible to add a conformance check for that...

nex3:
All right, looks like everything is passing but https://github.com/sass/embedded-host-node/pull/253. Once https://github.com/sass/embedded-host-node/pull/255 lands, I'll merge these all in and we should be good to go. Thanks for the hard work, @connorskees!

lib/src/visitor/async_evaluate.dart
```diff
@@ -418,6 +418,19 @@ final class _EvaluateVisitor
         });
       }, url: "sass:meta"),
 
+      BuiltInCallable.function("module-mixins", r"$module", (arguments) {
```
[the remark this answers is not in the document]
nex3: No, they aren't expected to be exposed globally.

PR #2074, opened 2023-08-26
Fix docker example

danog:
The order of the base stages is wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,8 +305,8 @@ commands:
 
 ```Dockerfile
 # Dart stage
-FROM dart:stable AS dart
 FROM bufbuild/buf AS buf
+FROM dart:stable AS dart
 
 # Add your scss files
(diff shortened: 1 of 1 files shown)


jathak:
It's not clear to me why this change is necessary. Can you share the error message you get when running the example as it currently exists?

danog:
```
[daniil@pi4 a]$ DOCKER_BUILDKIT=1 docker build .
[+] Building 0.5s (2/2) FINISHED            docker:default
 => [internal] load .dockerignore                     0.2s
 => => transferring context: 2B                       0.0s
 => [internal] load build definition from Dockerfile  0.2s
 => => transferring dockerfile: 445B                  0.0s
ERROR: failed to solve: circular dependency detected on stage: buf
```

jathak (approved):

PR #2078, opened 2023-09-02
Run cli compilations in parallel dart isolates

ntkme:
This PR enables parallel complications in isolates when running the pure dart version of sass cli.

```
$ sass --version
1.66.1
$ time sass bootstrap/scss:css
sass bootstrap/scss:css  0.39s user 0.06s system 99% cpu 0.450 total
```

```
$ sass --version
1.66.2-dev
$ time sass bootstrap/scss:css
sass bootstrap/scss:css  0.44s user 0.08s system 145% cpu 0.353 total
```

The goal of this PR is to improve the performance of https://github.com/rails/dartsass-rails, which invokes the cli directly.


lib/src/executable/concurrent/vm.dart
```diff
@@ -0,0 +1,30 @@
...
+
+/// Compiles the stylesheet at [source] to [destination].
+///
+/// Runs in a new Dart Isolate, unless [source] is `null`.
+Future<(int, String, String?)?> compileStylesheet(ExecutableOptions options,
+    StylesheetGraph graph, String? source, String? destination,
+    {bool ifModified = false}) async {
+  // Reading from stdin does not work properly in dart isolate.
+  if (source == null) {
+    return c.compileStylesheet(options, graph, source, destination,
```
nex3, line 21: Nit: use `return await` when returning a future from an `async` function, or else `try`/`catch` won't work correctly if you add one later.

lib/src/executable/concurrent/vm.dart
```diff
@@ -0,0 +1,30 @@
...
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'dart:isolate';
+
+import 'package:term_glyph/term_glyph.dart' as term_glyph;
+
+import '../options.dart';
+import '../../stylesheet_graph.dart';
+import 'js.dart' as c;
```
nex3, line 11: If this file is getting used from the VM, it shouldn't be called `js.dart`. Maybe `shared.dart`? Or could the logic here just be merged into `compile_stylesheet.dart` given that the watch function is calling out to the concurrent logic as well now?

lib/src/executable/watch.dart
```diff
@@ -41,15 +36,18 @@ Future<void> watch(ExecutableOptions options, StylesheetGraph graph) async {
   // they currently exist. This ensures that changes that come in update a
   // known-good state.
   var watcher = _Watcher(options, graph);
-  for (var (source, destination) in _sourcesToDestinations(options).pairs) {
+  var sourcesToDestinations = _sourcesToDestinations(options);
+  for (var (source, _) in sourcesToDestinations.pairs) {
```
nex3, line 40: If you're not using the destination, you can just iterate through `.keys`.

lib/src/executable/watch.dart
```diff
@@ -226,34 +182,41 @@ final class _Watcher {
...
     var allSucceeded = true;
-    while (toRecompile.isNotEmpty) {
-      var node = toRecompile.removeFirst();
-      if (!seen.add(node)) continue;
-
-      var success = await _compileIfEntrypoint(node.canonicalUrl);
-      allSucceeded = allSucceeded && success;
-      if (!success && _options.stopOnError) return false;
+    while (nodes.isNotEmpty) {
+      nodes = nodes.where((node) => seen.add(node)).toList();
```
nex3, line 187: It's more idiomatic these days to use list syntax for this sort of thing:

```suggestion
      nodes = [for (var node in nodes) if (seen.add(node)) node];
```

Also below.

lib/src/executable/watch.dart
```diff
@@ -226,34 +182,41 @@ final class _Watcher {
...
+  Iterable<(String, String)> _sourceEntrypointsToDestinationsPairs(
+      Iterable<StylesheetNode> nodes) {
+    var pairs = <(String, String)>[];
+    for (var node in nodes) {
+      var url = node.canonicalUrl;
+      if (url.scheme != 'file') continue;
+
+      var source = p.fromUri(url);
+      var destination = _destinationFor(source);
+      if (destination != null) {
```
nex3, line 215: In Dart 3, you can check for null and assign in a single statement:

```suggestion
      if (_destinationFor(source) case var destination?) {
```

test/cli/shared/update.dart
```diff
@@ -66,16 +66,19 @@ void sharedTests(Future<TestProcess> runSass(Iterable<String> arguments)) {
       await d.file("test2.scss", r"$var: 2; @import 'other'").create();
 
       var sass = await update(["test1.scss:out1.css", "test2.scss:out2.css"]);
-      expect(sass.stdout, emits(endsWith('Compiled test1.scss to out1.css.')));
-      expect(sass.stdout, emits(endsWith('Compiled test2.scss to out2.css.')));
```
nex3, line 70: If the issue here is that the order is now non-deterministic, you can use [`emitsInAnyOrder`](https://pub.dev/documentation/test/latest/expect/emitsInAnyOrder.html) to test for it without mandating a specific order.

nex3 (changes requested):

lib/src/executable/concurrent/vm.dart
ntkme: I made this a normal sync function so that we can avoid nested `return await`.

nex3 (approved):

nex3:
@ntkme The test failure here suggests that we may need to print all messages in the main isolate to ensure they don't end up stepping on one another.

ntkme:
@nex3 I think the issue is that the internal implementation of `print` uses `writeln`, which actually write the line and newline character in two separate calls. I think by adding the newline character to the buffer and then use `write` instead of `writeln`, it should be able to avoid this issue.

ntkme:
@nex3 I put up small hack that overrides `print` for dart vm. Obviously it's not the most elegant solution but supposedly should work. I think a cleaner solution would be to add newline directly to all `StringBuffer`s, and then replace all `print` with `stdout.write`.

lib/src/io/vm.dart
```diff
@@ -37,8 +37,24 @@ bool get supportsAnsiEscapes {
   return io.stdout.supportsAnsiEscapes;
 }
 
+void print(Object? message) {
```
nex3, line 40: Rather than overriding the default `print()` I think it's better to call this something like `safePrint()`.

nex3 (changes requested):
Would you mind filing an issue against the Dart SDK about `print()` not being thread-safe? It seems like the best way to fix that is probably in the SDK itself.

ntkme:
> Would you mind filing an issue against the Dart SDK about print() not being thread-safe? It seems like the best way to fix that is probably in the SDK itself.

https://github.com/dart-lang/sdk/issues/53471

nex3 (approved):

lib/src/io/vm.dart
```diff
@@ -37,8 +37,24 @@ bool get supportsAnsiEscapes {
...
 void printError(Object? message) {
-  io.stderr.writeln(message);
+  _threadSafeWriteLn(io.stderr, message);
+}
+
+void _threadSafeWriteLn(io.IOSink sink, Object? message) {
+  // For better performance the buffer is mutated.
+  if (message is StringBuffer) {
+    message.writeln();
+    sink.write(message);
```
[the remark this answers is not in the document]
ntkme: Made the change to eat the cost and just duplicate the buffer. I reviewed the places we should use thread-safe print and printError, we don't have any places with excessive payload, so it should be fine.

PR #2085, opened 2023-09-15
Fix changelog for 1.67.0

nex3:


jathak (approved):

stof:
The Deprecation enum actually also has it as a value.

nex3:
Oops, addressed in https://github.com/sass/dart-sass/pull/2096.

PR #2090, opened 2023-09-18
Cut a release

nex3:


ntkme:
@nex3 Sass api version must be bumped as well. There is a check for that in the double check step: https://github.com/sass/dart-sass/blob/13c9fb3c9d97d4d9582ccd33689111d74024261d/tool/grind/double_check.dart#L36-L47

nex3:
@ntkme Good catch, I was thinking we'd already bumped it in one of the prior CLs.

pkg/sass_api/pubspec.yaml
```diff
@@ -2,15 +2,15 @@ name: sass_api
...
+version: 9.1.0
 description: Additional APIs for Dart Sass.
 homepage: https://github.com/sass/dart-sass
 
 environment:
   sdk: ">=3.0.0 <4.0.0"
 
 dependencies:
-  sass: 1.67.0
+  sass: 1.67.1
```
ntkme, line 13: Now this should be 1.68.0?

ntkme:
Another missing item here is that we need to cut a release for embedded protocol and update the dependency version of embedded protocol.

nex3:
Thanks @ntkme! Included the protocol version bump in https://github.com/sass/sass/pull/3693.

jathak (approved):

PR #2092, opened 2023-09-20
Fix an error during embedded compiler shutdown

ntkme:
@nex3 This PR fixes the error in this failed CI run: https://github.com/sass/dart-sass/actions/runs/6254921481/job/16983359566?pr=2090


nex3 (approved):

ntkme:
Looks like 1.68.0 release CI run failed because of this issue. 😅 

I guess we can just try re-run the failed job and see if it works since it's a timing issue that does not always occur.

nex3:
I'm happy to just fold this into the release and re-run it.

PR #2095, opened 2023-09-25
Avoid useless allocations for interpolations without maps

stof:
The `_performInterpolationHelper` method was implemented with support for skipping the allocation of the InterpolationMap, but `_performInterpolation` was not using that feature, allocating an InterpolationMap to would be ignored.


nex3 (approved):
Oops, thanks!

PR #2097, opened 2023-09-27
Update protocol-version during embedded-host-node release

ntkme:
Closes https://github.com/sass/embedded-host-node/pull/249

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -619,7 +619,7 @@ jobs:
         id: version
         run: |
           echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
-          echo "protocol_version=$(curl https://raw.githubusercontent.com/sass/sass/HEAD/spec/EMBEDDED_PROTOCOL_VERSION)" | tee --append "$GITHUB_OUTPUT"
+          echo "protocol_version=$(curl -fsSL -H "Authorization: Bearer ${{ github.token }}" https://raw.githubusercontent.com/sass/sass/HEAD/spec/EMBEDDED_PROTOCOL_VERSION)" | tee --append "$GITHUB_OUTPUT"
 
       - name: Update version
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -617,7 +617,9 @@ jobs:
 
       - name: Get version
         id: version
-        run: echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
+        run: |
+          echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
+          echo "protocol_version=$(curl https://raw.githubusercontent.com/sass/sass/HEAD/spec/EMBEDDED_PROTOCOL_VERSION)" | tee --append "$GITHUB_OUTPUT"
```
nex3, line 622: There's a risk that this will run into quota limits. We should either pass the GitHub token to it, or find a pre-existing action we can use to download a file with token support. (We could also just use actions/checkout but that'll take longer than necessary.)

.github/workflows/ci.yml
```diff
@@ -617,7 +617,9 @@ jobs:
 
       - name: Get version
         id: version
-        run: echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
+        run: |
+          echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
+          echo "protocol_version=$(curl -fsSL -H "Authorization: Bearer ${{ github.token }}" https://raw.githubusercontent.com/sass/sass/HEAD/spec/EMBEDDED_PROTOCOL_VERSION)" | tee --append "$GITHUB_OUTPUT"
```
ntkme, line 622: Note: `curl -fsSL` is commonly used in scripting:

`-f`: Fail fast with no output at all on server errors.
`-s`: Silent or quiet mode. Do not show progress meter or error messages.
`-S`: When used with -s, --silent, it makes curl show an error message if it fails.
`-L`: If the server reports that the requested page has moved to a different location (indicated with a Location: header and a 3XX response code), this option will make curl redo the request on the new place.

nex3 (approved):

nex3:
Thanks!

PR #2098, opened 2023-09-27
Implement support for the relative color syntax of CSS Color 5

stof:
Implements the change for https://github.com/sass/sass/pull/3673

https://github.com/sass/sass-spec/pull/1940

### DIFF
--- a/lib/src/functions/color.dart
+++ b/lib/src/functions/color.dart
@@ -736,6 +736,14 @@ Object /* SassString | List<Value> */ _parseChannels(
   }
 
   var list = channels.asList;
+  if (list.length > 1) {
+    switch (list[0]) {
+      case SassString(:var text, hasQuotes: false)
+          when equalsIgnoreCase(text, "from"):
+        return _functionString(name, [originalChannels]);
+    }
+  }
+
   if (list.length > 3) {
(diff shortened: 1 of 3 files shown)


lib/src/functions/color.dart
```diff
@@ -736,6 +736,14 @@ Object /* SassString | List<Value> */ _parseChannels(
   }
 
   var list = channels.asList;
+  if (list.length > 1) {
+    switch (list[0]) {
+      case SassString(:var text, hasQuotes: false)
+          when equalsIgnoreCase(text, "from"):
+        return _functionString(name, [originalChannels]);
+    }
+  }
```
nex3, line 745:
```suggestion
  if (list case [SassString(:var text, hasQuotes: false), _, ...]
      when equalsIgnoreCase(text, "from")) {
    return _functionString(name, [originalChannels]);
  }
```

nex3 (changes requested):

lib/src/functions/color.dart
stof: updated

nex3 (approved):

nex3:
Thanks!

PR #2102, opened 2023-09-28
Update the version of Sass used by the website on release

nex3:


jathak (approved):

.github/workflows/ci.yml
```diff
@@ -597,6 +597,18 @@ jobs:
...
+        id: version
+        run: echo "version=${GITHUB_REF##*/}" | tee --append "$GITHUB_OUTPUT"
+
+      - name: Update Dart Sass version
+        run: |
+          cat package.json |
+            jq --arg version ^${{ steps.version.outputs.version }} '
+              .devDependencies.sass |= $version
+            ' > package.json.tmp
+          mv package.json.tmp package.json
```
stof, line 610: As there is a committed lock file, this won't be enough

ntkme: We can run `npm update sass` after that, which should update only `sass` in the package-lock.json.

However, this job should depends on `deploy_npm` so that it's updated only after npm package has been published.

Issue #2113, opened 2023-08-20
Sass  in browser not workig


nbsharath:
I am trying to use sass in the browser through webpack bundle.

```
import * as sass from "sass";
const compiledCSS = await sass.compileStringAsync(value);
```
But its giving this error 

```
error TypeError: J.getInterceptor$x(...) is undefined
    get$isTTY$x sass.dart.js:28035
    hasTerminal0 sass.dart.js:22668
    compileStringAsync1 sass.dart.js:19377
    installTearOffs sass.dart.js:112321
    Primitives_applyFunction sass.dart.js:1143
    Function_apply sass.dart.js:5375
    _callDartFunctionFast sass.dart.js:7244
    ret sass.dart.js:7222
    logic scss-to-css-converter.ts:5
    onClickHandle index.tsx:54
    callCallback react-dom.development.js:19443
    invokeGuardedCallbackImpl react-dom.development.js:19492
    invokeGuardedCallback react-dom.development.js:19567
    invokeGuardedCallbackAndCatchFirstError react-dom.development.js:19581
    executeDispatch react-dom.development.js:30635
    processDispatchQueueItemsInOrder react-dom.development.js:30667
    processDispatchQueue react-dom.development.js:30680
    dispatchEventsForPlugins react-dom.development.js:30691
    dispatchEventForPluginEventSystem react-dom.development.js:30881
    batchedUpdates$1 react-dom.development.js:23761
    batchedUpdates react-dom.development.js:27597
    dispatchEventForPluginEventSystem react-dom.development.js:30880
    dispatchEvent react-dom.development.js:28653
    dispatchDiscreteEvent react-dom.development.js:28624
```

I have webpack.

## Browser:
firefox 
## OS: 
windows
## Version 
^1.66.1

PR #2114, opened 2023-10-09
Fix crash in browser when running alongside NextJS

Goodwine:
Fixes https://github.com/sass/dart-sass/issues/2113


pkg/sass_api/pubspec.yaml
nex3, line 1: You'll need to bump the pubspec version here as well to 9.2.2.

nex3 (approved):

PR #2117, opened 2023-10-13
[Color 4] Dart JS API implementation

jamesnw:
Spec - https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces-js.d.ts.md

- https://github.com/sass/sass/pull/3718
- https://github.com/sass/sass-spec/pull/1949
- https://github.com/sass/embedded-host-node/pull/259

[skip sass-embedded]


jamesnw:
@nex3 This is ready for an initial review. Thanks!

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,366 @@
...
+    ColorSpace spaceClass = ColorSpace.fromName(spaceName);
+    SassColor color = self.toSpace(spaceClass);
+    return color.toGamut();
+  });
+
+  jsClass.defineGetter(
+      'channelsOrNull', (SassColor self) => ImmutableList(self.channelsOrNull));
+
+  jsClass.defineGetter('channels', (SassColor self) {
+    final channelsOrNull = self.channelsOrNull;
```
nex3, line 314: Can't you just use `self.channels` here?

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,366 @@
...
+    ColorSpace spaceClass = ColorSpace.fromName(space);
+    return self.toSpace(spaceClass);
+  });
+
+  jsClass.defineMethod('isInGamut', (SassColor self, [String? space]) {
+    String spaceName = space ?? self.space.name;
+    ColorSpace spaceClass = ColorSpace.fromName(spaceName);
+    SassColor color = self.toSpace(spaceClass);
+    return color.isInGamut;
+  });
```
nex3, line 301: Style nit: this uses a lot of intermediate variables for very small expressions. It would probably be cleaner as just

```suggestion
  jsClass.defineMethod('isInGamut', (SassColor self, [String? space]) =>
      self.toSpace(ColorSpace.fromName(space ?? self.space.name)).isInGamut);
```

You could also potentially factor out some of this into a `SassColor _toSpace(SassColor self, String? space)` function, since the logic gets re-used a lot.

lib/src/js/value/color.dart
```diff
@@ -86,4 +447,30 @@ class _Channels {
...
+
+@JS()
+@anonymous
+class _ConstructionOptions extends _Channels {
+  external String? get space;
+}
+
+@JS()
+@anonymous
+class ChannelOptions {
```
nex3, line 466: These new classes should be private, since they aren't intended to be used outside this library.

lib/src/js/value/color.dart
```diff
@@ -72,7 +372,68 @@ final JSClass colorClass = () {
...
+/// `initial` and returns the result of applying the change for `channel` to
+/// `initial`.
+double? _changeComponentValue(
+    SassColor initial, String channel, Map<String, double?> changes) {
+  var initialValue = initial.channel(channel);
+  if (!changes.containsKey(channel)) {
+    return initialValue;
+  }
+  return changes[channel];
+}
```
nex3, line 387: I think you can more cleanly write this as

```suggestion
        SassColor initial, String channel, Map<String, double?> changes) =>
  changes.containsKey(channel) ? changes[channel] : initial.channel(channel);
```

this also avoids assigning `initialValue` in the case where it's not used.

lib/src/js/value/color.dart
```diff
@@ -72,7 +372,68 @@ final JSClass colorClass = () {
...
+/// Determines the construction space based on the provided options.
+ColorSpace _constructionSpace(_ConstructionOptions options) {
+  if (options.space != null) return ColorSpace.fromName(options.space!);
+  if (options.red != null) return ColorSpace.rgb;
+  if (options.saturation != null) return ColorSpace.hsl;
+  if (options.whiteness != null) return ColorSpace.hwb;
+  throw "No color space found";
+}
+
+void _checkNullAlphaDeprecation(_ConstructionOptions options) {
```
nex3, line 398: Document this and other new functions.

lib/src/js/value/color.dart
```diff
@@ -72,7 +372,68 @@ final JSClass colorClass = () {
...
+  // @todo Verify if options.space is equivalent to "not set"
```
nex3, line 399: I'm not sure what this means.

lib/src/js/value/color.dart
```diff
@@ -72,7 +372,68 @@ final JSClass colorClass = () {
...
+  if (options.alpha == null && options.space == null) {
```
nex3, line 400: `options.alpha == null` will return true if alpha is `undefined`, but I think we do want to allow `undefined` alpha without a deprecation.

lib/src/js/value/color.dart
```diff
@@ -72,7 +372,68 @@ final JSClass colorClass = () {
...
+      "More info: https://sass-lang.com/d/null-alpha",
+      Deprecation.nullAlpha);
+}
+
+void _emitColor4ApiDeprecation(String name) {
+  warnForDeprecationFromJsApi(
+      "$name is deprecated, use `channel` instead.", Deprecation.color4Api);
+}
+
+Map<String, double?> _changedOptions(_ConstructionOptions options) {
```
nex3, line 418: You can do this more easily with [`jsify()`](https://api.dart.dev/stable/3.1.3/dart-js_util/jsify.html). But why bother converting this to a map in the first place instead of just interacting with it as a JS Object?

lib/src/js/value/color.dart
```diff
@@ -86,4 +447,30 @@ class _Channels {
...
+@JS()
+@anonymous
+class ChannelOptions {
+  String? space;
+}
+
+@JS()
+@anonymous
+class InterpolationOptions {
+  external SassColor color2;
```
nex3, line 473: It seems like `color2` should probably be a positional parameter, not an option. Sorry, I should have caught that during API review.

test/embedded/function_test.dart
```diff
@@ -1004,13 +1004,13 @@ void main() {
             });
 
             test("360", () async {
-              expect(
-                  await _deprotofy(_hsl(360, 50, 50, 1.0)), "hsl(0, 50%, 50%)");
+              expect(await _deprotofy(_hsl(360, 50, 50, 1.0)),
+                  "hsl(360, 50%, 50%)");
```
nex3, line 1008: We should still be normalizing polar angles to [0, 360).

nex3 (changes requested):

test/embedded/function_test.dart
jamesnw: Updated in bbccca47560c19647c2c1842b52287e96f0699e9

jamesnw (commented):
@nex3 Thanks for the review!

I think we've addressed everything ~except for the `color2` Interpolation option~. We also moved the constructors from using `SassColor.forSpaceInternal` to the Color-space specific factories, which gives us clamping, etc.

lib/src/js/value/color.dart
jgerigmeyer: Adjusted in https://github.com/oddbird/dart-sass/commit/cfeb456e1bb29ca6400c2972aa8085b75673cc54 and https://github.com/oddbird/sass/commit/424b0d295623c77b5e04c82d8f2e4a59f6a6f492.

CHANGELOG.md
```diff
@@ -152,6 +152,32 @@
 * Added `InterpolationMethod` and `HueInterpolationMethod` which collectively
   represent the method to use to interpolate two colors.
 
+### JS API
+
+* Modify constructor to accept a new `space` option, with support for all the
```
nex3, line 157: Replace "constructor" with the name of the constructor.

lib/src/evaluation_context.dart
```diff
@@ -25,6 +26,16 @@ abstract interface class EvaluationContext {
     }
   }
 
+  /// The current evaluation context, or null if there isn't a Sass stylesheet
+  /// currently being evaluated.
+  static EvaluationContext? get currentOrNull {
```
nex3, line 31: I think this could be private.

lib/src/evaluation_context.dart
```diff
@@ -58,6 +69,17 @@ void warnForDeprecation(String message, Deprecation deprecation) {
...
 
+/// Prints a deprecation warning with [message] of type [deprecation],
+/// using stderr if there is no [EvaluationContext.current].
+void warnForDeprecationFromApi(String message, Deprecation deprecation) {
+  var context = EvaluationContext.currentOrNull;
+  if (context == null) {
+    Logger.stderr().warnForDeprecation(deprecation, message);
+  } else {
+    context.warn(message, deprecation);
+  }
```
nex3, line 80: You can do this a bit more cleanly with pattern matching:

```suggestion
  if (EvaluationContext.currentOrNull case var context?) {
    context.warn(message, deprecation);
  } else {
    Logger.stderr().warnForDeprecation(deprecation, message);
  }
```

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,351 @@
...
+        if (!space.channels.any((channel) => channel.name == key)) {
+          jsThrow(JsError("`$key` is not a valid channel in `$space`."));
+        }
+      }
+
+      var color = self.toSpace(space);
+
+      SassColor changedColor;
+
+      changedValue(String channel) {
```
nex3, line 134: Nit: Add a return type.

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,351 @@
...
+        case ColorSpace.hsl when spaceSetExplicitly:
+          changedColor = SassColor.hsl(
+              changedValue('hue'),
+              changedValue('saturation'),
+              changedValue('lightness'),
+              changedValue('alpha'));
+          break;
+
+        case ColorSpace.hsl:
+          if ((hasProperty(options, 'hue') && options.hue == null) ||
```
nex3, line 148: This will still return true if the user explicitly sets the property to `undefined` rather than just omitting it. If you want to check if it's specifically the value `null`, you'll need to do `identical(options.hue, null)`.

There are three distinct states a property can be in that all look like `options.hue == null` to Dart:

1. Absent (`{}`)
2. Explicitly undefined (`{hue: undefined}`)
3. Explicitly null (`{hue: null}`)

We generally want the first two states to work the same way, and for cases where we support missing channels we want the third to work differently. You can check for each of them as follows, respectively:

1. `!hasProperty(options, 'hue')`
2. `hasProperty(options, 'hue') && isUndefined(options.hue)`
3. `identical(options.hue, null)`

...but since you usually want to have the same behavior for the first two cases, `isUndefined(options.hue)` or `identical(options.hue, null)` are usually the checks you want.

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,351 @@
...
+
+      return changedColor.toSpace(self.space);
+    },
+    'interpolate':
+        (SassColor self, SassColor color2, _InterpolationOptions options) {
+      InterpolationMethod interpolationMethod;
+
+      if (options.method != null) {
+        var hue = HueInterpolationMethod.values.byName(options.method!);
+        interpolationMethod = InterpolationMethod(self.space, hue);
```
nex3, line 299:
```suggestion
      if (options.method case var method?) {
        var hue = HueInterpolationMethod.values.byName(method);
        interpolationMethod = InterpolationMethod(self.space, hue);
```

lib/src/js/value/color.dart
```diff
@@ -72,7 +357,54 @@ final JSClass colorClass = () {
...
 /// component.
-num? _handleUndefinedAlpha(num? alpha) => isUndefined(alpha) ? 1 : alpha;
+double? _handleUndefinedAlpha(double? alpha) => isUndefined(alpha) ? 1 : alpha;
+
+/// This procedure takes a `channel` name, an object `changes` and a SassColor
+/// `initial` and returns the result of applying the change for `channel` to
+/// `initial`.
+double? _changeComponentValue(
+        SassColor initial, String channel, _ConstructionOptions changes) =>
+    hasProperty(changes, channel)
```
nex3, line 367: We want to allow authors to explicitly pass `undefined` to say "use the previous channel value", which means checking whether the property exists isn't sufficient to tell whether you should use the original value. You can use the `isUndefined()` function from `lib/src/js/utils.dart` instead.

This is probably worth adding a test for.

lib/src/js/value/color.dart
```diff
@@ -72,7 +357,54 @@ final JSClass colorClass = () {
...
+
+// Return a SassColor in a named space, or in its original space.
+SassColor _toSpace(SassColor self, String? space) {
+  return self.toSpace(ColorSpace.fromName(space ?? self.space.name));
+}
+
+// If alpha is explicitly null and space is not set, emit deprecation.
+void _checkNullAlphaDeprecation(_ConstructionOptions options) {
+  if (hasProperty(options, 'alpha') &&
+      options.alpha == null &&
```
nex3, line 388: Same `null`/`undefined` issues as before.

nex3 (changes requested):

CHANGELOG.md
jgerigmeyer: Addressed in https://github.com/sass/dart-sass/pull/2117/commits/63c17e31c3f5f6b168bba3f32ecc33ac65113593

lib/src/evaluation_context.dart
jgerigmeyer: Addressed in https://github.com/sass/dart-sass/pull/2117/commits/63c17e31c3f5f6b168bba3f32ecc33ac65113593

lib/src/evaluation_context.dart
jgerigmeyer: Addressed in https://github.com/sass/dart-sass/pull/2117/commits/63c17e31c3f5f6b168bba3f32ecc33ac65113593

lib/src/js/value/color.dart
```diff
@@ -72,7 +357,54 @@ final JSClass colorClass = () {
...
+  if (options.space != null) return ColorSpace.fromName(options.space!);
+  if (options.red != null) return ColorSpace.rgb;
+  if (options.saturation != null) return ColorSpace.hsl;
+  if (options.whiteness != null) return ColorSpace.hwb;
+  throw "No color space found";
+}
+
+// Return a SassColor in a named space, or in its original space.
+SassColor _toSpace(SassColor self, String? space) {
+  return self.toSpace(ColorSpace.fromName(space ?? self.space.name));
```
jgerigmeyer, line 382: This should use the [converting a color](https://github.com/sass/sass/blob/main/proposal/color-4-new-spaces.md#converting-a-color) algorithm, not `color.to-space()`. See https://oddbird.slack.com/archives/C059UTVUVEC/p1697587714921619

jgerigmeyer: @nex3 I’m not seeing where the “removes missing channels when converting to a legacy space” is happening currently in the Dart Sass version of `toSpace` — is that yet to be implemented?

I'm trying to see what we should use internally instead of `toSpace` -- something e.g.:

```dart
  return self.space.convert(ColorSpace.fromName(space ?? self.space.name),
      self.channel0, self.channel1, self.channel2, self.alpha);
```

lib/src/js/value/color.dart
jamesnw: In running some tests, I found that `identical(options.hue, null)` is true if `options.hue` is explicitly undefined. What I was able to determine for each of the checks was this-

1. Absent- `!hasProperty(options, 'hue')`
2. Explicitly undefined- `hasProperty(options, 'hue') && isUndefined(options.hue)`,
3. Explicitly null - `!isUndefined(options.hue) && identical(options.hue, null)`

I also found that `identical(options.hue, null)` was identical to `options.hue == null` in all cases I tested. Does this track with your expectations?

lib/src/js/value/color.dart
nex3: The naming here is a bit confusing. `SassColor.toSpace()` in the Dart API implements the "converting a color" algorithm, specifically because in most _API_ applications you don't want special legacy behavior. The legacy color behavior is handled in the definition of the `color.to-space()` Sass function, because that's the only place where it's relevant.

lib/src/js/value/color.dart
jamesnw: I updated this in edc9b2cfed9d656ceb1522df4f5896e43de34a9d, and added tests to sass-spec in https://github.com/sass/sass-spec/pull/1949/commits/863047107776590ddef78bc225df956be3da7aad, and updated the spec to reflect this in https://github.com/sass/sass/pull/3723/commits/030aa3eba37de8324ed2b80bb0807fbc0f638201.

> This is probably worth adding a test for.

This has higher-level test coverage in sass-spec, but are you recommending unit tests within the `dart-sass` repo?

lib/src/js/value/color.dart
jgerigmeyer: Add a JS util `isExplicitlyNull` function for this.

lib/src/js/value/color.dart
nex3: No, I meant in sass-spec

lib/src/js/value/color.dart
jamesnw: Added `isNull` (to match `isUndefined`) in 632df8cd4908441bc0732a3ae710a7b91fb00f91.

lib/src/js/value/color.dart
```diff
@@ -2,66 +2,364 @@
...
+          if (isNull(options.hue)) {
+            _emitColor4ApiNullDeprecation('hue');
+          } else if (isNull(options.whiteness)) {
+            _emitColor4ApiNullDeprecation('whiteness');
+          } else if (isNull(options.blackness)) {
+            _emitColor4ApiNullDeprecation('blackness');
+          }
+          if (isNull(options.alpha)) {
+            _emitNullAlphaDeprecation();
+          }
```
nex3, line 188: Style nit:

```suggestion
          if (isNull(options.alpha)) _emitNullAlphaDeprecation();
```

nex3 (approved):

nex3:
It looks like the powerless channel tests need to be adjusted for https://github.com/sass/sass/issues/3654

jgerigmeyer:
> It looks like the powerless channel tests need to be adjusted for [sass/sass#3654](https://github.com/sass/sass/issues/3654)

@nex3 Updated in https://github.com/sass/sass-spec/pull/1949/commits/bc2bd0bd33a6352c086af122dea9a3a9204e1657

I may be missing something, but it looks to me like the the remaining failing tests are cases where Dart Sass is trailing the spec (e.g. powerless channels, clamped channels, legacy color equality). Let me know if you think that's incorrect!

nex3:
That's very plausible. Feel free to just mark those tests as todo for now and I'll reenable them as I get the implementation into shape.

lib/src/js/value/color.dart
```diff
@@ -87,8 +87,10 @@ final JSClass colorClass = () {
     'toSpace': (SassColor self, String space) => _toSpace(self, space),
     'isInGamut': (SassColor self, [String? space]) =>
         _toSpace(self, space).isInGamut,
-    'toGamut': (SassColor self, [String? space]) =>
-        _toSpace(self, space).toGamut(),
+    'toGamut': (SassColor self, [String? space]) {
+      var originalSpace = self.space;
+      return _toSpace(self, space).toGamut().toSpace(originalSpace);
```
jgerigmeyer, line 92: > That's very plausible. Feel free to just mark those tests as todo for now and I'll reenable them as I get the implementation into shape.

One of the tests was failing because of this bug, but I added comments to the rest of the failing tests, which I think are all caused by implementation lagging spec changes.

nex3 (approved):

PR #2129, opened 2023-10-30
Remove dead code in the extension store

stof:
The `sources` Map will never contain `extension.extender` as it is an `Extender` object, not a ComplexSelector.
I updated that by assuming that the comment just above it is true and that the selector should actually be removed.
An alternative could be to remove that line of code entirely, preserving the current behavior where nothing is removed.

The dart compiler does not catch this mistake as `Map.remove` allows passing any object, not just objects matching the key type.

I caught this when porting that code to PHP for the Scssphp project, where phpstan (the PHP static analyzer we use) has a different signature for the storage we use, requiring a valid key type.

### DIFF
--- a/lib/src/extend/extension_store.dart
+++ b/lib/src/extend/extension_store.dart
@@ -355,11 +355,6 @@ class ExtensionStore {
           }
         }
       }
-
-      // If [selectors] doesn't contain [extension.extender], for example if it
-      // was replaced due to :not() expansion, we must get rid of the old
-      // version.
-      if (!containsExtension) sources.remove(extension.extender);
     }
 
     return additionalExtensions;


jathak:
Could you look into the failing specs?

stof:
Based on those failures, maybe not removing the selector is the right behavior, and we should remove the line that was doing nothing.
I'd like to have @nex3's opinion on this as the extend code is quite complex.

nex3:
The change in `spec/directives/extend/pseudo/into_pseudo/extends_after` looks potentially desirable to me, but it's not obvious why this was actually necessary so I'd say just rip out the dead code. We can dive deep into a fix later if we find a broken case.

stof:
@nex3 I updated the PR to only remove the dead code.

nex3 (approved):

PR #2130, opened 2023-10-30
[Package Importer] Dart Implementation

jamesnw:
https://github.com/sass/sass/issues/2739

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md) is accepted.

- [x]  sass - https://github.com/sass/sass/pull/3728
- [x] sass-spec - https://github.com/sass/sass-spec/pull/1952
- [x] embedded-host-node - https://github.com/sass/embedded-host-node/pull/260

[skip sass-embedded]


lib/src/importer/filesystem.dart
```diff
@@ -20,6 +20,8 @@ class FilesystemImporter extends Importer {
   final String _loadPath;
 
   /// Creates an importer that loads files relative to [loadPath].
+  /// @nodoc
+  @internal
```
jamesnw, line 24: @nex3 Is the change you requested in our meeting? My recollection was that this was attempting to prevent many `FilesystemImporters` from being created, which doesn't match my understanding of this annotation. I think I'm misunderstanding either the problem or change, and want to double check.

nex3: No, marking the existing constructor as internal and nodoc is effectively a breaking change for anyone who's already using it :sweat_smile:. I think what I suggested is creating a top-level `final` `FilesystemImporter('.')` so that it can be shared among multiple users, and marking _that_ as internal and nodoc.

jamesnw: @nex3 To make sure I understand-

- Add an `internal/nodoc` FilesystemImporter to `lib/sass.dart`
- Everywhere there is a `FilesystemImporter('.')` created, replace that by importing the singleton from `import 'package:sass/sass.dart' as sass;`

I'm seeing roughly 15 of these created throughout `dart-sass`, so I want to confirm before I make the change.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,359 @@
...
+    if (url.scheme != 'pkg') return null;
+    // TODO(jamesnw) Can these errors even be thrown? Or are these cases
+    // filtered out before this?
+    if (url.path.startsWith('/')) {
+      throw "pkg: URL $url must not be an absolute path.";
+    }
+    if (url.path.isEmpty) {
+      throw "pkg: URL $url must not have an empty path.";
+    }
+    if (url.userInfo != '' || url.hasPort || url.hasQuery || url.hasFragment) {
```
jamesnw, line 40: @nex3 I'm wondering  if these checks are necessary. My tests to check these cases are failing before `canonicalize` is ever called, but I didn't see in the `modules` spec any specifications on validation before `canonicalize` is called.

I added these checks as they are explicitly named in the [Semantics for all Package Importers](https://github.com/sass/sass/blob/main/proposal/package-importer.d.ts.md#package-importers).

lib/src/js/compile.dart
```diff
@@ -182,6 +184,15 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
 /// Converts [importer] into an [AsyncImporter] that can be used with
 /// [compileAsync] or [compileStringAsync].
 AsyncImporter _parseAsyncImporter(Object? importer) {
+  if (importer == nodePackageImporter) {
+    if (isBrowser) {
+      jsThrow(JsError(
+          "The Node Package Importer can not be used without a filesystem."));
+    }
+    // TODO(jamesnw) Can we get an actual filename for parity? Is it needed?
+    Uri entryPointURL = Uri.parse(p.absolute('./index.js'));
```
jamesnw, line 193: @nex3 I touched on this in a recent meeting, but the spec specifies using `require.main.filename` here. However, as I looked more, I'm thinking we may want to amend the spec. 

`require.main.filename` is only available when using CommonJS, and there isn't an ECMAScript equivalent. I'm thinking we specify-

"with an associated `entryPointURL` of the absolute path to the directory of the executing script".

Is that specific enough for implementations? We only need the directory, as the filename itself is discarded.

ntkme: > the executing script

I think another question would be the meaning of "the executing script" here. Do we consider the JS file running sass-compilation as "the executing script", or the sass input as "the executing script". My feeling is that this should be interpreted as sass input, because the JS script starting the compilation can be inside a npm global installation of a package and the sass input might be a local file.

jamesnw: @ntkme You're right, there is definitely ambiguity with "the executing script". I'm realizing that both the JS file and the Sass input are both only rough approximations for what we really need. We need the directory-

- Of the npm package that contains the third party packages that will be imported
- That conforms to the Node standards and conventions the Package Importer relies on

Because of how the algorithm works, it can also be any subdirectory of this directory, but it cannot be a parent or sibling. 

Given this, perhaps we should specify the `entryPointURL` is the `current working directory`. This is what is currently implemented here in the Dart implementation, and matches the conventions of the Node Ecosystem, that you are working within an npm package.

ntkme: In node there are two cases:
1. If we're in node REPL, the import is then resolved from `current working directory`.
2. If we're start a script, the import is checked from the directory containing the script, and looked up from that directory.

Let's say we have:
- `a/node_modules`
- `a/b/c/node_modules`
- `a/b/c/test.js`

If we chdir to `a`, and then start `node b/c/test.js`, the resolve would first look at `node_modules` in `a/b/c`, which is **not** `current working directory`...

---

IMO for sass it can be the same. If the sass input has a `file:` url, then the look up begins with the directory containing the file, otherwise, the look up begins with `current working directory`.

lib/src/importer/filesystem.dart
jamesnw: Changed in c22c8ca32c0df0660da66c9c0c740b3aead195d2

lib/src/importer/node_package.dart
jamesnw: Confirmed yes in a meeting, added in ece578e8fcd66ebdf28ce45824bc47e2d031c918.

lib/src/js/compile.dart
jamesnw: Sorry for taking so long to come back to this. I've added a proposed change that I think covers this in https://github.com/sass/sass/pull/3748.

Essentially, for `compile`, we can use the provided `path`, and for `compileString`, we can use a provided `url` option that exists for similar reasons for `importer`, and fall back to the current working directory.

nex3 (changes requested):

lib/src/importer/js_to_dart/async_file.dart
```diff
@@ -21,7 +21,7 @@ import '../utils.dart';
 /// [JSToDartAsyncFileImporter].
 ///
 /// This allows us to avoid duplicating logic between the two importers.
-final _filesystemImporter = FilesystemImporter('.');
+final _filesystemImporter = FilesystemImporter.cwd;
```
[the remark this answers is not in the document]
jamesnw: Removed in 4cdd953a9d677ffb698bcab6ed0b4ec4d768ec92

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+      Uri packageRoot, List<String> subpathVariants, Object exports) {
+    Uri? processVariant(String subpath) {
+      if (subpath == '') {
+        Object? mainExport = _getMainExport(exports);
+        if (mainExport == null) return null;
+        return _packageTargetResolve(subpath, mainExport, packageRoot);
+      } else {
+        if (exports is Map<String, dynamic> &&
+            exports.keys.every((key) => key.startsWith('.'))) {
+          var matchKey = subpath.startsWith('/') ? ".$subpath" : "./$subpath";
```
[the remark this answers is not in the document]
jamesnw: You're right, I don't see a way for subpath to reach this point with a leading `/`. Added in 7f6d15a9de432ac49d6f4254f369f2752a17faa7

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+      if (!keyA.contains("*")) return 1;
+      if (!keyB.contains("*")) return -1;
+      if (keyA.length > keyB.length) return -1;
+      if (keyB.length > keyA.length) return 1;
+      return 0;
+    }
+
+    keys.sort(sorter);
+    return keys;
+  }
```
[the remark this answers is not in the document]
jamesnw: Addressed in 7f6d15a9de432ac49d6f4254f369f2752a17faa7

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+
+  /// Recurses through `exports` object to find match for `subpath`.
+  Uri? _packageTargetResolve(String subpath, Object exports, Uri packageRoot,
+      [String? patternMatch]) {
+    switch (exports) {
+      case String string:
+        if (!string.startsWith('./')) {
+          throw "Invalid Package Target";
```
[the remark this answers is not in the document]
jamesnw: Updated to "Export '$string' must be a path relative to the package root at '$packageRoot'." in 7f6d15a9de432ac49d6f4254f369f2752a17faa7

jamesnw (commented):
@nex3 I think I have addressed all the comments. I'll resolve all the ones that don't have questions or other items for you to check.

jamesnw:
Using require.main.filename will need this fix - https://github.com/google/dart_cli_pkg/pull/151

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+    // If there is a subpath, attempt to resolve the path relative to the
+    // package root, and resolve for file extensions and partials.
+    var relativeSubpath = "${packageRoot.toFilePath()}${p.separator}$subpath";
+    return _filesystemImporter.canonicalize(Uri.file(relativeSubpath));
+  }
+
+  @override
+  ImporterResult? load(Uri url) => _filesystemImporter.load(url);
+
+  /// Takes a string, `path`, and returns a tuple with the package name and the
```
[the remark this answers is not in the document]
jamesnw: I looked a bit more into it, and there are some parts where `/` is required, and some where either would be allowed. 

- Export keys _must_ start with a `./`
- Package name and subpath _must_ be split on `/`.  
- The package name must _not_ contain a `/`.
- The export key must _exactly_ match the subpath.

The one exception that I'm seeing based on the Node source code, it's possible that a path like `package/src\index.scss` could match to an export of `./src\index.scss` in `package`, but since the pkg URL can't have backslashes, we don't need to cover that edge case.

I switched to using path.posix, and threw a couple more errors in ed585906ec0ee6b57518d83d03c5e872e6958402

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+  /// subpath if it is present.
+  (String, String) _packageNameAndSubpath(String path) {
+    var parts = path.split('/');
```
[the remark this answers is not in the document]
jamesnw: Using path.posix in https://github.com/sass/dart-sass/commit/ed585906ec0ee6b57518d83d03c5e872e6958402

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+    var name = parts.removeAt(0);
+    if (name.startsWith('@')) {
+      name = '$name/${parts.removeAt(0)}';
+    }
+    return (name, parts.isNotEmpty ? parts.join('/') : '');
```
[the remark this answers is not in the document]
jamesnw: Using path.posix in https://github.com/sass/dart-sass/commit/ed585906ec0ee6b57518d83d03c5e872e6958402

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,362 @@
...
+    }
+    if (url.path.isEmpty) {
+      throw "pkg: URL $url must not have an empty path.";
+    }
+    if (url.hasQuery || url.hasFragment) {
+      throw "pkg: URL $url must not have a query or fragment.";
+    }
+
+    var baseURL = containingUrl?.scheme == 'file'
+        ? Uri.parse(containingUrl!.toFilePath())
```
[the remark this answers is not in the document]
jgerigmeyer: My conclusion after testing:

- The illegal character was a `:`, because the `containingUrl` being passed to `Uri.directory()` was e.g. `file:///D:/a/dart-sass/dart-sass/sass-spec/spec/sandbox/11582968143023309/js-api-spec.ts`
- It seems like `Uri.directory` is expecting a path, not a full URL with the `file:///` scheme. One hack that got the test passing in Windows was to replace `p.dirname(p.fromUri(Uri.directory(baseURL.toString())))` with `p.dirname(p.fromUri(Uri.directory(baseURL.path.substring(1))))`.
- A better solution seems to be to skip the call to `Uri.directory()` entirely, even in Windows -- `p.fromUri(baseURL)` seems to work fine in Windows as well.

See [full diff here](https://github.com/sass/dart-sass/pull/2130/files/ed585906ec0ee6b57518d83d03c5e872e6958402..b106be19bbe98cb89561c00f171fad7ac60cae28).

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,394 @@
...
+import 'package:sass/src/util/nullable.dart';
+
+import '../importer.dart';
+import './utils.dart';
+import 'dart:convert';
+import '../io.dart';
+import 'package:path/path.dart' as p;
+
+/// An [Importer] that resolves `pkg:` URLs using the Node resolution algorithm.
+class NodePackageImporterInternal extends Importer {
```
nex3, line 16: Why the "internal" on the name?

lib/src/embedded/compilation_dispatcher.dart
```diff
@@ -223,6 +224,11 @@ final class CompilationDispatcher {
       case InboundMessage_CompileRequest_Importer_Importer.notSet:
         _checkNoNonCanonicalScheme(importer);
         return null;
+
+      case InboundMessage_CompileRequest_Importer_Importer.nodePackageImporter:
+        var entryPointUrl =
+            Uri.parse(importer.nodePackageImporter.entryPointPath);
```
nex3, line 230: Again, you _cannot safely parse a path as a URL_. Use `p.toUri()` instead.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,394 @@
...
+  final Uri entryPointURL;
```
nex3, line 17: * Please document this.
* This should be private.
* [Capitalize acronyms and abbreviations longer than two letters like words](https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words)

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,394 @@
...
+
+  /// Creates a Node Package Importer with the associated entry point url
+  NodePackageImporterInternal(this.entryPointURL);
```
nex3, line 20: I'd rather keep this parameter as a path, but if it has to be a URL I think we should probably consume this as a path and only convert it to a URL when we set the internal field. As a rule of thumb, when something is always going to represent a path on disk, it's better to keep it as a path until the last moment.

nex3 (changes requested):
It looks like there are still a lot of places where path functions are being used on URLs, or paths are being converted to URLs and then immediately back again. Please make sure you have a clear and consistent distinction between paths and URLs, and that you're using the right functions with the right types.

lib/src/embedded/compilation_dispatcher.dart
jamesnw: Updated in b686cc134b845344403352014736bf48806f8734

jamesnw (commented):
@nex3 I've gone through and reduced the number of conversions between URIs and paths significantly, which has simplified things quite a bit. 

I think I was trying to follow the datatypes as specified in the spec, but moving to using paths as long as possible as you suggested makes a lot more sense and doesn't change behavior. 

Thanks!

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+              subpath, value as Object, packageRoot, patternMatch);
+          if (result != null) {
+            return result;
+          }
+        }
+
+        return null;
+
+      default:
+        throw "Invalid 'exports' value in ${p.join(packageRoot, 'package.json')}";
```
nex3, line 332: Include the value in the error message. Remember, the point of an error message isn't just to tell a user _that_ something has gone wrong, but to help them understand _why_ it's gone wrong so that they can fix it.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      case []:
+        return null;
+
+      case List<dynamic> array:
+        for (var value in array) {
+          var result = _packageTargetResolve(
+              subpath, value as Object, packageRoot, patternMatch);
+          if (result != null) {
+            return result;
+          }
```
nex3, line 326: Either make this a single-line `return` or make it an `if`/`case` expression so the `result` variable is scoped specifically to the block where it's returned. Either way, make it consistent with the `case` above.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+        return null;
+
+      default:
+        throw "Invalid 'exports' value in ${p.join(packageRoot, 'package.json')}";
+    }
+  }
+
+  /// Returns a path to a package's export without a subpath.
+  Object? _getMainExport(Object exports) {
+    Object? parseMap(Map<String, dynamic> map) {
```
nex3, line 338: Why is this a separate function instead of being inline in the `switch` below?

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+  }
+
+  /// Returns a file path specified by a `subpath` in the `exports` section of
+  /// package.json.
+  ///
+  /// `packageName` is used for error reporting.
+  String? _resolvePackageExports(String packageRoot, String? subpath,
+      Map<String, dynamic> packageManifest, String packageName) {
+    if (packageManifest['exports'] == null) return null;
+    var exports = packageManifest['exports'] as Object;
```
nex3, line 178:
```suggestion
    var exports = packageManifest['exports'] as Object?;
    if (exports == null) return null;
```

Not only does this avoid indexing the manifest twice, it's also clearer that the cast is just to convert `dynamic` into something that has actual type checks.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      Object exports,
+      String? subpath,
+      String packageName) {
+    if (exports is Map<String, dynamic>) {
+      if (exports.keys.any((key) => key.startsWith('.')) &&
+          exports.keys.any((key) => !key.startsWith('.'))) {
+        throw 'Invalid Package Configuration';
+      }
+    }
+    String? processVariant(String? subpath) {
```
nex3, line 218: It's confusing to shadow the outer `subpath` parameter with this one. Consider renaming it.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+            _packageTargetResolve(subpath, mainExport, packageRoot));
+      }
+      if (exports is! Map<String, dynamic> ||
+          exports.keys.every((key) => !key.startsWith('.'))) {
+        return null;
+      }
+      var matchKey = "./$subpath";
+      if (exports.containsKey(matchKey) && !matchKey.contains('*')) {
+        return _packageTargetResolve(
+            matchKey, exports[matchKey] as Object, packageRoot);
```
nex3, line 230: This will currently crash with an unhelpful error if `exports[matchKey]` is null. Same with other `as Object` casts below.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      List<String> list => list,
+      Map<String, dynamic> map => parseMap(map),
+      _ => null
+    };
+  }
+
+  /// Returns a list of all possible variations of `subpath` with extensions and
+  /// partials.
+  ///
+  /// `subpath` is part of the [bare import specifier], so we use the `/`
```
nex3, line 359: This link has no referent.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+  /// separator.
```
nex3, line 360: "Using the `/` separator" doesn't mean anything. Separators are not the only _or_ the most relevant distinction between path syntaxes. A path is either a native filesystem path, in which case it should use `p.*`, or a URL, in which case it should use `p.url.*`. In this case, the `subpath`s being passed in are native paths, so this should use `p.*` everywhere.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+    return switch (exports) {
+      String string => string,
+      List<String> list => list,
+      Map<String, dynamic> map => parseMap(map),
+      _ => null
+    };
+  }
+
+  /// Returns a list of all possible variations of `subpath` with extensions and
+  /// partials.
```
nex3, line 357: Document what it means for null values to appear in the returned list.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+    var paths = <String>[];
+
+    if (subpath == null && addIndex) {
+      subpath = 'index';
+    } else if (subpath != null && addIndex) {
+      subpath = p.posix.join(subpath, 'index');
+    }
+    if (subpath == null) return [null];
+
+    if (['scss', 'sass', 'css'].any((ext) => subpath!.endsWith(ext))) {
```
nex3, line 371: This will have false positives for subpaths like `foocss`. You should check `const {'.scss', '.sass', '.css'}.contains(p.extension(subpath))` instead.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+        '$subpath.scss',
+        '$subpath.sass',
+        '$subpath.css',
+      ]);
+    }
+    var subpathSegments = p.posix.split(subpath);
+    var basename = subpathSegments.last;
+    if (!basename.startsWith('_')) {
+      List<String> prefixedPaths = [];
+      for (final path in paths) {
```
nex3, line 384: Use `var` for local variables.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      paths.addAll([
+        '$subpath.scss',
+        '$subpath.sass',
+        '$subpath.css',
+      ]);
+    }
+    var subpathSegments = p.posix.split(subpath);
+    var basename = subpathSegments.last;
+    if (!basename.startsWith('_')) {
+      List<String> prefixedPaths = [];
```
nex3, line 383: Don't type-annotate local variables. Use `<String>[]` instead, as above.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      paths.add(subpath);
+    } else {
+      paths.addAll([
+        '$subpath.scss',
+        '$subpath.sass',
+        '$subpath.css',
+      ]);
+    }
+    var subpathSegments = p.posix.split(subpath);
+    var basename = subpathSegments.last;
```
nex3, line 381: There's no need to do a full split on the path. Just use `p.dirname()` and `p.basename()`. `p.dirname()` is guaranteed to return `'.'` if the path has only one component.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+        var dirPath = subpathSegments.sublist(0, subpathSegments.length - 1);
+        var pathBasename = p.posix.split(path).last;
+        if (dirPath.isEmpty) {
+          prefixedPaths.add('_$pathBasename');
+        } else {
+          dirPath.add('_$pathBasename');
+          prefixedPaths.add(p.posix.joinAll(dirPath));
+        }
+      }
+      paths.addAll(prefixedPaths);
```
nex3, line 394: Rather than modifying a mutable list, try using list comprehensions to construct the return value, as in:

```dart
var dirname = p.dirname(subpath);
return [
  ...paths,
  for (var path in paths)
    if (dirname == '.') '_${p.basename(path)}'
    else p.join(dirname, '_${p.basename(path)}')
];
```

lib/src/js/compile.dart
```diff
@@ -321,3 +359,18 @@ List<AsyncCallable> _parseFunctions(Object? functions, {bool asynch = false}) {
   });
   return result;
 }
+
+class NodePackageImporterClass {
```
nex3, line 363: Does this need to be a separate class than the Dart `NodePackageImporter`? If so, it should live in its own file and have a name that clearly indicates the distinction between the two, like `JSNodePackageImporter`.

lib/src/js/compile.dart
```diff
@@ -182,6 +185,23 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
 /// Converts [importer] into an [AsyncImporter] that can be used with
 /// [compileAsync] or [compileStringAsync].
 AsyncImporter _parseAsyncImporter(Object? importer) {
+  if (importer is NodePackageImporterClass) {
+    if (isBrowser) {
+      jsThrow(JsError(
+          "The Node Package Importer cannot be used without a filesystem."));
```
nex3, line 191: "Node Package Importer" isn't a proper name, so either refer to the identifier (`NodePackageImporter`) or write it using normal case conventions ("Node package importer"). Also below.

lib/src/js/compile.dart
```diff
@@ -182,6 +185,23 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
...
+    }
+    var entryPointPath = importer.entryPointPath != null
+        ? p.join(p.current, importer.entryPointPath)
+        : requireMainFilename;
+
+    if (entryPointPath == null) {
+      jsThrow(JsError(
+          "The Node Package Importer cannot determine an entry point."
```
nex3, line 199: There's a space missing after this period.

lib/src/js/compile.dart
```diff
@@ -182,6 +185,23 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
...
+          "Please provide an `entryPointPath` to the Node Package Importer."));
```
nex3, line 200: This message should indicate that this is because `require.main.filename` isn't defined.

lib/src/js/utils.dart
```diff
@@ -33,6 +33,17 @@ external JSClass get jsErrorClass;
 /// Returns whether [value] is a JS Error object.
 bool isJSError(Object value) => instanceof(value, jsErrorClass);
 
+final _jsEquals = JSFunction('a', 'b', 'return a===b');
+
+/// Returns where [a] and [b] are considered equal in JS.
+bool jsEquals(Object? a, Object? b) => _jsEquals.call(a, b) == true;
```
nex3, line 39: This appears to be unused.

lib/src/js/utils.dart
```diff
@@ -33,6 +33,17 @@ external JSClass get jsErrorClass;
...
+
+@JS("Symbol")
+class JSSymbol {}
+
+@JS("Symbol")
+external JSSymbol createJSSymbol();
```
nex3, line 45: This is also unused now that we changed the way we initialize the importer.

lib/src/visitor/async_evaluate.dart
```diff
@@ -1706,7 +1707,8 @@ final class _EvaluateVisitor
             return (stylesheet, importer: importer, isDependency: isDependency);
           }
         }
-      } else {
+      }
+      if (_nodeImporter != null) {
```
nex3, line 1711: Empty line between `if` statements.

nex3 (changes requested):

lib/src/importer/node_package.dart
jamesnw: Refactored with guard clauses in c3b1c8b0ea952119418f3df93b1a068149333c38

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+    return recurseUpFrom(baseDirectory);
+  }
+
+  /// Returns a file path specified by the `sass` or `style` values in a package
+  /// manifest, or an `index` file relative to the package root.
+  String? _resolvePackageRootValues(
+      String packageRoot, Map<String, dynamic> packageManifest) {
+    if (packageManifest['sass'] case String sassValue) {
+      if (validExtensions.contains(p.extension(sassValue))) {
+        return p.url.join(packageRoot, sassValue);
```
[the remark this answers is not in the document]
jamesnw: Fixed in c3b1c8b0ea952119418f3df93b1a068149333c38

lib/src/importer/node_package.dart
jamesnw: Updated in c3b1c8b0ea952119418f3df93b1a068149333c38

Side note, is there a way to add a if case after an && in an existing if statement?

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+              subpath, target, packageRoot, patternMatch);
+        }
+      }
+
+      return null;
+    }
+
+    var matches = subpathVariants.map(processVariant).whereNotNull().toList();
+
+    switch (matches) {
```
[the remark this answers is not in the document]
jamesnw: Updated in c3b1c8b0ea952119418f3df93b1a068149333c38

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+      case [var path]:
+        return path;
+      case [_, _, ...] && var paths:
+        throw "Unable to determine which of multiple potential resolutions "
+            "found for ${subpath ?? 'root'} in $packageName should be used. "
+            "\n\nFound:\n"
+            "${paths.join('\n')}";
+    }
+    return null;
```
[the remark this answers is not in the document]
jamesnw: Updated in c3b1c8b0ea952119418f3df93b1a068149333c38

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+  /// will be replaced with `patternMatch`.
+  ///
+  /// Implementation of `PACKAGE_TARGET_RESOLVE` from the [Resolution Algorithm
+  /// Specification](https://nodejs.org/api/esm.html#resolution-algorithm-specification).
+  String? _packageTargetResolve(
+      String? subpath, Object exports, String packageRoot,
+      [String? patternMatch]) {
+    switch (exports) {
+      case String string:
+        if (!string.startsWith('./')) {
```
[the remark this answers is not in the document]
jamesnw: Nice, I hadn't encountered guard clauses before, I like this. Updated in c3b1c8b0ea952119418f3df93b1a068149333c38

lib/src/js/compile.dart
jgerigmeyer: I think it needs to be separate because `entryPointPath` is optional in the JS class, but required in the Dart class -- @jamesnw, can you confirm?

Addressed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc.

lib/src/js/compile.dart
jgerigmeyer: Addressed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

lib/src/js/compile.dart
jgerigmeyer: Addressed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

lib/src/js/compile.dart
jgerigmeyer: Addressed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

lib/src/js/utils.dart
jgerigmeyer: Removed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

lib/src/js/utils.dart
jgerigmeyer: Removed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

lib/src/visitor/async_evaluate.dart
jgerigmeyer: Addressed in https://github.com/oddbird/dart-sass/commit/15e33b9b2fc0a9848076f2356a1c99680f8198bc

jamesnw (commented):
There are still more changes to come, but I wanted to release this to mark some as resolved, and because I have a few follow up questions.

lib/src/js/compile.dart
jamesnw: Yes, and I think it mirrors the intent of the earlier Symbol-based interface, in that there is a clear separation of the public object the user uses to signify that the NodePackageImporter should be used, and the private logic.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,398 @@
...
+
+    if (name.startsWith('.')) {
+      throw "pkg: name $name must not start with a '.'.";
+    }
+    if (name.contains('\\')) {
+      throw "pkg: name $name must not contain a '\\'.";
+    }
+    if (name.contains('%')) {
+      throw "pkg: name $name must not contain a '%'.";
+    }
```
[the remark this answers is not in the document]
nex3: > > These checks effectively add additional, unspecified restrictions to the valid form of a `pkg:` URL.
> 
> This is from the [Node Resolution Algorithm Specification](https://nodejs.org/api/esm.html#resolution-algorithm-specification) for PACKAGE_RESOLVE, so it is an implicit part of the spec already. My understanding is that these are rules that the names of valid Node packages must follow, but I can see the case for not throwing here. Given this, what do you recommend?

For package names with `.` or `\`, I think we should just consider those to be unmatchable by the Node.js package importer. Return `null` and move on, in case another importer can handle them.

For `%`, I think it's really important that `pkg:%66oo` and `pkg:foo` be identical, because that's how URLs behave. We should URL-decode the package name (probably with `p.fromUri()`) and only _then_ return `null` if it still contains a `%`.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+  ///
+  /// Because this is a bare import specifier and not a path, we always use `/`
+  /// to avoid invalid values on non-Posix machines.
+  (String, String?) _packageNameAndSubpath(String specifier) {
+    var parts = p.url.split(specifier);
+    var name = p.fromUri(parts.removeAt(0));
+
+    if (name.contains('%')) {
+      throw "pkg: name $name must not contain a '%'.";
+    }
```
nex3, line 108: If this is a Node restriction, `canonicalize()` should return null rather than failing.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+
+    if (name.startsWith('@')) {
+      if (parts.isEmpty) {
+        throw "pkg: name $name is an invalid package name."
+            "Scoped packages, which start with '@', must have a second segment.";
```
nex3, line 113: I think this should also return null rather than failing. You could imagine another package system that allows `@foo` as a package name.

This makes me think that maybe we should add to the spec for package importers that custom implementations should return null for package names that are invalid for that particular packaging system, rather than throwing an error.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+    return (name, subpath);
+  }
+
+  /// Returns an absolute path to the root directory for the most proximate
+  /// installed `packageName`.
+  ///
+  /// Implementation of `PACKAGE_RESOLVE` from the [Resolution Algorithm
+  /// Specification](https://nodejs.org/api/esm.html#resolution-algorithm-specification).
+  String? _resolvePackageRoot(String packageName, String basePath) {
+    String? baseDirectory;
```
nex3, line 127: Why not declare this as `var baseDirectory = basePath`, since it gets `??`ed with `basePath` anyway immediately afterwards?

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+      if (p.split(baseDirectory).length == 1) return null;
+    }
+  }
+
+  /// Returns a file path specified by the `sass` or `style` values in a package
+  /// manifest, or an `index` file relative to the package root.
+  String? _resolvePackageRootValues(
+      String packageRoot, Map<String, dynamic> packageManifest) {
+    if (packageManifest['sass'] case String sassValue) {
+      if (validExtensions.contains(p.extension(sassValue))) {
```
nex3, line 142: Don't pass URL paths to native path methods. In the case of `p.extension()` it _probably_ doesn't matter in practice, but it would make it clearer what's a native path and what's a URL path.

nex3, line 142: Rather than an `if` nested in an `if ... case`, use `if ... case ... when`.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+
+    if (packageManifest['style'] case String styleValue) {
+      if (validExtensions.contains(p.extension(styleValue))) {
+        return p.join(packageRoot, styleValue);
+      }
+    }
+
+    var result = resolveImportPath(p.join(packageRoot, 'index'));
+    if (result != null) return result;
+    return null;
```
nex3, line 155: This is equivalent to just returning the result directly.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+      String packageName) {
+    if (exports is Map<String, dynamic> &&
+        exports.keys.any((key) => key.startsWith('.')) &&
+        exports.keys.any((key) => !key.startsWith('.'))) {
+      throw '`exports` in $packageName can not have both conditions and paths '
+          'at the same level.\n'
+          'Found ${exports.keys.map((key) => '"$key"').join(',')} in '
+          '${p.join(packageRoot, 'package.json')}.';
+    }
+    String? processVariant(String? variant) {
```
nex3, line 207: Just write this inline in the `map()` call below.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+    var matches = subpathVariants.map(processVariant).whereNotNull().toList();
+
+    return switch (matches) {
+      [var path] => path,
+      [_, _, ...] && var paths =>
+        throw "Unable to determine which of multiple potential resolutions "
+            "found for ${subpath ?? 'root'} in $packageName should be used. "
+            "\n\nFound:\n"
+            "${paths.join('\n')}",
+      _ => null
```
nex3, line 257: It would be simpler to reorder these:

```suggestion
      [] => null,
      var paths =>
        throw "Unable to determine which of multiple potential resolutions "
            "found for ${subpath ?? 'root'} in $packageName should be used. "
            "\n\nFound:\n"
            "${paths.join('\n')}"
```

...since this guarantees that by the time you hit the last case there's more than one path.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+
+  /// Returns a path to a package's export without a subpath.
+  Object? _getMainExport(Object exports) {
+    return switch (exports) {
+      String string => string,
+      List<String> list => list,
+      Map<String, dynamic> map
+          when !map.keys.any((key) => key.startsWith('.')) =>
+        map,
+      Map<String, dynamic> map when map.containsKey('.') && map['.'] != null =>
```
nex3, line 338: You can make this a map pattern: `<String, dynamic>{'.': var export?} => export`

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+    var paths = <String>[];
+
+    if (subpath == null && addIndex) {
+      subpath = 'index';
+    } else if (subpath != null && addIndex) {
+      subpath = p.join(subpath, 'index');
+    }
+    if (subpath == null) return [null];
+
+    if (const {'.scss', '.sass', '.css'}.contains(p.extension(subpath))) {
```
nex3, line 359: If you're going to define `validExtensions` this should probably use it.

lib/src/js/compile.dart
```diff
@@ -182,6 +186,16 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
 /// Converts [importer] into an [AsyncImporter] that can be used with
 /// [compileAsync] or [compileStringAsync].
 AsyncImporter _parseAsyncImporter(Object? importer) {
+  if (importer is JSNodePackageImporter) {
+    if (isBrowser) throwMissingFileSystem();
+    var entryPointPath = importer.entryPointPath != null
+        ? p.join(p.current, importer.entryPointPath)
```
nex3, line 192: This is equivalent to `p.absolute(importer.entryPointPath)`, but why is making this absolute necessary at all?

nex3 (changes requested):

lib/src/importer/js_node_package.dart
```diff
@@ -0,0 +1,26 @@
...
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:node_interop/js.dart';
+
+import '../js/utils.dart';
+
+/// The JS [NodePackageImporter] class that can be added to the
+/// `importers` option to enable loading `pkg:` URLs from `node_modules`.
+class JSNodePackageImporter {
```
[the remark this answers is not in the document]
jgerigmeyer: @nex3 We discussed that [here](https://github.com/sass/dart-sass/pull/2130#discussion_r1443312185). My understanding was it needs to be separate because `entryPointPath` is optional in the JS class, but required in the Dart class. When I tried just exposing the Dart class to JS, it complained about that discrepancy.

James said:

> I think it mirrors the intent of the earlier Symbol-based interface, in that there is a clear separation of the public object the user uses to signify that the NodePackageImporter should be used, and the private logic.

lib/src/importer/js_node_package.dart
```diff
@@ -0,0 +1,26 @@
...
+
+Never throwMissingFileSystem() {
+  jsThrow(JsError(
+      "The Node package importer cannot be used without a filesystem."));
+}
+
+Never throwMissingEntryPointPath() {
+  jsThrow(JsError("The Node package importer cannot determine an entry point "
+      "because `require.main.filename` is not defined. "
+      "Please provide an `entryPointPath` to the `NodePackageImporter`."));
```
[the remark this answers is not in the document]
jamesnw: I moved these into the constructor in 0434343d417baafb330e91cc2a09ad9501bd0d91. 

This makes it so that the entryPointPath is typed as optional, but programmatically required, which I don't love, but it does make sense to not offload this logic to the constructing class.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,382 @@
...
+
+/// An [Importer] that resolves `pkg:` URLs using the Node resolution algorithm.
+class NodePackageImporter extends Importer {
+  /// The starting path for canonicalizations without a containing URL.
+  final String _entryPointPath;
+
+  /// Creates a Node package importer with the associated entry point.
+  NodePackageImporter(this._entryPointPath);
+
+  static const validExtensions = {'.scss', '.sass', '.css'};
```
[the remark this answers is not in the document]
jamesnw: I documented and made it private in 1cb845c031cf86703a91214766a856dd194e746f, but I'm not sure what you mean by `global` here- should this be moved to `lib/src/importer/utils.dart`? But then I wouldn't want it to be private, correct?

lib/src/js/compile.dart
jamesnw: The entryPointPath may be relative or absolute, and the spec specifies that relative paths will be resolved relative to the CWD. The NodePackageImporter requires an associated absolute path. 

I realized I could move this into the constructor as well, and did that in 125acf351141814789ab7ae3797a48226dd7137f.

lib/src/importer/node_package.dart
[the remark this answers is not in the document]
jamesnw: I moved this to private and documented it in https://github.com/sass/dart-sass/pull/2130/commits/1cb845c031cf86703a91214766a856dd194e746f. I wasn't sure what you meant by global and private- I could move it to `./utils` to make it accessible elsewhere, but then wouldn't be able to make it private.

lib/src/importer/js_node_package.dart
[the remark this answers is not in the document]
nex3: Sorry, GitHub's UI makes it really hard to follow conversations in older revisions :weary:.

JS can have a different constructor than Dart (like we do for the value types), and that constructor can fill in the missing value before calling the Dart constructor.

lib/src/importer/js_node_package.dart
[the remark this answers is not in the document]
nex3: It's functionally optional with a default value, which is a pretty standard way for a parameter to work.

lib/src/importer/node_package.dart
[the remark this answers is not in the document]
nex3: "Global" means "not a member of a class"—that is, at the top level of this library.

lib/src/js/legacy.dart
```diff
@@ -289,6 +297,31 @@ NodeImporter _parseImporter(RenderOptions options, DateTime start) {
   return NodeImporter(contextOptions, includePaths, importers);
 }
 
+/// Creates an [AsyncImportCache] for Package Importers.
+AsyncImportCache? _parsePackageImportersAsync(
+    RenderOptions options, DateTime start) {
+  if (options.pkgImporter is JSNodePackageImporter) {
+    var entryPointPath = options.pkgImporter?.entryPointPath != null
+        ? options.pkgImporter!.entryPointPath
+        : requireMainFilename;
```
nex3, line 306: This can now be `options.pkyImporter?.entryPointPath ?? requireMainFilename`

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+  late final String _entryPointPath;
+
+  /// Creates a Node package importer with the associated entry point.
+  NodePackageImporter(String? entryPointPath) {
+    if (entryPointPath == null) {
+      throw "The Node package importer cannot determine an entry point "
+          "because `require.main.filename` is not defined. "
+          "Please provide an `entryPointPath` to the `NodePackageImporter`.";
+    }
+    if (isBrowser) {
```
nex3, line 28: Nit: `else if`

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+      throw "The Node package importer cannot be used without a filesystem.";
+    }
+    _entryPointPath = p.absolute(entryPointPath);
+  }
+
+  /// The set of file extensions that Sass can parse. NodePackageImporter will
```
nex3, line 34: [Separate the first sentence of a doc comment into its own paragraph](https://dart.dev/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph)

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+        : _entryPointPath;
+
+    var (packageName, subpath) = _packageNameAndSubpath(url.path);
+
+    // If the package name is not a valid Node package name, return null in case
+    // another importer can handle.
+    if (packageName.startsWith('.') ||
+        packageName.contains('\\') ||
+        packageName.contains('%') ||
+        packageName.startsWith('@') && !packageName.contains(p.url.separator)) {
```
nex3, line 68: Use parentheses to clarify precedence when mixing `||` and `&&`

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+    return (name, subpath);
+  }
+
+  /// Returns an absolute path to the root directory for the most proximate
+  /// installed `packageName`.
+  ///
+  /// Implementation of `PACKAGE_RESOLVE` from the [Resolution Algorithm
+  /// Specification](https://nodejs.org/api/esm.html#resolution-algorithm-specification).
+  String? _resolvePackageRoot(String packageName, String basePath) {
+    String baseDirectory = basePath;
```
nex3, line 136:
```suggestion
    var baseDirectory = basePath;
```

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+  /// Returns a file path specified by the `sass` or `style` values in a package
+  /// manifest, or an `index` file relative to the package root.
+  String? _resolvePackageRootValues(
+      String packageRoot, Map<String, dynamic> packageManifest) {
+    if (packageManifest['sass'] case String sassValue
+        when _validExtensions.contains(p.url.extension(sassValue))) {
+      return p.join(packageRoot, sassValue);
+    }
+
+    if (packageManifest['style'] case String styleValue
```
nex3, line 155: Nit: `else if`

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,389 @@
...
+          '${p.join(packageRoot, 'package.json')}.';
+    }
+
+    var matches = subpathVariants
+        .map((String? variant) {
+          if (variant == null) {
+            return _getMainExport(exports).andThen((mainExport) =>
+                _packageTargetResolve(variant, mainExport, packageRoot));
+          }
+          if (exports is! Map<String, dynamic> ||
```
nex3, line 220: Nit: `else if`

lib/src/js/compile.dart
```diff
@@ -182,6 +185,11 @@ OutputStyle _parseOutputStyle(String? style) => switch (style) {
 /// Converts [importer] into an [AsyncImporter] that can be used with
 /// [compileAsync] or [compileStringAsync].
 AsyncImporter _parseAsyncImporter(Object? importer) {
+  if (importer is JSNodePackageImporter) {
+    var entryPointPath = importer.entryPointPath ?? requireMainFilename;
```
nex3, line 189: At this point you can just inline this into the constructor call.

nex3 (changes requested):

lib/src/importer/js_node_package.dart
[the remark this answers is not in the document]
jamesnw: Ah, ok. I made that change in 7f2132ad741547d18cf2ea265e0c25addd39fd7f.

lib/src/importer/node_package.dart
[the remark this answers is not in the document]
jamesnw: Updated in https://github.com/sass/dart-sass/pull/2130/commits/8e60938a09515e9909edf9f0dc26ec3b78d77060

nex3 (approved):

ntkme:
Our spec does not cover command line usage, but I think it would be useful to add a command line flag to for dart-sass command line. E.g. `sass --node-package-importer /entrypoint/path`.

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,388 @@
...
+    return (name, subpath);
+  }
+
+  /// Returns an absolute path to the root directory for the most proximate
+  /// installed `packageName`.
+  ///
+  /// Implementation of `PACKAGE_RESOLVE` from the [Resolution Algorithm
+  /// Specification](https://nodejs.org/api/esm.html#resolution-algorithm-specification).
+  String? _resolvePackageRoot(String packageName, String basePath) {
+    var baseDirectory = basePath;
```
jamesnw, line 131: @nex3 `basePath` here can now either be a file (from `containingUrl` or directory (from `entryPointDirectory`). Because of this, I moved the `p.dirname` to the end, otherwise the last segment in the provided directory would never be tested. 

The downside to this is that for paths, it will now test something like `file://containingUrl.scss/node_modules/packagename`. It will never succeed, because `containingUrl.scss` can not be simultaneously a path and a directory.

I considered adding a `dirExists` check initially, but realized that would always add an additional disk read, where this only adds an additional disk read for canonicalizations with a containingUrl.

nex3: I don't think it makes sense to handle this as either a file _or_ a directory since, as you point out, there's not a reliable way to distinguish the two without checking the filesystem. Instead, when you assign `basePath` in `canonicalize()`, add a `p.dirname()` there if it's coming from `containingUrl` so that the same variable always represents the same type of data.

jamesnw: Good point- I adjusted in https://github.com/sass/dart-sass/pull/2130/commits/53472fdb880494924f06d5b0974cfac94b36b433

lib/src/importer/node_package.dart
```diff
@@ -0,0 +1,388 @@
...
+    } else if (p.url.isAbsolute(url.path)) {
+      throw "A pkg: URL's path must not begin with /.";
+    } else if (url.path.isEmpty) {
+      throw "A pkg: URL must not have an empty path.";
+    } else if (url.hasQuery || url.hasFragment) {
+      throw "A pkg: URL must not have a query or fragment.";
+    }
+
+    var basePath = containingUrl?.scheme == 'file'
+        ? p.fromUri(containingUrl!)
```
nex3: This still needs a `p.dirname()`, since `containingUrl` points to a file and not a directory.

nex3 (changes requested):

lib/src/importer/node_package.dart
jamesnw: That change was made in https://github.com/sass/dart-sass/pull/2130/commits/53472fdb880494924f06d5b0974cfac94b36b433

nex3 (approved):

PR #2134, opened 2023-11-21
[Shared Resources] dart-sass implementation

jerivas:
[Issue](https://github.com/sass/sass/issues/3296)

Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/shared-resources.d.ts.md) is accepted.

In dart-sass the compiler classes are just thin wrappers around the compile* functions because there is no underlying compiler process

- https://github.com/sass/sass/pull/3744
- https://github.com/sass/embedded-host-node/pull/261
- https://github.com/sass/sass-spec/pull/1954

[skip sass-embedded]


jerivas:
[I want to share resources across multiple invocations of the Sass compiler's JS API.](https://trello.com/c/BT4VgchI/4-i-want-to-share-resources-across-multiple-invocations-of-the-sass-compilers-js-api)

lib/src/js/compiler.dart
```diff
@@ -0,0 +1,104 @@
...
+      _compilations.remove(compilation);
+    }, (error) {
+      _compilations.remove(compilation);
+    });
+  }
+}
+
+/// The JS API Compiler class
+///
+/// See https://github.com/sass/sass/spec/tree/main/js-api/compile.d.ts for
```
jamesnw, line 36: This link isn't working for me- I think this should be https://github.com/sass/sass/blob/main/spec/js-api/compile.d.ts.md. Although I do wonder, since the complexity of the Compiler is much less than `compile`, if it makes sense to inline these descriptions.

jamesnw (approved):

lib/src/js/compiler.dart
jerivas: You're right, the other JS classes don't have links to the spec. Updated in e6ff49b3a05882c8a935466e461e8aa43d1c945c

lib/src/js/compiler.dart
```diff
@@ -0,0 +1,90 @@
+import 'dart:js_util';
+
+import 'package:node_interop/js.dart';
+
+import 'compile.dart';
+import 'compile_options.dart';
+import 'reflection.dart';
+import 'utils.dart';
+
+class Compiler {
```
nex3, line 10: Document this and its members.

nex3, line 10: This should be a private class, since it's not used (by Dart) outside this file.

lib/src/js/compiler.dart
```diff
@@ -0,0 +1,90 @@
...
+
+  void throwIfDisposed() {
+    if (_disposed) {
+      jsThrow(JsError('Compiler has already been disposed.'));
+    }
+  }
+}
+
+class AsyncCompiler extends Compiler {
+  final Set<Promise> _compilations = {};
```
nex3, line 21: Use a [`FutureGroup`](https://pub.dev/documentation/async/latest/async/FutureGroup-class.html) for this.

nex3, line 21: Since this member is accessed outside of its class (and the class is anyway), it should be public to help document that it's not exclusively class-internal.

lib/src/js/compiler.dart
```diff
@@ -0,0 +1,90 @@
...
+      _compilations.remove(compilation);
+    }, (error) {
+      _compilations.remove(compilation);
+    });
+  }
+}
+
+/// The JavaScript `Compiler` class.
+final JSClass compilerClass = () {
+  var jsClass = createJSClass('sass.Compiler', () => Compiler());
```
nex3, line 36: I think we should throw an error if the user attempts to manually construct a compiler instance without going through the `init` functions. Can you add that to the spec/tests?

nex3 (changes requested):

lib/src/js/compiler.dart
jamesnw: @nex3 Looking at the proposal, we should only be exposing the Compiler type, but not the class itself. 

That means the change we need to make is updating the embedded host and dart-sass here to not export the class, and then there wouldn't be a risk of a user calling `new Compiler()`- correct?

lib/src/js/compiler.dart
jamesnw: This will be contingent on the outcome of https://github.com/sass/dart-sass/pull/2134/files/e6ff49b3a05882c8a935466e461e8aa43d1c945c#r1423107599

jamesnw (commented):
@nex3 I've made most of the requested changes, but have some questions that could use your input.

lib/src/js/compiler.dart
ntkme: You can avoid this behavior by wrapping the future: Chain the original Future with a `.catch` that returns the thrown exception, add this wrapped Future to FutureGroup, and returns the original Future to the user. It's a bit ugly but should work as needed.

jamesnw: Thanks, I added a `catchError` in bafe182f948ae956b52804835769915f36e163ec. Since we are only using this to determine that all Futures have completed and are returning the compilation results as they come, we ldon't need to worry about the return value.

lib/src/js/compiler.dart
jerivas: I see James has already implemented a constructor that throws in https://github.com/sass/dart-sass/pull/2134/commits/53514e390f50cd5bcbf82136222cce8946a76253. The Dart codebase uses that trick in several places (and I used it too to make SassCalculations throw), but I wonder how we should implement this in the embedded host.

We can update the spec to mark the constructor as private so tsc yells at authors for constructing Compiler instances, but this won't stop anyone from actually doing it. In fact, this is the case for [`SassCalculation`](https://github.com/sass/embedded-host-node/blob/06974e1f0fdded65af93135fb1ea39fcbec2c2eb/lib/src/value/calculations.ts#L34). I just checked and I can instantiate it in plain JS because there's no runtime check, which is an inconsistency with the dart-sass implementation.

This raises the bigger picture question of why do we want to use the `init*` functions and disallow direct instantiation? I think enforcing this will require some closure tricks (which is what Dart does on the code it generates), unless I'm missing a more straight-forward approach.

jerivas: Clarification: I had forgotten Natalie mentioned that the types used by the embedded host are not consumed by end users, so the fact that the constructor is private there is not relevant unless it's also private in the spec.

jamesnw: > This raises the bigger picture question of why do we want to use the `init*` functions and disallow direct instantiation?

I went through the same thought process- the crucial point is that `initAsyncCompiler` returns a Promise, and constructors can't be async.

Because the types are defined in the sass pkg, in node-embedded-host I was able to add a private flag to the constructor that can only be passed via the init function - https://github.com/sass/embedded-host-node/pull/261/commits/7b7fbf7b01d2850a2a9353630ffcc21768719fab, allowing us to avoid the IIFE alternatives.

lib/src/js/compiler.dart
```diff
@@ -1,33 +1,41 @@
...
-      _compilations.remove(compilation);
-    }, (error) {
-      _compilations.remove(compilation);
+  /// Adds a compilation to the FutureGroup.
+  void addCompilation(Promise compilation) {
+    Future<dynamic> comp = promiseToFuture(compilation);
+    comp.catchError((err) {
+      return;
     });
+    compilations.add(comp);
```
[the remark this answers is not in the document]
jamesnw: Updated. Tests added in https://github.com/sass/sass-spec/pull/1954/commits/2730112fb2a85205209a33b99842e5c430ae8818

nex3 (approved):

jgerigmeyer (approved):

PR #2149, opened 2023-12-14
Upload releases for musl-libc and android

### LINKED ISSUE #2148: Doesn't run on alpine linux
On alpine there's no glibc only musllibc.

Errors:

        /dart-sass # ldd src/dart
                /lib64/ld-linux-x86-64.so.2 (0x7f26ab540000)
                libdl.so.2 => /lib64/ld-linux-x86-64.so.2 (0x7f26ab540000)
                libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f26ab540000)
                libm.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f26ab540000)
                libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f26ab540000)
        /dart-sass # ls  /lib64/ld-linux-x86-64.so.2
        ls: /lib64/ld-linux-x86-64.so.2: No such file or directory

and here

        [INFO] --- dart-sass:1.2.0:compile-sass (generate-css-using-dart-sass) @ cms ---
        [INFO] Auto-detected operating system: linux
        [INFO] Auto-detected operating system architecture: x64
        [INFO] Auto-detected latest version: 1.69.5
        /tmp/dart-sass-maven-plugin/linux/x64/1.69.5/sass: exec: line 20: /tmp/dart-sass-maven-plugin/linux/x64/1.69.5/src/dart: not found

my system:

        Apache Maven 3.9.5 (57804ffe001d7215b5e7bcb531cf83df38f93546)
        Maven home: /usr/share/maven
        Java version: 1.8.0_372, vendor: IcedTea, runtime: /usr/lib/jvm/java-1.8-openjdk/jre
        Default locale: en_US, platform encoding: UTF-8
        OS name: "linux", version: "6.6.4-100.fc38.x86_64", arch: "amd64", family: "unix"

ntkme:
This PR creates release assets for linux-musl and android using unofficial dart-sdk builds that I'm maintaining:

- https://github.com/dart-musl/dart/releases
- https://github.com/dart-android/dart/releases

This PR also does a refactor of CI job compositions to make it easier to understand and maintain. A behavior difference is that now the GitHub Release is only created if all the builds are successful, which avoid the issue we had in the past that some of the release artifact were missing due to random issues. In other words, the release assets are either all or none.

Closes #2148


ntkme:
Sample run (modified to run release only): https://github.com/ntkme/dart-sass-test/actions/runs/7204714042
Sample release: https://github.com/ntkme/dart-sass-test/releases/tag/1.69.6

---

Screenshot 2023-12-13 at 9 13 33 PM

farnetto:
The musl version works well for us. Thanks!

.github/workflows/build-android.yml
nex3, line 1: Have we ever had anyone request native Android support for Dart Sass? I'm not sure this is worth the trouble.

.github/workflows/build-linux-musl.yml
```diff
@@ -0,0 +1,89 @@
...
+                     --volume "$PWD:$PWD" \
+                     --workdir "$PWD" \
+                     docker.io/library/dart <<'EOF'
+          set -e
+          curl -fsSL -H "Authorization: Bearer ${{ github.token }}" "https://github.com/bufbuild/buf/releases/latest/download/buf-$(uname -s)-$(uname -m).tar.gz" | tar -xzC /usr/local --strip-components 1
+          dart pub get
+          dart run grinder protobuf
+          EOF
+
+      # There is a bug in qemu's mremap causing pthread_getattr_np in musl to stuck in a loop on arm.
```
nex3, line 44: Can you link a tracking issue for this bug?

.github/workflows/build-linux-musl.yml
```diff
@@ -0,0 +1,89 @@
...
+      # Unless qemu fixes the bug or we get a real linux-arm runner, we cannot build aot-snapshot
+      # for arm on CI. So, we create an kernel snapshot for arm build in amd64 container instead.
```
nex3, line 46: Nit:
```suggestion
      # for arm on CI. So, we create a kernel snapshot for arm build in amd64 container instead.
```

.github/workflows/build-linux-musl.yml
```diff
@@ -0,0 +1,89 @@
...
+      #
+      # In addition, we need to rename the artifact from -linux- to -linux-musl- to avoid conflict
+      # with glibc builds.
```
nex3, line 49: Nit, but I'd put this comment above the `find` command in the run section.

.github/workflows/build-linux-musl.yml
```diff
@@ -0,0 +1,89 @@
...
+                     --workdir "$PWD" \
+                     ghcr.io/dart-musl/dart <<'EOF'
+          set -e
+          dart pub get
+          dart run grinder pkg-standalone-linux-${{ matrix.arch }}
+          find build -name '*.tar.gz' -print0 | xargs -0 -n 1 -- sh -xc 'mv "$1" "$(echo "$1" | sed -e "s/linux/linux-musl/")"' --
+          EOF
+
+      # The kernel snapshot created for arm in the previous step is bundling a glibc based dart runtime
+      # due to how cli_pkg download the sdk for building non-native platforms. Therefore we need to
```
nex3, line 64: Nit:
```suggestion
      # due to how cli_pkg downloads the sdk for building non-native platforms. Therefore we need to
```

.github/workflows/build-linux-musl.yml
```diff
@@ -0,0 +1,89 @@
...
+      # replace it with musl-libc based dart runtime to create a working linux-musl-arm package.
```
nex3, line 65: Would it make sense to build more of this logic into `cli-pkg`? We could teach it `pkg-standalone-linux-musl-*` with your SDK releases.

.github/workflows/build-windows.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x64
+            runner: windows-latest
+          - arch: ia32
+            runner: windows-latest
+          # - arch: arm64
+          #   runner: windows-latest
```
nex3, line 22: Add a comment explaining why this is commented out

.github/workflows/ci.yml
```diff
@@ -1,332 +1,21 @@
...
-        dart-sdk: ${{ matrix.dart_channel }}
-        github-token: ${{ github.token }}
-
-    - run: dart run grinder pkg-npm-dev
-    - name: Run tests
-      run: dart run test -p chrome -j 2
-      env:
-        CHROME_EXECUTABLE: chrome
+  test:
+    uses: ./.github/workflows/test.yml
```
nex3, line 11: I didn't realize you could put `uses:` directly within a job.

nex3 (changes requested):
This refactor looks really nice, thanks for taking the time!

.github/workflows/ci.yml
ntkme: https://docs.github.com/en/actions/using-workflows/reusing-workflows#nesting-reusable-workflows

.github/workflows/build-android.yml
ntkme: There were a few people asking this from jekyll community. Interestingly, now there are some active users:

- https://rubygems.org/gems/sass-embedded/versions/1.69.5-aarch64-linux-android (downloaded 242 times as of writing)
- https://rubygems.org/gems/sass-embedded/versions/1.69.5-arm-linux-androideabi (downloaded 159 times as of writing)

.github/workflows/build-linux-musl.yml
ntkme: That's certainly possible, but I would consider that at a later time due to the complexity. The problem mainly lies on dart-sdk only have the knowledge of itself running as some "linux" variant, and have no concept of "linux-gnu" or "linux-musl", which would require lots of structural changes to cli_pkg to accommodate this. Also, we need a way to know what is the current running variant, that either the current running dart-sdk need to be able to tell what kind of linux variant by itself, or we need to parse the dart executable file as ELF, and then detect what it is based on the interpreter field.

stof:
should we also add support for musl-libc in the embedded-host-node repository so that the npm package `sass-embedded` can have this executable among its dependencies as well ?

ntkme:
@stof I’m planning to look into that after this change goes out. We don’t necessarily need to support both at the same time.

.github/workflows/build-android.yml
ntkme: I guess another point is SWC (https://swc.rs/) also has native android support: https://www.npmjs.com/package/@swc/core-android-arm64

This means there are people out there using android for web development...

.github/workflows/build-linux-musl.yml
nex3: Would it be possible for your fork to add "musl" somewhere in the `dart --version` output? That seems like the easiest way to tell—we don't need an actual API call in the SDK itself necessarily.

nex3 (approved):

.github/workflows/build-linux-musl.yml
ntkme: My "fork" does not modify any dart-sdk source code. Only thing it does is running a CI/CD with a different build toolchain for building dart-sdk on alpine linux.

The reason I said it is difficult to do in either dart or cli_pkg is because the platform logic in dart is closely tied to the `Abi` class in `ffi` package. For example, the `dart --version` pretty much prints [`Abi.current.toString()`](https://api.dart.dev/stable/3.2.3/dart-ffi/Abi/toString.html). In order to print musl, we can possibly either add new constant, but this will likely break existing code that does the platform detection via `Abi` class; or modify `dart --version` to hard code a different string instead of using constant, which again might be unexpected.

I don't think change the version string is the correct behavior. I think the right way to deal with this in dart-sdk would be to provide new API that returns the ["target triplet"](https://wiki.osdev.org/Target_Triplet). For example, in ruby we have `RbConfig::CONFIG['arch']` which would return `x86_64-linux-musl` that clearly identify the right platform.

nex3:
@ntkme Looks like the release is failing to find `/.../dart/bin/snapshots/gen_kernel_aot.dart.snapshot` for [linux-musl 386](https://github.com/sass/dart-sass/actions/runs/7351959267/job/20016221691z) and [android amd64](https://github.com/sass/dart-sass/actions/runs/7351959267/job/20016222507). There's also a failure in [the website deploy](https://github.com/sass/dart-sass/actions/runs/7351959267/job/20016232485), which seems like npm isn't making the new package version available quickly enough. Not sure what we can do there other than add some additional wait time.

ntkme:
@nex3 It looks like there might be a change in dart SDK causing cli_pkg to regress... for any ia32 platform it should not attempt to generate aot-snapshot...

ntkme:
@nex3 I think the issue was this change: https://github.com/google/dart_cli_pkg/commit/3c5a2ccf0af48f128596c08a7da4081fbacc2e13

nex3:
Oh, I see... I skimmed https://github.com/dart-lang/sdk/issues/47177 and just noticed the fact that it was closed as completed and my comment that I was testing against the wrong SDK and assumed it was fixed. I'll re-add the exemption.

ntkme:
The previous comment in the cli_pkg was indeed a bit misleading about all 32-bit. Dart SDK do have arm 32-bit aot support, but it never had ia32 aot support (see comments in https://github.com/dart-lang/sdk/issues/49969). In fact, the implementation was correct that it only exempt ia32.

.github/workflows/build-android.yml
larsgrefer: > Have we ever had anyone request native Android support for Dart Sass? I'm not sure this is worth the trouble.

Now that dart-sass has an Android variant, I actually had a request to get my embedded host working on Android:
https://github.com/larsgrefer/dart-sass-java/issues/461#issuecomment-1951485085

PR #2156, opened 2023-12-30
Use implementation name dart-sass for VersionResponse

ntkme:
This is mostly cosmetic.

As documented by https://sass-lang.com/documentation/js-api/variables/info/:

> For Dart Sass, the implementation name is `dart-sass`.

The motivation is that I'm [enhancing the `.info` string for ruby host](https://github.com/sass-contrib/sass-embedded-host-ruby/commit/d6ed0d95e41f8789bd1b0df1f47c4cca5d0e4022), and I want to use the implementation name returned by the VersionResponse and produce an output like below:

```
irb(main):001> require 'sass-embedded';
irb(main):002> puts Sass.info;
sass-embedded	1.69.6	(Embedded Host)	[Ruby]
dart-sass	1.69.6	(Sass Compiler)	[Dart]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.69.8
+
+### Embedded Sass
+
+* The Dart Sass embedded compiler now reports its name as "dart-sass" rather
+  than "Dart Sass", to match the JS API's `info` field.
+
 ## 1.69.7
 
 ### Embedded Sass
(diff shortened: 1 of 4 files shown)


nex3:
I'm a little worried that this is technically a breaking change (although _hopefully_ not one that affects much, since there's only one embedded compiler implementation at the moment). I'm not sure it's strictly an improvement, either; "Dart Sass" is generally a more appropriate name for human presentation. If this is desired, I think I'd be more comfortable adding a new `identifier` field to the protocol that's required to be a valid CSS identifier and leaving `name` as human-readable.

ntkme:
@nex3 I checked some known host implementations and no one depends on this field right now. I feel we can just change it instead of making protocol change which is more work.

- https://github.com/ahabhgk/sass-embedded-host-rust
- https://github.com/bep/godartsass
- https://github.com/charitybell/ESass
- https://github.com/larsgrefer/dart-sass-java
- https://github.com/sass/embedded-host-node
- https://github.com/sass-contrib/sass-embedded-host-ruby

nex3 (approved):
All right, if no one's using it I suppose this works.

PR #2157, opened 2023-12-30
check in the dart pub lockfile in tree

selfisekai:
this package provides an executable, and as such should contain the lockfile


ntkme:
https://github.com/sass/dart-sass/pull/1948#issuecomment-1539185332

selfisekai:
that goes against distro policies that are in place so we know what we're actually building and shipping to users. lockfiles include checksums of the dependencies used

ntkme:
I don't think check this in is the right solution, as it adds more problem to development and testing.

A better solution would be to add a `pubspec.lock` as release artifact. The release procedure would be:

1. Generate a `pubspec.lock`, upload as the action artifact.
2. Download from action artifact, build for each architecture with the same `pubspec.lock`.
3. Upload `pubspec.lock` as a release artifact.

nex3:
@selfisekai What's the specific goal of having a `pubspec.lock` file? Is it to ensure you're using some notion of "the same" dependency versions as a given Dart Sass release? Because not all installation methods use those same versions—anyone installing via pub or Homebrew are only guaranteed a _compatible_ dependency constellation, per semver. Nor do we do any particular verification of the checksums—we just trust the GitHub Actions VM's HTTPS connection to pub.dev, so I'm not sure they provide any particular value beyond the checksums you'd get by constructing a pubspec.lock manually.

If this is still very important, I suppose there's no harm in uploading a lockfile along with the release, but I'm not sure it'll actually provide any material benefits beyond what you'd get by creating one yourself.

selfisekai:
> What's the specific goal of having a `pubspec.lock` file?

reproducible builds (versions and checksums pinned, so builds are not changed by dependency's new release or a swap). from the most practical side tho, we have to rebuild packages regularly (e.g. with new dart release). a lockfile just limits the possibilities of other, possibly breaking changes

> so I'm not sure they provide any particular value beyond the checksums you'd get by constructing a pubspec.lock manually.

not separately maintaining a >600 lines file that has to be updated on version bumps

ntkme:
To provide a bit more context. @selfisekai a maintainer of alpine-linux. They have they own packaging of dart-sdk and dart-sass that is different from https://github.com/dart-musl/dart and the linux-musl package we ship. 

This is https://github.com/dart-musl/dart:

```
/ # ldd /usr/lib/dart/bin/dart
	/lib/ld-musl-aarch64.so.1 (0xffff8b6f0000)
	libc.musl-aarch64.so.1 => /lib/ld-musl-aarch64.so.1 (0xffff8b6f0000)
```

This is dart-sdk from alpine-linux:

```
/ # ldd /usr/bin/dart
	/lib/ld-musl-aarch64.so.1 (0xffffac9f9000)
	libz.so.1 => /lib/libz.so.1 (0xffffaa53e000)
	libicuuc.so.74 => /usr/lib/libicuuc.so.74 (0xffffaa34b000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0xffffaa0a6000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0xffffaa075000)
	libc.musl-aarch64.so.1 => /lib/ld-musl-aarch64.so.1 (0xffffac9f9000)
	libicudata.so.74 => /usr/lib/libicudata.so.74 (0xffffaa055000)
```

The main difference is that https://github.com/dart-musl/dart is optimized for portability that all libraries except musl-libc are bundled. It is the same way like the official dart-sdk that everything except glibc are bundled.

The dart-sdk shipped with alpine-linux is optimized for binary size, that it is dynamically linked with dependencies libraries. The dart-sass package that alpine-linux ships is also optimized for binary size, that the package is effectively only an aot-snapshot and a shell script. The dart-sass package does not ship any dart runtime, as the shell script just runs the dart runtime provided by the dart-sdk package. - Therefore if the dart-sdk package gets a update, the dart-sass package would need a rebuild, because aot-snapshot is sdk version dependent. - I think what @selfisekai what is that each time they rebuild dart-sass because they ship a new sdk version, it would get build with the same dependency.

ntkme:
@selfisekai Question for you is that, what's the issue with the approach that you already have `pubspec.lock` checked-in in aports?

https://git.alpinelinux.org/aports/tree/testing/dart-sass/lock.patch

This should already give you re-producible builds. If your goal is to provide a build that is same as what we've released here in this repo, I have to remind you we don't rebuild dart-sass release when new dart-sdk is available, so even if you have `pubspec.lock` it is not going to be the same.

nex3:
Yeah, that's the thing—I don't think checking in `pubspec.lock` (or uploading it as a build artifact) materially makes builds any more "reproducible" than storing a `pubspec.lock` yourself. It would allow you to use the same constellation of dependencies that we use for _some of_ our releases of a given version of Dart Sass, but even that is contingent (in that we don't necessarily use those dependencies for pub or Homebrew releases) and coincidental (in that we don't consider that set of dependencies canonical and they undergo no special vetting).

selfisekai:
it does not, it's just about the fact of maintaining it separately

nex3:
If this is just for convenience, I'd prefer to avoid checking in the lockfile.

amake:
As a MacPorts port maintainer who developed [Dart pub-related machinery](https://github.com/macports/macports-ports/blob/649cd8ddd814b23e94d39c2dbf034b6c35aeb5e3/_resources/port1.0/group/pub-1.0.tcl) for MacPorts, I also found the lack of a pubspec.lock for this project to be a blocker from meeting our goals.

- We want to provide builds that are as reproducible as possible. That means that we want the deps locked (binary-level reproducibility is nice to have, but not always feasible and thus not as important)
- We want to be able to enumerate all package deps before build time, so that we can mirror them on our infrastructure (to support older systems e.g. without modern TLS)
- We want to patch/modify/maintain as little as possible, leaving as much as possible to upstream projects
- We want to ship binaries built with the same deps versions as the official binaries, but we can't know what those are unless you publish your lockfile

Fundamentally as the application author, only you know if the application is functioning correctly. Any deviation in dependencies can represent a difference in functionality, and downstream folks can't necessarily know if that's desired or not. Even the official Dart guidance says that applications *should* commit their pubspec.lock:
https://dart.dev/tools/pub/private-files#pubspec-lock

nex3:
You're welcome to store and maintain a lockfile on your end; any version constellation that's mutually compatible and abides by the version ranges outlined here is expected to work.

As I mentioned above, there is no canonical set of "the same deps versions as the official binaries". Different means of installing Dart Sass from different sources may have slightly different dependency versions; Homebrew and Chocolatey both build the executable at install-time, and installing via pub _inherently_ uses the latest versions of dependencies because it's being treated as a library. 

> Fundamentally as the application author, only you know if the application is functioning correctly. Any deviation in dependencies can represent a difference in functionality, and downstream folks can't necessarily know if that's desired or not.

I think this is the wrong way to look at it. We _don't_ know for sure that Sass works with the exact set of package versions we publish against. It's entirely possible that the latest version of `string_scanner` just crashes on inputs larger than 100KB and we just never test that. All we really know is the API that our dependencies _claim to_ support. In any of our releases, we're just relying on those claims to remain accurate, with some basic verification in the form of our tests (which certainly don't represent a comprehensive exercise of the sum total API surface of our dependencies).

One of those claims is that the dependencies follow semver, which is intended to guarantee that any use of the package that's not forbidden by types or documentation will have the same behavior across releases of the same major version number. Relying on this claim to remain accurate is no different in principle than relying on other API guarantees.

I'm not _strictly_ opposed to including a lockfile, but it does represent concretely more toil for my team in terms of constantly fielding dependabot PRs for every minor/patch update to every dependency, so I would like a stronger justification that it's actually providing concrete value.

PR #2167, opened 2024-01-30
Use macos-14 runner instead of macos-latest-xlarge runner

ntkme:
GitHub now offers a free m1 runner for open source: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

This PR switches from the paid runner to the free runner.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -16,9 +16,9 @@ jobs:
         include:
           - arch: x64
             runner: macos-latest
-          # https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/
+          # https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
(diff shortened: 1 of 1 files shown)


.github/workflows/build-macos.yml
```diff
@@ -15,10 +15,10 @@ jobs:
       matrix:
         include:
           - arch: x64
-            runner: macos-latest
-          # https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/
+            runner: macos-13
+          # https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/
```
nex3, line 19: I think we can cut this line now that this matches the format of the other runners.

nex3 (changes requested):

.github/workflows/build-macos.yml
ntkme: Done.

nex3 (approved):

PR #2169, opened 2024-02-07
Add a `--pkg-importer` flag

nex3:
See sass/sass#2739

### DIFF
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -10,6 +10,7 @@ import 'package:pub_semver/pub_semver.dart';
 import 'package:term_glyph/term_glyph.dart' as term_glyph;
 
 import '../../sass.dart';
+import '../importer/node_package.dart';
 import '../io.dart';
 import '../util/character.dart';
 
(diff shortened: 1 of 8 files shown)


lib/src/executable/options.dart
```diff
@@ -218,6 +226,12 @@ final class ExecutableOptions {
   /// The set of paths Sass in which should look for imported files.
   List<String> get loadPaths => _options['load-path'] as List<String>;
 
+  /// The list of built-in importers to use to load `pkg:` URLs.
+  List<Importer> get pkgImporters => [
+        for (var _ in _options['pkg-importer'] as List<String>)
+          NodePackageImporter('.')
```
ntkme, line 232: It might be better if the entry point dir is customizable. E.g. `--node-pkg-importer=/entrypoint/dir` and maybe implicitly use pwd if not directory is not provided.

nex3: This would only ever be relevant when resolving a `pkg:` importer from stdin, at which point I think it's easy enough to just change the CWD of the executable if it _really_ matters. Adding an additional flag for it is pretty heavyweight, so I'd like to avoid it.

ntkme: My concern is that tomorrow if we have another `--pkg-importer=x` and user may want to have both `--pkg-importer=node` and `--pkg-importer=x`. In theory it's fine to have more than one pkg importer, that they should be attempted in order just like other importers. While it is ok to repeat the `--pkg-importer` option, but at the same time, it is a bit confusing whether it means overriding or repeating with different parameters, that's why I thought `--node-pkg-importer` might be better than `--pkg-importer=node`.

ntkme: Or maybe this should be a list option e.g. `--pkg-importer=node,x,y`

nex3: It was intended to be a list option. Having it that way, rather than separate named options, allows users to explicitly specify the *order* of multiple built-in `pkg:` importers, which I think is important.

jathak (approved):

PR #2172, opened 2024-02-13
Use musl support in cli_pkg

ntkme:
This PR uses the native support for building kernel snapshot for musl-linux from cli_pkg 2.8.0, instead of the current complex workaround.


nex3 (approved):

nex3:
Thanks!

PR #2175, opened 2024-02-16
Fix linux musl builds

### LINKED ISSUE #2174: linux-musl 1.71.0 release builds are broken
Somehow they are snapshot build created with glibc dart runtime.

ntkme:
- Closes #2174

### DIFF
--- a/.github/workflows/build-linux-musl.yml
+++ b/.github/workflows/build-linux-musl.yml
@@ -54,9 +54,7 @@ jobs:
                      ghcr.io/dart-musl/dart <<'EOF'
           set -e
           dart pub get
-          dart run grinder pkg-standalone-linux-${{ matrix.arch }}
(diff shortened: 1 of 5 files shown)


ntkme:
I noticed that this is a potential breaking change for downloading script in embedded-host-node: https://github.com/sass/embedded-host-node/blob/6944074bb227660883253d2c57183301dd1bd1d6/tool/prepare-optional-release.ts#L172-L179

This is because we used to have `linux-musl-arm64` and now it's `linux-arm64-musl`.

@nex3 Shall we change cli_pkg or the download script in embedded-host-node?

nex3:
I think we should fix it in the embedded host (which it looks like your PR already does), at least partly because changing `cli_pkg` now would technically be a breaking change.

nex3 (approved):

Issue #2178, opened 2024-02-16
Node package importer requires an entryPointDirectory in ESM


jamesnw:
In addition to #2176, we are seeing that `require.main` is undefined in an ESM context. This means that `@JS("require.main.filename")` is throwing the error `reading "filename" of undefined, which throws before the more helpful error can be thrown-

```
"The Node package importer cannot determine an entry point because
`require.main.filename` is not defined.
Please provide an `entryPointDirectory` to the `NodePackageImporter`."
```

Optional chaining is not supported by `@JS` syntax, and I wasn't able to find an alternate syntax that worked. 

The second more significant issue is that currently `entryPointDirectory` is required in an `ESM` context. Some options-

- import.meta.dirname and filename are added in Node 21- https://nodejs.org/api/esm.html#importmetadirname
- `process.cwd()` seems to be available within the ESM module

I think we had discussed this possibility at one point, but had thought that the fixes made to [`dart_cli_pkg` ](https://github.com/google/dart_cli_pkg/pull/151) would address this as well.

PR #2179, opened 2024-02-18
Check require.main for null before getting require.main.filename

jamesnw:
Addresses #2178. This allows Sass in ESM to throw with the correct error message regarding needing to set an entry point directory, instead of throwing with `reading "filename" of undefined`.


jamesnw:
For reasons not yet clear, this seems to be breaking the existing CommonJS functionality.

jamesnw:
Superceded by #2181

Issue #2199, opened 2024-03-18
update flag doesn't work with absolute source paths


sleeuwen:
Hi,

I've run into a problem when using the `--update` flag with the cli and specifying an absolute path for the source file/directory. It seems to always update the output .css file(s) even if the scss file hasn't changed. This is causing a problem in a project that I maintain which integrates the dart-sass executable in a C#/ASP.NET Core application. This project is using the dart vm version which we download directly from the GitHub release assets.

I've tried the following to come to the conclusion that it has something to do with specifying an absolute path for the source file/folder (I only tested this on Linux):
- Create a .scss file in a folder (doesn't have to contain anything special, can just be valid css)
- Run `sass folder:folder --update`, this works as expected and only updates the css when the scss has changed
- Run `sass /absolute/path/to/folder:/absolute/path/to/folder --update` or `sass /absolute/path/to/folder:folder --update`, this will always update the output .css file
- Run `sass folder:/absolute/path/to/folder --update`, this works as expected and only updates the css when the scss changes.

Interestingly, if I specify a folder as a source and run the command multiple times it also tries to (re)compile the css file that was generated the first time. This succeeds only some of the time and fails with an error other times.

I've tried it on multiple older versions and it seems to happen since version 1.67.0. In version 1.66.1 it does work correctly.

Thanks in advance!

nex3:
In #2077, we (correctly) fixed a bug by making the current importer for a given file unable to load absolute URLs. Now absolute URLs are only ever handled by the set of importers the user passes in explicitly. However, this means that if the user doesn't pass in any importers that can handle absolute `file:` URLs, they can't be loaded at all. Although this is arguably desirable in general, since it means the `importers` and `loadPaths` options are fully authoritative for which absolute URLs can be loaded, it causes several knock-on issues.

The first knock-on issue is the one reported here. In the stylesheet graph we use to determine when updates are necessary, the import cache doesn't have a global `FilesystemImporter`. This means that it doesn't know how to load absolute `file:` URLs (which absolute paths on the CLI get converted into), so they're treated as "new files" and always recompiled.

The second knock-on issue is maybe even worse: stylesheets compiled by path on the CLI can't load absolute `file:` URLs at all through `@use` or `@import`. Users can work around this by passing a `--load-path`, but these loads _should_ work by default.

[...]

PR #2203, opened 2024-03-21
Better handle filesystem importers when load paths aren't necessary

nex3:
See #2199
See sass/sass#3815


lib/src/importer/filesystem.dart
```diff
@@ -5,30 +5,86 @@
...
 
-  /// Creates an importer relative to the current working directory.
-  static final cwd = FilesystemImporter('.');
+  /// A [FilesystemImporter] that loads files relative to the current working
+  /// directory.
+  ///
+  /// Historically, this was the best default for supporting `file:` URL loads
+  /// when the load path didn't matter. However, adding the current working
+  /// directory to the load path wasn't always desirable, so it's no longer
+  /// recommneded. Instead, either use [FilesystemImporter.noLoadPath] if the
```
jathak, line 56:
```suggestion
  /// recommended. Instead, either use [FilesystemImporter.noLoadPath] if the
```

jathak (approved):

PR #2206, opened 2024-03-31
Use thread-safe writeln in dart-sdk 3.3

ntkme:
The thread safety issue of `writeln` has been fixed in https://github.com/dart-lang/sdk/commit/1d3d819ecfa2000ed11a137f5ee80df5bb687819, which is in dart-sdk >3.3.0.


ntkme:
Unfortunately, dart-sdk only fixed thread-safety issue with `print`, but not `writeln`: https://github.com/dart-lang/sdk/issues/55345

PR #2209, opened 2024-04-03
Fix cache for relative imports handled by importers other than the base importer

### LINKED ISSUE #2208: ImportCache of CanonicalizeResult is broken for relative imports handled by any importer that is not a baseImporter
- ~~Deprecate importer without base URL (https://github.com/sass/dart-sass/pull/2213)~~
- [x] Don't cache canonicalizations that might access `containingUrl` (https://github.com/sass/sass/issues/3831)
- [x] Track which canonicalizations _actually_ access `containingUrl` and cache the rest (https://github.com/sass/dart-sass/pull/2220) (https://github.com/sass/dart-sass/pull/2219)
- ~~Drop support for importer without base URL~~

---

See https://github.com/sass/sass-spec/pull/1969/files for reproduction of the issue.

ntkme:
This PR changes it that all imports are cached with key `url, forImport, importer, baseUrl, resolveUrl`, in order to fix cache key conflicts in a few cases.

https://github.com/sass/sass-spec/pull/1969

- Fix #2208

### DIFF
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -40,31 +40,21 @@ final class AsyncImportCache {
 
   /// The canonicalized URLs for each non-canonical URL.
   ///
-  /// The `forImport` in each key is true when this canonicalization is for an
-  /// `@import` rule. Otherwise, it's for a `@use` or `@forward` rule.
+  /// The map's keys have five parts:
   ///
-  /// This cache isn't used for relative imports, because they depend on the
-  /// specific base importer. That's stored separately in
-  /// [_relativeCanonicalizeCache].
-  final _canonicalizeCache =
-      <(Uri, {bool forImport}), AsyncCanonicalizeResult?>{};
-
-  /// The canonicalized URLs for each non-canonical URL that's resolved using a
-  /// relative importer.
-  ///
-  /// The map's keys have four parts:
-  ///
-  /// 1. The URL passed to [canonicalize] (the same as in [_canonicalizeCache]).
+  /// 1. The URL passed to [_canonicalize] (the same as in [_canonicalizeCache]).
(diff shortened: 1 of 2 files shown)


lib/src/async_import_cache.dart
```diff
@@ -186,13 +198,10 @@ final class AsyncImportCache {
   /// If [resolveUrl] is `true`, this resolves [url] relative to [baseUrl]
   /// before passing it to [importer].
   Future<AsyncCanonicalizeResult?> _canonicalize(
-      AsyncImporter importer, Uri url, Uri? baseUrl, bool forImport,
-      {bool resolveUrl = false}) async {
```
ntkme, line 190: Removing `{bool resolveUrl = false}` is never used.

lib/src/async_import_cache.dart
```diff
@@ -153,18 +153,30 @@ final class AsyncImportCache {
...
-          () => _canonicalize(baseImporter, baseUrl?.resolveUri(url) ?? url,
-              baseUrl, forImport));
-      if (relativeResult != null) return relativeResult;
+    if (url.scheme == '') {
+      if (baseImporter != null) {
+        var resolved = baseUrl?.resolveUri(url) ?? url;
+        var relativeResult = await putIfAbsentAsync(
+            _relativeCanonicalizeCache,
+            (
+              resolved,
```
ntkme, line 162: This cache key need to be `resolved`, despite that in some cases this is no longer "relative".

Because it's possible that the "second" importer is actually the same as baseImporter (which is used in embedded-host to emulate legacy API). We need to make sure baseImporter's cache key contains the actual attempted url, rather than original url, so that the original url can be attempted one more time in the "second" importer.

PR #2213, opened 2024-04-09
*ImportCache.canonicalize: Deprecate base importer without URL

nex3:
See #2208
See sass/sass#3832
See sass/sass-spec#1971

[skip sass-embedded]

### DIFF
--- a/lib/src/syntax.dart
+++ b/lib/src/syntax.dart
@@ -2,6 +2,7 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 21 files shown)


lib/src/syntax.dart
```diff
@@ -27,6 +28,16 @@ enum Syntax {
...
   final String _name;
 
+  /// The file extension used for this syntax, including the period.
+  ///
+  /// :nodoc:
+  @internal
+  String get extension => switch (this) {
+        Syntax.sass => '.sass',
+        Syntax.scss => '.scss',
+        Syntax.css => 'css'
```
stof, line 38: shouldn't this be `.css` ?

nex3:
Per https://github.com/sass/dart-sass/issues/2208#issuecomment-2046013595, we've decided not to deprecate this after all.

PR #2214, opened 2024-04-09
Make it possible to build npm with a linked language repo

nex3:
This makes several changes:

* It renames the `UPDATE_SASS_PROTOCOL` environment variable used by
  the Grinder `protobuf` task to `UPDATE_SASS_SASS_REPO` to make it
  more generic and so usable by other tasks. The previous name still
  works but is considered deprecated.

* The `pkg-npm-*` grinder tasks now respects the
  `UPDATE_SASS_SASS_REPO` environment variable. This allows repos to
  ensure that the linked language repo's version of the TypeScript
  types are used when building the npm package.

* `UPDATE_SASS_SASS_REPO=false` is set for all the `pkg-npm-*` tasks
  run by this repo, so that they will use the linked language repo's
  version of the TypeScript types.


tool/grind.dart
```diff
@@ -251,10 +251,7 @@ dart run protoc_plugin "\$@"
     run('chmod', arguments: ['a+x', 'build/protoc-gen-dart']);
   }
 
-  if (Platform.environment['UPDATE_SASS_PROTOCOL'] != 'false') {
-    cloneOrCheckout("https://github.com/sass/sass.git", "main",
-        name: 'language');
-  }
+  _updateLanguageRepo();
```
Goodwine, line 254: The function returns a value, but it's not being used. Is that intended?

nex3: Oops, missed a usage. Good catch!

PR #2215, opened 2024-04-09
Don't cache canonicalize calls when `containingUrl` is available

nex3:
See #2208

### DIFF
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -174,6 +174,12 @@ final class AsyncImportCache {
     var key = (url, forImport: forImport);
     if (_canonicalizeCache.containsKey(key)) return _canonicalizeCache[key];
 
+    // Each indivudal call to a `canonicalize()` override may not be cacheable
+    // (specifically, if it has access to `containingUrl` it's too
+    // context-sensitive to usefully cache). We want to cache a given URL across
+    // the _entire_ importer chain, so we use [cacheable] to track whether _all_
+    // `canonicalize()` calls we've attempted are cacheable. Only if they are do
+    // we store the result in the cache.
     var cacheable = true;
     for (var importer in _importers) {
       switch (await _canonicalize(importer, url, baseUrl, forImport)) {
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -154,64 +154,79 @@ final class AsyncImportCache {
     }
 
     if (baseImporter != null && url.scheme == '') {
-      var relativeResult = await putIfAbsentAsync(
-          _relativeCanonicalizeCache,
-          (
-            url,
-            forImport: forImport,
-            baseImporter: baseImporter,
-            baseUrl: baseUrl
-          ),
-          () => _canonicalize(baseImporter, baseUrl?.resolveUri(url) ?? url,
-              baseUrl, forImport));
+      var relativeResult = await putIfAbsentAsync(_relativeCanonicalizeCache, (
+        url,
+        forImport: forImport,
+        baseImporter: baseImporter,
+        baseUrl: baseUrl
+      ), () async {
+        var (result, cacheable) = await _canonicalize(
+            baseImporter, baseUrl?.resolveUri(url) ?? url, baseUrl, forImport);
+        assert(
+            cacheable,
+            "Relative loads should always be cacheable because they never "
+            "provide access to the containing URL.");
+        return result;
+      });
       if (relativeResult != null) return relativeResult;
     }
 
-    return await putIfAbsentAsync(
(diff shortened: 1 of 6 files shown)


nex3:
@ntkme

lib/src/async_import_cache.dart
```diff
@@ -154,64 +154,79 @@ final class AsyncImportCache {
...
       if (relativeResult != null) return relativeResult;
     }
 
-    return await putIfAbsentAsync(
-        _canonicalizeCache, (url, forImport: forImport), () async {
-      for (var importer in _importers) {
-        if (await _canonicalize(importer, url, baseUrl, forImport)
-            case var result?) {
+    var key = (url, forImport: forImport);
+    if (_canonicalizeCache.containsKey(key)) return _canonicalizeCache[key];
```
ntkme, line 175: I can think of an edge case with two importers `[importerA, importerB]`:

1. Resolve an import `x`, with containing url `u:y`. `importerA` accessed `containingUrl`, but did not resolve it, then `importerB` resolved it without using `containingUrl` thus get cached.
2. Resolve an import `x`, with containing url `file://a.scss`. `importerA` would have been able to resolve it because now this has a different `containingUrl` that `importerA` can handle, however, because of the previous cache from `importerB`, `importerA` would not even get attempted.

I think this would be unexpected, because `importerA` is before `importerB`.

nex3: In this case, `_canonicalize()` would return `(null, false)` for because `containingUrl` is passed, so `cacheable` will be set to `false` on line 188 and `importerB`'s result won't be cached even if it would be cacheable in isolation.

ntkme: The expectation here is that the result with cache should be exactly the same as if we completely remove the cache capability, thus the order of importers would matter. Implement the order meaning we need cache per importer here, which would can have performance impact if a non-cacheable importer has high priority and low hit rate...

However, for "correctness", it seems to be unavoidable.

ntkme: Sorry, I forgot this is the first iteration that cacheable is based on whether it's passed or not, instead of whether it's used or not.

nex3: The logic should work the same if it just checks access—we'd still return `(null, true)` in the case where `importerA` accessed `containingUrl` but couldn't resolve the import.

ntkme (approved):

lib/src/async_import_cache.dart
ntkme: I think this works correctly.

If I read it correctly, any non-cacheable importer will make all importer after it non-cacheable for the current import. And therefore if a FilesystemImporter is low priority it would not be cached and cost can be high? If that’s the case it might still be worth to have per importer cache key instead of single cache key for all importers.

ntkme: If we have `[nonCacheableImporter, loadPathImporter]`, with the implementation in this PR the `loadPathImporter` would not be cached, and repeatedly loading the same file from load path can get slower due to repeated I/O calls.

Thus I think it's worth to make the trade-off to use a little bit more memory to cache per importer (and even cache failed canonicalization if cacheable), and a bit more CPU to re-check cache per importer, like the pseudo code below:  

``` dart
for (var importer in importers) {
    var key = (url, forImport: forImport, importer: importer);
    if (_canonicalizeCache.containsKey(key)) {
      var cached = _canonicalizeCache[key];
      if (cached != null) return cached;
    } else {
      var (result, cacheable) = await _canonicalize(importer, url, baseUrl, forImport);
      if (cacheable) {
        _canonicalizeCache[key] = result;
      }
      if (result != null) return result;
    }
}
return null;
```

The cost of CPU/memory overhead would likely be lower than the I/O overhead.

nex3: I don't think it's crazy to have a per-importer canonicalize cache in addition to the current whole-load-path cache, although I think we might want to be a bit more sophisticated about only filling it if we run into an uncacheable load in practice. Either way, let's save that for a follow-up after we fix the initial bug.

lib/src/async_import_cache.dart
```diff
@@ -154,64 +154,79 @@ final class AsyncImportCache {
...
+
+    var cacheable = true;
+    for (var importer in _importers) {
```
Goodwine, line 178: Nit: I was able to follow the caching logic, but it wasn't very obvious at the beginning. It would be great if the caching strategy was documented, but at the same time the documenting the strategy also feels like documenting an implementation detail so maybe it doesn't have to be documented at all 🤔.

Goodwine (approved):

lib/src/async_import_cache.dart
nex3: Added a comment to explain in more detail

PR #2218, opened 2024-04-12
Quote channel names in color functions

nex3:
See sass/sass-spec#1975

### DIFF
--- a/lib/src/functions/color.dart
+++ b/lib/src/functions/color.dart
@@ -448,11 +448,8 @@ final module = BuiltInModule("color", functions: <Callable>[
       r"$color, $channel",
       (arguments) => SassBoolean(arguments[0]
           .assertColor("color")
-          .isChannelMissing(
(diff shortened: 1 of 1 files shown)


lib/src/functions/color.dart
```diff
@@ -475,7 +475,10 @@ final module = BuiltInModule("color", functions: <Callable>[
 
   _function("channel", r"$color, $channel, $space: null", (arguments) {
     var color = _colorInSpace(arguments[0], arguments[2]);
-    var channelName = arguments[1].assertString("channel").text.toLowerCase();
+    var channelName = (arguments[1].assertString("channel")
```
Goodwine, line 478: Nit: This pattern is used in one other place besides these two new uses (although the other one doesn't transform the string to lowercase). This can be extracted to a helper function, which may be useful if the same has to be repeated more times 🤔

Goodwine (approved):
small suggestion, but otherwise LGTM

PR #2219, opened 2024-04-12
Add a per-importer cache for loads that aren't cacheable en masse

nex3:
Inspired by comments on #2215

### DIFF
--- a/lib/src/async_import_cache.dart
+++ b/lib/src/async_import_cache.dart
@@ -43,30 +43,20 @@ final class AsyncImportCache {
   /// The `forImport` in each key is true when this canonicalization is for an
   /// `@import` rule. Otherwise, it's for a `@use` or `@forward` rule.
   ///
-  /// This cache isn't used for relative imports, because they depend on the
-  /// specific base importer. That's stored separately in
-  /// [_relativeCanonicalizeCache].
+  /// This cache covers loads that go through the entire chain of [_importers],
+  /// but it doesn't cover individual loads or loads in which any importer
+  /// accesses `containingUrl`. See also [_perImporterCanonicalizeCache].
   final _canonicalizeCache =
       <(Uri, {bool forImport}), AsyncCanonicalizeResult?>{};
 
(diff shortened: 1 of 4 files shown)


nex3:
@ntkme

lib/src/async_import_cache.dart
```diff
@@ -181,17 +168,34 @@ final class AsyncImportCache {
...
+          _perImporterCanonicalizeCache[(importer, url, forImport: forImport)] =
+              result;
+          if (result != null) return result;
+
+        case (var result, false):
+          if (cacheable) {
+            // If this is the first uncacheable result, add all previous results
+            // to the per-importer cache so we don't have to re-run them for
+            // future uses of this importer.
+            for (var j = 0; i < j; j++) {
```
ntkme, line 188:
```suggestion
            for (var j = 0; j < i; j++) {
```

Goodwine: nice catch!

lib/src/async_import_cache.dart
```diff
@@ -181,17 +168,34 @@ final class AsyncImportCache {
     // `canonicalize()` calls we've attempted are cacheable. Only if they are do
     // we store the result in the cache.
     var cacheable = true;
-    for (var importer in _importers) {
+    for (var i = 0; i < _importers.length; i++) {
+      var importer = _importers[i];
```
Goodwine, line 172: There should be a check around here for the presence of `_perImporterCanonicalizeCache`, no? 

Otherwise `_canonicalize` gets executed for all the importers previous to the one that returned a non-null and non-cacheable result

Goodwine (approved):

lib/src/async_import_cache.dart
nex3: Yeah, actually checking the cache is an important part of a caching scheme :sweat_smile:

Goodwine (approved):

nex3:
@ntkme Can you take another look at this?

ntkme (approved):

PR #2220, opened 2024-04-12
Implement access tracking for containingUrl

ntkme:
This PR adds a `CanonicalizeContext` class in dart which is able to track if `containingUrl` is accessed or not, it also changes the dart2js `JSCanonicalizeContext` class to be a wrapper, so that the containingUrl in JS would lazily access the dart `containingUrl` method instead of eagerly during construction. Finally for embedded host, this adds a `containingUrlWithoutMarking` method that allow the serialization of embedded protocol without marking, and we can mark it accordingly depends on the result from host. 

https://github.com/sass/sass/pull/3835
https://github.com/sass/embedded-host-node/pull/285

### DIFF
--- a/lib/src/embedded/importer/file.dart
+++ b/lib/src/embedded/importer/file.dart
@@ -21,8 +21,8 @@ final class FileImporter extends ImporterBase {
       ..importerId = _importerId
       ..url = url.toString()
       ..fromImport = fromImport;
-    var containingUrl = canonicalizeContext.containingUrlWithoutMarking;
-    if (containingUrl case var containingUrl?) {
+    if (canonicalizeContext.containingUrlWithoutMarking
+        case var containingUrl?) {
       request.containingUrl = containingUrl.toString();
     }
     var response = dispatcher.sendFileImportRequest(request);
--- a/lib/src/embedded/importer/file.dart
+++ b/lib/src/embedded/importer/file.dart
@@ -21,10 +21,12 @@ final class FileImporter extends ImporterBase {
       ..importerId = _importerId
       ..url = url.toString()
       ..fromImport = fromImport;
+    var containingUrl = canonicalizeContext.containingUrlWithoutMarking;
     if (containingUrl case var containingUrl?) {
       request.containingUrl = containingUrl.toString();
     }
     var response = dispatcher.sendFileImportRequest(request);
+    if (!response.containingUrlUnused) canonicalizeContext.containingUrl;
 
     switch (response.whichResult()) {
       case InboundMessage_FileImportResponse_Result.fileUrl:
--- a/lib/src/import_cache.dart
+++ b/lib/src/import_cache.dart
@@ -5,7 +5,7 @@
 // DO NOT EDIT. This file was generated from async_import_cache.dart.
 // See tool/grind/synchronize.dart for details.
 //
-// Checksum: 37dd173d676ec6cf201a25b3cca9ac81d92b1433
+// Checksum: 36bc42050cf2eb3a43f36376c4f06c1708eee777
 //
 // ignore_for_file: unused_import
 
@@ -18,6 +18,7 @@ import 'package:path/path.dart' as p;
 import 'ast/sass.dart';
 import 'deprecation.dart';
 import 'importer.dart';
+import 'importer/canonicalize_context.dart';
 import 'importer/no_op.dart';
 import 'importer/utils.dart';
 import 'io.dart';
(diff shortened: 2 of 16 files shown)


lib/src/embedded/importer/file.dart
```diff
@@ -21,10 +21,12 @@ final class FileImporter extends ImporterBase {
       ..importerId = _importerId
       ..url = url.toString()
       ..fromImport = fromImport;
-    if (containingUrl case var containingUrl?) {
+    var containingUrl = canonicalizeContext.containingUrlWithoutMarking;
+    if (containingUrl != null) {
```
nex3, line 25: I'd prefer to keep using `case var` here.

lib/src/importer/async.dart
```diff
@@ -54,7 +55,16 @@ abstract class AsyncImporter {
...
+  Uri? get containingUrl => utils.canonicalizeContext.containingUrl;
+
+  /// The canonicalize context of the stylesheet that caused the current
+  /// [canonicalize] invocation.
+  ///
+  /// Subclasses should only access this from within calls to [canonicalize].
+  /// Outside of that context, its value is undefined and subject to change.
+  @protected
+  @nonVirtual
+  CanonicalizeContext get canonicalizeContext => utils.canonicalizeContext;
```
nex3, line 67: I think we should make this `@internal` and `:nodoc`. We don't currently have any use case for allowing end-users to access it directly.

lib/src/importer/canonicalize_context.dart
```diff
@@ -0,0 +1,24 @@
+// Copyright 2014 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import 'package:meta/meta.dart';
+
+class CanonicalizeContext {
```
nex3, line 7: Please document this and its members. Also, it should be `final class` and marked as `@internal`.

lib/src/importer/utils.dart
```diff
@@ -15,16 +16,18 @@ import '../io.dart';
 /// canonicalization should be identical for `@import` and `@use` rules. It's
 /// admittedly hacky to set this globally, but `@import` will eventually be
 /// removed, at which point we can delete this and have one consistent behavior.
-bool get fromImport => Zone.current[#_inImportRule] as bool? ?? false;
+bool get fromImport => (Zone.current[#_inImportRule] as bool? ??
+    (Zone.current[#_canonicalizeContext] as CanonicalizeContext?)?.fromImport ??
+    false);
```
nex3, line 21: It seems like we should either have `CanonicalizeContext` or `Zone.current[#_inImportRule]` be the sole source of truth for whether we're currently evaluating an import rule, not both. The easiest answer is probably to just have `CanonicalizeContext` _only_ hold `containingUrl`, although you could also create a new `CanonicalizeContext` for `inImportRule()` instead.

lib/src/js/utils.dart
```diff
@@ -197,6 +200,25 @@ Uri jsToDartUrl(JSUrl url) => Uri.parse(url.toString());
 /// Converts a Dart [Uri] object to a standard JS `URL` object.
 JSUrl dartToJSUrl(Uri url) => JSUrl(url.toString());
 
+/// The use of  `@JSExport()` together with `createDartExport` allow us
+/// to create a JSObject that lazily access containingUrl
+@JSExport()
+class _JSExportCanonicalizeContext {
```
nex3, line 206: This should be in its own file, probably in a `lib/src/js/importer` directory. Also, I'm not sure `@JSExport()` is better than doing the same thing we do for value types and editing the class structure so that we expose JS-friendly members on an existing Dart type. At the very least, this will involve more allocations, which are nice to avoid when possible.

nex3 (changes requested):

lib/src/import_cache.dart
```diff
@@ -206,18 +207,17 @@ final class ImportCache {
...
+        baseUrl != null &&
+                (url.scheme == '' || importer.isNonCanonicalScheme(url.scheme))
+            ? baseUrl
+            : null,
+        forImport);
+
+    var result = withCanonicalizeContext(
+        canonicalizeContext, () => importer.canonicalize(url));
+
+    var cacheable = !canonicalizeContext.wasContainingUrlAccessed;
```
nex3, line 220: Consider treating the containing URL as not accessed if `containingUrlAvailable: url.scheme != '' && !importer.isNonCanonicalScheme(url.scheme)`, since in that case it's not even accessible to the importer. That would make us a little more robust against importers that access `containingUrl` more aggressively than they should.

lib/src/js/utils.dart
ntkme: I've split the file out. Conceptually I understand what you're saying here, but I'm having a hard time dealing with js interop. I'm aware that this `@JSExport()` will create additional delegate objects, and I settled with it only because it's relatively easy to understand.

@nex3 I'd like to get some help from you on this if you have a way to achieve the same without the delegate.

lib/src/importer/utils.dart
ntkme: I moved it all into `CanonicalizeContext` because I want to have only one zoned variable. Having both `fromImport` and `containingUrl` maintained in a single place is more appealing, because we can just cast it to a `JSCanonicalizeContext` and have `containingUrl` still being a lazy getter in JS. Unfortunately, I'm struggling with the JS casting part that I would like to have your help to deal with `@JSExport` as we discussed in the other thread.

nex3 (changes requested):

lib/src/importer/canonicalize_context.dart
```diff
@@ -0,0 +1,49 @@
...
+  /// @nodoc
+  @internal
+  bool get wasContainingUrlAccessed => _wasContainingUrlAccessed;
+  var _wasContainingUrlAccessed = false;
+
+  /// Runs [callback] in a context with specificed [fromImport].
+  ///
+  /// @nodoc
+  @internal
+  T withFromImport<T>(bool fromImport, T callback()) {
```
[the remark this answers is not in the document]
ntkme: Due to the access tracking of containingUrl, `CanonicalizeContext` by nature is already mutable, and if we fork another copy to modify import, it will make the access tracking complicated as the fork need to propagate the access tracking to the original copy.

In the current implementation, `CanonicalizeContext` is only initialized as zone variable, and `withFromImport` is only called on within zones, so that the mutation should be safe in practice. I understand there is risk if it is called un-zoned, so I added an assertion to make this method is only callable when the CanonicalizeContext being invoked is the current context inside the current zone.

nex3 (approved):

Issue #2197, opened 2024-03-16
Drop support for plain-CSS-like functions and mixins


nex3:
See https://github.com/sass/sass/issues/3787

PR #2230, opened 2024-04-27
Deprecate function and mixin names beginning with `--`

nex3:
See #2197
See sass/sass#3854
See sass/sass-spec#1982
[skip sass-embedded]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
-## 1.75.1
+## 1.76.0
 
 * Throw errors for misplaced statements in keyframe blocks.
+
+* Mixins and functions whose names begin with `--` are now deprecated for
+  forwards-compatibility with the in-progress CSS functions and mixins spec.
+  This deprecation is named `css-function-mixin`.
 
 ## 1.75.0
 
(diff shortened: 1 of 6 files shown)


Goodwine (approved):

WebMechanic:
Hello,

I updated to 1.77.0 and noticed this also affects mixins starting with **double underscores** such as `@mixin __fubar()` and their `@include __fubar()`.

If this is intended it should probably be mentioned on the [documentation](https://sass-lang.com/documentation/breaking-changes/css-function-mixin/)

We only had a couple of old mixins using the `__` prefix to mark them as "private". No big deal as the warning in the console was very helpful.

Bye.

Goodwine:
It was intended. Underscores and hyphens are equivalent in Sass identifiers. We assume users are familiar with that fact, but I understand it may not be obvious.

From the docs:

> Mixin names, like all Sass identifiers, treat hyphens and underscores as identical. This means that `reset-list` and `reset_list` both refer to the same mixin.

https://sass-lang.com/documentation/at-rules/mixin/

WebMechanic:
Thanks, indeed not so obvious :-)
From this example with the **inline** hyphens and underscores, I had never deduced that this also applies to **prefixes** which most languages handle very differently -- except CSS and as I learned now, Sass.
It's always these "little things" that cause confusion :-)

nex3:
I think @WebMechanic is right that this is a bug. The spec is ambiguous here, and there's not a strong reason for us to forbid functions from being written and referenced with double leading underscores.

PR #2232, opened 2024-05-03
[Color 4] Update `color.same()` and color equality

nex3:
See sass/sass#3852
See sass/sass#3858
See sass/sass-spec#1984

### DIFF
--- a/lib/src/functions/color.dart
+++ b/lib/src/functions/color.dart
@@ -504,12 +504,37 @@ final module = BuiltInModule("color", functions: <Callable>[
     var color1 = arguments[0].assertColor('color1');
     var color2 = arguments[1].assertColor('color2');
 
-    // Convert both colors into the same space to compare them. Usually we
-    // just use color1's space, but since HSL and HWB can't represent
-    // out-of-gamut colors we use RGB for all legacy color spaces.
-    var targetSpace = color1.isLegacy ? ColorSpace.rgb : color1.space;
-    return SassBoolean(
-        color1.toSpace(targetSpace) == color2.toSpace(targetSpace));
+    /// Converts [color] to the xyz-d65 space without any mising channels.
+    SassColor toXyzNoMissing(SassColor color) => switch (color) {
+          SassColor(space: ColorSpace.xyzD65, hasMissingChannel: false) =>
+            color,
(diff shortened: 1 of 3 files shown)


lib/src/util/number.dart
```diff
@@ -30,6 +30,17 @@ bool fuzzyEquals(num number1, num number2) {
...
+/// Like [fuzzyEquals], but allows null values for [number1] and [number2].
+///
+/// null values are only equal to one another.
+bool fuzzyEqualsNullable(num? number1, num? number2) {
+  if (number1 == number2) return true;
+  if (number1 == null || number2 == null) return false;
+  return (number1 - number2).abs() <= _epsilon &&
+      (number1 * _inverseEpsilon).round() ==
+          (number2 * _inverseEpsilon).round();
+}
```
Goodwine, line 42: Nit: To avoid possible divergence, `fuzzyEqualsNullable` could call `fuzzyEquals` after line 38

Goodwine (approved):

lib/src/util/number.dart
nex3: That would duplicate the `number1 == number2` test, which isn't super relevant but for code this low-level I'd like to keep it as efficient as possible since it's likely to be run very frequently.

lib/src/functions/color.dart
```diff
@@ -504,12 +504,37 @@ final module = BuiltInModule("color", functions: <Callable>[
...
+                ColorSpace.xyzD65, channel0, channel1, channel2, alpha)
+        };
+
+    return SassBoolean(color1.space == color2.space
+        ? fuzzyEquals(color1.channel0, color2.channel0) &&
+            fuzzyEquals(color1.channel1, color2.channel1) &&
+            fuzzyEquals(color1.channel2, color2.channel2) &&
+            fuzzyEquals(color1.alpha, color2.alpha)
+        // Use [ColorSpace.convert] manually so that we can convert missing
+        // channels to 0 without having to create new intermediate color objects.
```
stof, line 536: shouldn't this comment be near line 526 instead as this is where `ColorSpace.convert` is used ?

Goodwine (approved):

PR #2235, opened 2024-05-06
Deprecate the `feature-exists` function

stof:
See https://github.com/sass/sass/issues/3702
https://github.com/sass/sass/pull/3860
https://github.com/sass/sass-spec/pull/1986

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.78.0
 
+* The `meta.feature-exists` function is now deprecated. This deprecation is
+  named `feature-exists`.
+
 ### JS API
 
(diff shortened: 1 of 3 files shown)


pkg/sass_api/pubspec.yaml
nex3, line 1: You'll need to update the CHANGELOG for this package as well.

stof:
@nex3 I updated this PR as well as the related PR for sass-spec and the language repo

pkg/sass_api/pubspec.yaml
stof: not needed right now for the update, as it already contains the info about the upcoming 1.78.0 release (as there are already other pending changes)

nex3 (approved):

PR #2246, opened 2024-05-15
Fix deprecation flags in the CLI and add tests

### LINKED ISSUE #2244: Leave invalid CSS after compilation option
Hi, is it possible to leave invalid CSS after running sass compilation? Sass cli command is only part of our styling pipe, unfortunately one that have to be process at the beginning. We're making use of "deep selector" (vuejs term) which looks like this "& >>> li", and is processed by different software. When running sass cli this part of css is cut with message: "The selector "& > > > li" is invalid CSS. It will be omitted from the generated CSS.
This will be an error in Dart Sass 2.0.0". 

Can we somehow take this problem? I tried to use --silence-deprecation=bogus-combinators but without success.

nex3:
Closes #2244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.77.2
+
+### Command-Line Interface
+
+* Properly handle the `--silence-deprecation` flag.
+
+* Handle the `--fatal-deprecation` and `--future-deprecation` flags for
+  `--interactive` mode.
+
 ## 1.77.1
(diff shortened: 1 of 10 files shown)


jathak (approved):

nex3:
Holding off on a release until https://github.com/sass/dart-sass/pull/2247 lands as well.

UtoywilsonTv (approved):

PR #2248, opened 2024-05-18
Parse Deprecation.forVersion on compiler side

ntkme:
https://github.com/sass/sass/pull/3868
https://github.com/sass/embedded-host-node/pull/299


jathak (approved):
Can you bump the pubspec versions and add changelog entries?

PR #2251, opened 2024-05-21
[Color 4] Update behavior to match updated spec

nex3:
See sass/sass#3869
See sass/sass-spec#1995


jathak (approved):

nex3:
Dartdoc seems to be failing due to https://github.com/dart-lang/dartdoc/issues/3818

PR #2253, opened 2024-05-21
Generate deprecations list from the language repo

jathak:
See https://github.com/sass/sass/pull/3872.

This updates the `Deprecation` enum to be generated from `spec/deprecations.yaml` in the language repo.


.gitignore
```diff
@@ -15,5 +15,6 @@ node_modules/
 /doc/api
 /pkg/*/doc/api
 
-# Generated protocol buffer files.
+# Generated Dart files.
 *.pb*.dart
+*.g.dart
```
nex3, line 20: This is a little odd, since the async files are also generated but _don't_ have a marked filename. Might it be simpler to just explicitly ignore the `deprecation.dart` file?

tool/grind/generate_deprecations.dart
```diff
@@ -0,0 +1,66 @@
...
+// https://opensource.org/licenses/MIT.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:crypto/crypto.dart';
+import 'package:dart_style/dart_style.dart';
+import 'package:yaml/yaml.dart';
+
+void updateDeprecationFile(File yamlFile) {
```
nex3, line 12: It would be more consistent with other libraries to define the task here directly.

tool/grind/generate_deprecations.dart
```diff
@@ -0,0 +1,66 @@
...
+  var yamlText = yamlFile.readAsStringSync();
+  var data = loadYaml(yamlText) as Map;
```
nex3, line 14: Pass in the source URL here as well.

nex3 (approved):

.gitignore
jathak: I was thinking in terms of having something to distinguish generated code that's not checked in, and I'd seen the `.g.dart` pattern with some other Dart projects, but maybe since `deprecation.dart` is the only file like that here (other than the protobuf files, which use their own extension) just ignoring it explicitly is better.

test/double_check_test.dart
```diff
@@ -31,6 +33,16 @@ void main() {
...
 
+  test("deprecations are up-to-date", () {
+    var inputText = File(deprecations.yamlPath).readAsStringSync();
+    var outputText = File(deprecations.dartPath).readAsStringSync();
+    var checksum = sha1.convert(utf8.encode(inputText));
+    if (!outputText.contains('// Checksum: $checksum')) {
+      fail('${deprecations.dartPath} is out-of-date.\n'
+          'Run `dart run grinder` to update it.');
+    }
+  }, testOn: "!windows");
```
nex3, line 44: Probably worth including the above comment here as well, or moving this condition onto a group that contains both up-to-date tests.

tool/grind/deprecation.dart.template
nex3, line 1: Rather than having a template and a source file, could it make sense to just allow the source file to be edited directly and only have a specific subsection of it get (re)generated?

tool/grind/double_check.dart
```diff
@@ -21,7 +20,7 @@ Future<void> doubleCheckBeforeRelease() async {
     fail("GITHUB_REF $ref is different than pubspec version ${pkg.version}.");
   }
 
-  if (listEquals(pkg.version.preRelease, ["dev"])) {
+  if (const ListEquality<Object?>().equals(pkg.version.preRelease, ["dev"])) {
```
nex3, line 23: Why doesn't `listEquals()` work here anymore?

jathak: It's defined within `lib/src/utils.dart`, which depends on the Sass AST, so we can't import it here

test/double_check_test.dart
jathak: Done

tool/grind/deprecation.dart.template
jathak: Yeah, good point. Now that the deprecations file is checked in, there's no reason not to just generate a section of it. Fixed.

PR #2257, opened 2024-05-31
Regenerate protobuf as part of default task

nex3:


jathak (approved):

ntkme:
I'm afraid that this will break linux/android release pipelines. For example, for android it builds protobuf on `dart:latest` container and then build the binary on `ghcr.io/dart-android/dart:latest` container. The reason is that there is simply no prebuilt binary of buf for android.

nex3:
I don't think those tasks use the default `pub run grinder` task, I think they explicitly call out to the tasks they need. The default task is really meant for human convenience.

PR #2258, opened 2024-06-06
Enable AOT build for linux-riscv64-musl

ntkme:
Alpine 3.20 has added stable support of `riscv64`, and I started to publish containers for `linux-riscv64-musl` for `dart >=3.4.3`: https://github.com/dart-musl/dart/pkgs/container/dart


nex3 (approved):
Thanks for keeping on top of this!

PR #2264, opened 2024-06-15
Fix a bug where comments were incorrectly forbidden in some cases

nex3:
See sass/sass-spec#1998


pkg/sass_api/pubspec.yaml
```diff
@@ -2,15 +2,15 @@ name: sass_api
 # Note: Every time we add a new Sass AST node, we need to bump the *major*
 # version because it's a breaking change for anyone who's implementing the
 # visitor interface(s).
-version: 10.4.5
+version: 10.4.6
```
Goodwine, line 5: per the lint error, I think this is missing a changelog update

PR #2265, opened 2024-06-17
Fix linux-ia32, linux-arm-musl, and windows-arm64 releases

ntkme:
@nex3 Sorry that my recent commit broken the release CI. This PR should fix it.

https://github.com/sass/dart-sass/actions/runs/9554504858/job/26336030878

### DIFF
--- a/.github/workflows/build-linux-musl.yml
+++ b/.github/workflows/build-linux-musl.yml
@@ -19,7 +19,7 @@ jobs:
             platform: linux/amd64
           - arch: ia32
             runner: ubuntu-latest
-            platform: linux/386
+            platform: linux/amd64
           - arch: arm64
             runner: linux-arm64
             platform: linux/arm64
@@ -51,7 +51,7 @@ jobs:
       - name: Build
         run: |
           docker run --rm -i \
-                     --platform ${{ matrix.arch == 'arm' && 'linux/amd64' || matrix.platform }} \
+                     --platform ${{ matrix.platform }} \
                      --volume "$PWD:$PWD" \
                      --workdir "$PWD" \
                      ghcr.io/dart-musl/dart <<'EOF'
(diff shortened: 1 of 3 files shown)


ntkme:
On the other hand, `windows-arm64` build is blocked by https://github.com/dart-lang/setup-dart/issues/118

ntkme:
`linux-musl-arm` seems to have an unknown issue.

https://github.com/sass/dart-sass/actions/runs/9554504858/job/26336034221

From the output of `Setup QEMU` step we can tell that it's not using QEMU for arm/v7:

```
Installing QEMU static binaries
  /usr/bin/docker run --rm --privileged tonistiigi/binfmt:latest --install all
  installing: ppc64le OK
  installing: 386 OK
  installing: s390x OK
  installing: riscv64 OK
  installing: mips64le OK
  installing: mips64 OK
  installing: amd64 OK
```

Also, the fact that `dart pub get` runs without issue also confirms that we're no longer hitting the QEMU bug, because it's running natively. However, I'm not sure what's wrong that it seems to be stuck after running `dart pub get`.

ntkme:
Turns out linux-arm-musl was running on the wrong platform. Added another commit to fix that.

.github/workflows/build-linux.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             platform: linux/amd64
           - arch: ia32
             runner: ubuntu-latest
-            platform: linux/386
+            platform: linux/amd64
```
ntkme, line 22: Official dart docker image do not have `linux/386` support. Use `linux/amd64` to build snapshot, as ia32 has no AOT support anyways.

.github/workflows/build-linux-musl.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             platform: linux/amd64
           - arch: ia32
             runner: ubuntu-latest
-            platform: linux/386
+            platform: linux/amd64
```
ntkme, line 22: This was ok, but updating just to match the linux-ia32 config.

nex3 (approved):

PR #2266, opened 2024-06-17
Parse silent comments in `_interpolatedDeclarationValue()`

### LINKED ISSUE #2263: Silent comments are incorrectly preserved in at-rule values
For example, the SCSS stylesheet

```scss
@foo bar // comment
```

currently produces the CSS

```css
@foo bar // comment;
```

([Playground link](https://sass-lang.com/playground/#eJwzNHRIy89XSEosUtDXV0jOz81NzSvhAgBMAQbX))

This is invalid, and an incorrect interpretation of the silent comment semantics.

nex3:
Closes #2263

See sass/sass-spec#2000

### DIFF
--- a/lib/src/parse/parser.dart
+++ b/lib/src/parse/parser.dart
@@ -133,7 +133,8 @@ class Parser {
     whitespace();
   }
 
-  /// Consumes and ignores a silent (Sass-style) comment.
+  /// Consumes and ignores a single silent (Sass-style) comment, not including
+  /// the trailing newline.
   ///
   /// Returns whether the comment was consumed.
   @protected
--- a/lib/src/parse/stylesheet.dart
+++ b/lib/src/parse/stylesheet.dart
@@ -380,7 +380,8 @@ abstract class StylesheetParser extends Parser {
     // Parse custom properties as declarations no matter what.
     var name = nameBuffer.interpolation(scanner.spanFrom(start, beforeColon));
     if (name.initialPlain.startsWith('--')) {
-      var value = StringExpression(_interpolatedDeclarationValue());
+      var value = StringExpression(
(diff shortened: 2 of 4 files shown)


lib/src/parse/parser.dart
```diff
@@ -133,7 +133,8 @@ class Parser {
     whitespace();
   }
 
-  /// Consumes and ignores a silent (Sass-style) comment.
+  /// Consumes and ignores a single silent (Sass-style) comment, not including
+  /// the trailing newline.
```
Goodwine, line 137: "not including the trailing newline" makes me wonder about what happens when there are consecutive lines with silent comments, but otherwise LGTM

Goodwine (approved):

lib/src/parse/parser.dart
nex3: This function will only consume one of them, but `whitespace()` will consume them all _and_ the whitespace.

lib/src/parse/stylesheet.dart
```diff
@@ -2854,7 +2877,22 @@ abstract class StylesheetParser extends Parser {
...
+              silentComment();
+              wroteNewline = false;
+
+            case _:
+              buffer.writeCharCode(scanner.readChar());
+              wroteNewline = false;
+          }
+
+        case $slash when silentComments && scanner.peekChar(1) == $slash:
           buffer.write(rawText(loudComment));
```
stof, line 2896: should this really parse a loud comment when finding `//` ? Or is this dead code because of going through the previous case all the time ?

nex3: You're right that it's incorrect and also that it's dead code. I'll send out a PR to remove it.

PR #2267, opened 2024-06-19
Deprecated mixed declarations

nex3:
See sass/sass#3885
See https://github.com/sass/sass-spec/pull/2002


jathak (approved):
This needs a changelog entry and the sass_api major version should be bumped.

jathak (approved):

nex3:
Dartdoc seems to be failing due to https://github.com/dart-lang/dartdoc/issues/3818

stefanv:
Confusingly, the page linked to in the warning gives an example that, due CSS specificity rules, renders the same under old and new behaviors. Do we have examples of where new behavior creates differences?

nex3:
Yes, any example where the nested rule doesn't affect specificity:

```scss
.example {
  color: red;

  @media screen {
    font-weight: bold;
  }

  font-weight: normal;
}
```

PR #2274, opened 2024-07-09
Fix windows-arm64 release

ntkme:
This PR attempts to fix the release failure in https://github.com/sass/dart-sass/actions/runs/9864595353/job/27240444712

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -24,8 +24,16 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: ./.github/util/initialize
-        with: {github-token: "${{ github.token }}"}
+      - uses: dart-lang/setup-dart@v1
+
+      - uses: bufbuild/buf-setup-action@v1.30.0
+        with: {github_token: "${{ github.token }}"}
+
+      - name: Install Dependencies
+        run: dart pub get
+
+      - name: Compile Protobuf
+        run: dart run grinder protobuf
 
       - name: Build
         run: dart run grinder pkg-standalone-windows-${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


nex3:
Can you add a comment here indicating that we can't use the shared workflow and linking to the GitHub discussion?

nex3 (approved):

ntkme:
https://github.com/sass/dart-sass/actions/workflows/build-windows.yml

@nex3 By the way, there should be a "run workflow" button on page the linked above. After we merge this PR, you can use the "run workflow" button to test it out, that it will build the windows artifacts without an actual release. I cannot personally try it as I do not have access to see the button, so it would be very helpful if you can trigger a run after merging to see if there is any other issue or not.

ntkme:
It looks like dart-sdk on windows 11 for arm might be missing CA certs like this issue: https://github.com/dart-lang/pub/issues/1882

ntkme:
@nex3 Can we try change the command to `dart pub get --verbose` so that we can see what exactly failed?

nex3:
Rerunning with `--verbose` now.

nex3:
[Here's the result](https://github.com/sass/dart-sass/actions/runs/9883494166/job/27298335389), although I'm not sure how much more information it gives us. I think it may be better at this point to just resurrect https://github.com/sass/dart-sass/pull/2275 to get releases back on track.

ntkme:
```
FINE: Wrapped exception: HandshakeException: Handshake error in client (OS Error: 
    | 	CERTIFICATE_VERIFY_FAILED: unable to get local issuer certificate(../../third_party/boringssl/src/ssl/handshake.cc:393))
```

It looks indeed like missing ca certs. I think unable to get local issuer means unable to get parent certificate for the server certificate. This parent is either an intermediate certificate (usually served by server) or a root ca (need to be trusted locally).

So most likely there is an issue with the default root CA config.

ntkme:
@nex3 If you want to do #2275, you have to just switch the runner to `windows-latest` to build a kernel snapshot. Disabling it will cause release failure in embedded-host-node.

PR #2275, opened 2024-07-09
Run Windows ARM64 releases on windows-latest instead

nex3:
See https://github.com/orgs/community/discussions/131594

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -18,9 +18,11 @@ jobs:
             runner: windows-latest
           - arch: ia32
             runner: windows-latest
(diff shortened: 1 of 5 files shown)


jathak (approved):

ntkme:
The problem seems to be that in composite action an explicit `shell` is required, but in a normal action the shell defaults to `bash` on linux and mac  and `pwsh` windows. However, there is no way to get the same behavior on the in a composite action.

For building the release, we can get it working without using the composite action. I've put up a PR for it.

@nex3 Would you mind try if #2274 works?

nex3:
Closing in favor of https://github.com/sass/dart-sass/pull/2274

jathak (approved):

PR #2286, opened 2024-07-23
Enable AOT build for windows-arm64

ntkme:
This PR is another attempt to enable AOT snapshot for windows-arm64. It turns out the TLS issue we're seeing is likely not a windows runner bug but a windows feature and a dart bug: https://github.com/dart-lang/sdk/issues/52266

In summary, the problem is that windows would lazily initialize some trusted certificates, but dart does not support this lazy initialization. The x86_64 windows runner image likely already have the missing root certificates initialized at some point by other calls made to other google services during its image provisioning, but this arm64 image isn't the same.

@nex3 I would like to give this another try, but I don't have access to arm64 runners. If you can push this commit to a branch directly in sass/dart-sass repo, you can use "run workflow" feature to test it without needing to merge this PR first.

### DIFF
--- a/.github/util/initialize/action.yml
+++ b/.github/util/initialize/action.yml
@@ -44,6 +44,8 @@ runs:
     - uses: bufbuild/buf-setup-action@v1.30.0
       with: {github_token: "${{ inputs.github-token }}"}
 
+    # This composite action requires bash, but bash is not available on windows-arm64 runner.
+    # Avoid running this composite action on non-PR, so that we can release on windows-arm64.
     - name: Check out the language repo
       if: github.event_name == 'pull_request'
       uses: sass/clone-linked-repo@v1
--- a/.github/util/initialize/action.yml
+++ b/.github/util/initialize/action.yml
@@ -20,18 +20,18 @@ runs:
     # See: https://github.com/dart-lang/sdk/issues/52266
     - run: Invoke-WebRequest https://pub.dev
       if: runner.os == 'Windows'
-      shell: pwsh
+      shell: powershell
 
     # See: https://github.com/orgs/community/discussions/131594
     # The composite action requires an explict shell, but bash is not available on windows-arm64 runner.
-    # For the following commands conditionally use bash or pwsh based on the runner.os:
+    # For the following commands conditionally use bash or powershell based on the runner.os:
     - run: dart pub get
       if: runner.os != 'Windows'
       shell: bash
 
     - run: dart pub get
       if: runner.os == 'Windows'
-      shell: pwsh
+      shell: powershell
 
     - run: npm install
       if: runner.os != 'Windows'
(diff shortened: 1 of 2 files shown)


.github/util/initialize/action.yml
```diff
@@ -17,20 +17,48 @@ runs:
       with:
         node-version: "${{ inputs.node-version }}"
 
+    # See: https://github.com/dart-lang/sdk/issues/52266
+    - run: Invoke-WebRequest https://pub.dev
+      if: runner.os == 'Windows'
+      shell: pwsh
```
ntkme, line 23: Hopefully this should fix the TLS issue.

.github/util/initialize/action.yml
```diff
@@ -17,20 +17,48 @@ runs:
...
 
+    - run: npm install
+      if: runner.os == 'Windows'
+      shell: pwsh
+
     - uses: bufbuild/buf-setup-action@v1.30.0
       with: {github_token: "${{ inputs.github-token }}"}
 
     - name: Check out the language repo
+      if: github.event_name == 'pull_request'
```
ntkme, line 48: This is to avoid running the `clone-linked-repo` during release build, which depend on `bash` and does not work on windows-arm64.

nex3: Document this in a comment.

nex3:
[Test build is running](https://github.com/sass/dart-sass/actions/runs/10066457948)

ntkme:
Turns out there are two versions of powershell: PowerShell Core (crossbow platform, named as `pwsh`) and PowerShell Desktop (windows-only, named as `powershell`): https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunshell

`pwsh` does not seem to work on windows-arm64, I added comment to try `powershell`. @nex3 Can you please help give this a try?

nex3:
[Running now](https://github.com/sass/dart-sass/actions/runs/10067296757)

ntkme:
The TLS certificate workaround is confirmed to be working. Now it turns out even `git` is not pre-installed. I pushed another commit which should address it.

nex3:
[Running](https://github.com/sass/dart-sass/actions/runs/10068499167)

ntkme:
Looks like everything worked!

nex3 (approved):

.github/util/initialize/action.yml
```diff
@@ -17,20 +17,55 @@ runs:
...
+    # See: https://github.com/orgs/community/discussions/131594
+    # The composite action requires an explict shell, but bash is not available on windows-arm64 runner.
+    # For the following commands conditionally use bash or powershell based on the runner.os:
     - run: dart pub get
+      if: runner.os != 'Windows'
       shell: bash
 
+    - run: dart pub get
+      if: runner.os == 'Windows'
+      shell: powershell
```
stof, line 34: as the command is the same in both case, this could be simplified by letting the default shell be used.

ntkme: That’s already refactored in a later PR.

PR #2287, opened 2024-07-24
Let dependabot check updates for composite actions

ntkme:
Directories containing composite actions need to be explicitly listed in dependabot config in order for it to check for updates: https://github.com/dependabot/dependabot-core/issues/6704

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,6 @@ updates:
   - package-ecosystem: "github-actions"
     directories:
       - "/"
-      - "/.github/util/initialize/"
-      - "/.github/util/sass-spec/"
+      - "/.github/util/*/"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,6 +5,9 @@ updates:
     schedule:
       interval: "weekly"
   - package-ecosystem: "github-actions"
-    directory: "/"
+    directories:
+      - "/"
+      - "/.github/util/initialize/"
+      - "/.github/util/sass-spec/"
```
nex3, line 11: Can we consolidate these into

```suggestion
      - "/.github/util/*/"
```

ntkme: Change committed. I haven't tested this but it seems that wildcard is supported according to dependabot doc.

PR #2293, opened 2024-07-29
Backport deprecation API to legacy JS API

jathak:
See https://github.com/sass/sass/pull/3907.
See https://github.com/sass/sass-spec/pull/2006.

This also fixes a bug where the deprecation flags would not always be respected when using embedded Sass, as `ImportCache`s could be created with an unwrapped logger, meaning any parse-time deprecation warnings (e.g. `@import`) couldn't be controlled. `ImportCache` now wraps the logger it gets in a `DeprecationProcessingLogger` to fix this.

[skip sass-embedded]


bin/sass.dart
```diff
@@ -46,18 +46,13 @@ Future<void> main(List<String> args) async {
...
-            silenceDeprecations: options.silenceDeprecations,
-            fatalDeprecations: options.fatalDeprecations,
-            futureDeprecations: options.futureDeprecations,
-            limitRepetition: false)));
+      importers: [...options.pkgImporters, FilesystemImporter.noLoadPath],
+      loadPaths: options.loadPaths,
+      logger: options.logger,
+      silenceDeprecations: options.silenceDeprecations,
+      fatalDeprecations: options.fatalDeprecations,
+      futureDeprecations: options.futureDeprecations,
```
nex3, line 54: Nit: no trailing comma

lib/src/deprecation.dart
```diff
@@ -6,6 +6,7 @@ import 'package:cli_pkg/js.dart';
 import 'package:collection/collection.dart';
 import 'package:pub_semver/pub_semver.dart';
 
+import 'logger.dart';
```
nex3, line 9: Why?

lib/src/async_import_cache.dart
```diff
@@ -95,20 +96,40 @@ final class AsyncImportCache {
       {Iterable<AsyncImporter>? importers,
       Iterable<String>? loadPaths,
       PackageConfig? packageConfig,
-      Logger? logger})
+      Logger? logger,
+      Set<Deprecation>? fatalDeprecations,
+      Set<Deprecation>? futureDeprecations,
+      Set<Deprecation>? silenceDeprecations})
```
nex3, line 102: I don't really like this change. It's not actually providing any better guarantees—the caller still has to remember to explicitly handle the deprecation options for a new `ImportCache`—and it's reducing the separation of concerns between the way a `Logger` is configured and the way an `ImportCache` is configured. I'd rather just wrap the logger explicitly in the places it needs to be wrapped.

lib/src/logger/deprecation_processing.dart
```diff
@@ -45,7 +45,10 @@ final class DeprecationProcessingLogger extends LoggerWithDeprecationType {
       {required this.silenceDeprecations,
       required this.fatalDeprecations,
       required this.futureDeprecations,
-      this.limitRepetition = true}) {
+      this.limitRepetition = true});
+
+  /// Warns if any of the deprecations options are incompatible or unnecessary.
+  void validate() {
```
nex3, line 51: Why split this out into a separate method?

nex3 (changes requested):

lib/src/async_import_cache.dart
jathak: Moved the wrapping for import cache into a static method so that it's now always wrapped explicitly, but we don't have to repeat the boilerplate (including the comment about not limiting repetition) in each place we wrap a logger this way.

lib/src/logger/deprecation_processing.dart
jathak: We only want to emit warnings for invalid/unnecessary deprecation flags once and we're now wrapping loggers more than once in many cases, so the warning part is separated out from the constructor and only called on the main wrapped logger.

nex3 (approved):

lib/src/async_import_cache.dart
jathak: Oh, whoops. I think I accidentally removed them from `ImportCache` instead of `AsyncImportCache` and then that got undone by synchronizing.

PR #2304, opened 2024-08-10
Initial implementation of a PostCSS-compatible parser JS API

nex3:
This is just a vertical slice designed to solidify the general
principles of the API design and ensure that everything works as
expected. It's not yet usable as a full-fledged parser.


tool/grind/sass_api.dart
```diff
@@ -0,0 +1,80 @@
...
+
+  // Remove the dependency override on `sass`, because otherwise it will block
+  // publishing.
+  var pubspecYaml = Map<dynamic, dynamic>.of(
+      loadYaml(File(pubspecPath).readAsStringSync()) as YamlMap);
+  pubspecYaml.remove("dependency_overrides");
+  File(pubspecPath).writeAsStringSync(json.encode(pubspecYaml));
+
+  // We use symlinks to avoid duplicating files between the main repo and
+  // child repos, but `pub lish` doesn't resolve these before publishing so we
```
Goodwine, line 43: TIL `pub lish` is an alias to `pub publish` :D

pkg/sass-parser/lib/src/statement/generic-at-rule.ts
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * An `@`-rule that isn't parsed as a more specific type. Extends
+ * [`postcss.AtRule`].
+ *
+ * [`postcss.AtRule`]: https://postcss.org/api/#atrule
+ *
+ * @category Statement
+ */
+export class GenericAtRule
+  extends _AtRule<Partial<GenericAtRuleProps>>
```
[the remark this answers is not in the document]
nex3: That's right. Our `Node` class is only used as a superclass for *non-statement* nodes like `Interpolation` and `Expression`s. Our statements inherit from their PostCSS equivalents (through a synthetic TypeScript-only class that tweaks its interface).

Goodwine (approved):

lib/src/ast/sass/expression.dart
```diff
@@ -13,15 +13,20 @@ import '../../value.dart';
 import '../../visitor/interface/expression.dart';
 import '../sass.dart';
 
+// Note: despite not defining any methods here, this hsa to be a concrete class
```
stof, line 16:
```suggestion
// Note: despite not defining any methods here, this has to be a concrete class
```

lib/src/ast/sass/statement.dart
```diff
@@ -5,10 +5,13 @@
 import '../../visitor/interface/statement.dart';
 import 'node.dart';
 
+// Note: despite not defining any methods here, this hsa to be a concrete class
```
stof, line 8:
```suggestion
// Note: despite not defining any methods here, this has to be a concrete class
```

lib/src/js/parser.dart
```diff
@@ -0,0 +1,94 @@
...
+          (JSStatementVisitorObject inner) => JSStatementVisitor(inner)));
+}
+
+/// Modifies the prototypes of the Sass AST classes to provide access to JS.
+///
+/// This API is not intended to be used directly by end users and is subject to
+/// breaking changes without notice. Instead, it's wrapped by the `sass-parser`
+/// package which exposes a PostCSS-style API.
+void _updateAstPrototypes() {
+  // We don't need explicit getters for field names, becuase dart2js preserves
```
stof, line 55:
```suggestion
  // We don't need explicit getters for field names, because dart2js preserves
```

pkg/sass-parser/lib/src/expression/string.ts
```diff
@@ -0,0 +1,203 @@
...
+    const rawText = this.text.raws.text;
+    const rawExpressions = this.text.raws.expressions;
+    for (let i = 0; i < this.text.nodes.length; i++) {
+      const element = this.text.nodes[i];
+      if (typeof element === 'string') {
+        const raw = rawText?.[i];
+        // The Dart Sass AST preserves string escapes for unquoted strings
+        // because they serve a dual purpose at runtime of representing
+        // identifiers (which may contain escape codes) and being a catch-all
+        // represenetation for unquoted non-identifier values such as `url()`s.
```
stof, line 118:
```suggestion
        // representation for unquoted non-identifier values such as `url()`s.
```

pkg/sass-parser/lib/src/interpolation.ts
```diff
@@ -0,0 +1,422 @@
...
+/**
+ * The type of new nodes that can be passed into an interpolation.
+ *
+ * @category Expression
+ */
+export type NewNodeForInterpolation =
+  | Interpolation
+  | Interpolation[]
+  | Expression
+  | Expression[]
```
stof, line 24: I suggest using `ReadonlyArray` for all those, to allow passing something typed as ReadonlyArray for the new node without having to copy the array first. The code in this class does not attempt to mutate the array anyway.

nex3: Good call. I think we can actually make this change to a number of parameters, and potentially also upstream in the PostCSS types.

Goodwine (approved):

PR #2312, opened 2024-08-17
Fix flaky shutdown

ntkme:
This PR fixes a few shutdown reliability issues including the following flaky test on GitHub Actions windows runner:

```
❌ test\embedded\importer_test.dart: emits a protocol error for a response without a corresponding request ID (failed)
  
  Process `dart_sass_embedded` was killed with SIGKILL in a tear-down. Output:
      canonicalizeRequest: {
        id: 0
        importerId: 1
        url: other
        fromImport: true
      }
      
  [e] Host caused params error: Response ID 1 doesn't match any outstanding requests in compilation 4321.
      error: {
        type: PARAMS
        id: 4294967295
        message: Response ID 1 doesn't match any outstanding requests in compilation 4321.
      }
  TimeoutException after 0:00:30.000000: Test timed out after 30 seconds. See https://pub.dev/packages/test#timeouts
  dart:isolate  _RawReceivePort._handleMessage
```

### DIFF
--- a/lib/src/embedded/isolate_dispatcher.dart
+++ b/lib/src/embedded/isolate_dispatcher.dart
@@ -115,25 +115,43 @@ class IsolateDispatcher {
     } else {
       var future = ReusableIsolate.spawn(_isolateMain);
       isolate = await future;
+      isolate.receivePort.listen((message) {
+        assert(isolate.borrowed,
+            "Shouldn't receive a message before being borrowed.");
+
+        var fullBuffer = message as Uint8List;
+
+        // The first byte of messages from isolates indicates whether the entire
+        // compilation is finished (1) or if it encountered an error (2). Sending
+        // this as part of the message buffer rather than a separate message
+        // avoids a race condition where the host might send a new compilation
+        // request with the same ID as one that just finished before the
+        // [IsolateDispatcher] receives word that the isolate with that ID is
+        // done. See sass/dart-sass#2004.
+        var category = fullBuffer[0];
+        var packet = Uint8List.sublistView(fullBuffer, 1);
+
+        switch (category) {
+          case 0:
+            _channel.sink.add(packet);
+          case 1:
+            _activeIsolates.remove(compilationId);
+            _inactiveIsolates.add(isolate);
+            _channel.sink.add(packet);
+            isolate.release();
+          case 2:
+            _activeIsolates.remove(compilationId);
+            _channel.sink.add(packet);
+            _channel.sink.close();
+            isolate.release();
+        }
+      }, onError: (Object error, StackTrace stackTrace) {
+        _handleError(error, stackTrace);
+      });
       _allIsolates.add(isolate);
     }
 
-    isolate.checkOut().listen(_channel.sink.add,
-        onError: (Object error, StackTrace stackTrace) {
-      if (error is ProtocolError) {
-        // Protocol errors have already been through [_handleError] in the child
-        // isolate, so we just send them as-is and close out the underlying
-        // channel.
-        sendError(compilationId, error);
-        _channel.sink.close();
-      } else {
-        _handleError(error, stackTrace);
-      }
-    }, onDone: () {
-      _activeIsolates.remove(compilationId);
-      _inactiveIsolates.add(isolate);
-      resource.release();
-    });
+    isolate.borrow(resource);
 
     return isolate;
   }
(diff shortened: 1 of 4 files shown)


ntkme:
Based on extensive amount of testing in Github Actions, before the PR it had failure rate higher than 1/100. With the first commit in this PR (`Exit isolate on ProtocolError`), it reduces the failure rate to less than 1/1000. The first commit guarantees reliable shutdown of isolate that had ProtocolError, which indicate that the shutdown failure must be caused by event loop in main isolate. After the second commit (`Remove intermediate StreamController`), I was able to run the test 10000 times (10 jobs, 1000 iterations per job) without an error here: https://github.com/ntkme/dart-sass-test/actions/runs/10443012158

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -43,6 +44,9 @@ class IsolateDispatcher {
   /// See https://github.com/sass/dart-sass/pull/2019
   final _isolatePool = Pool(sizeOf<IntPtr>() <= 4 ? 7 : 15);
 
+  /// Whether the stdin has been closed or not.
```
nex3, line 47: Nit: better to say

```suggestion
  /// Whether [_channel] has been closed or not.
```

because this class doesn't know or care whether that's connected to stdin (and in fact it may not be in future implementations if, for example, we use this via WASM in the future).

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -43,6 +44,9 @@ class IsolateDispatcher {
...
+  bool _closed = false;
```
nex3, line 48: Nit:
```suggestion
  var _closed = false;
```

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -105,25 +116,41 @@ class IsolateDispatcher {
     } else {
       var future = ReusableIsolate.spawn(_isolateMain);
       isolate = await future;
+      isolate.receivePort.listen((message) {
```
nex3, line 119: Why is it necessary to pull this logic out here, rather than keeping it in `ReusableIsolate`?

lib/src/embedded/reusable_isolate.dart
```diff
@@ -53,58 +48,20 @@ class ReusableIsolate {
...
-        _subscription.onData(_defaultOnData);
-        _subscription.onError(null);
-        controller.close();
-      }
-    });
-
-    _subscription.onError(controller.addError);
+  /// Request this isolate as part of a pool and mark it as in use.
+  void borrow(PoolResource resource) {
+    assert(!borrowed, 'ReusableIsolate has already been borrowed.');
```
nex3, line 56: Checks for "did you call this at the correct time" should throw `StateError`s, not just run assertions.

lib/src/embedded/isolate_dispatcher.dart
ntkme: From previous implementation:

```
        // Parse out the compilation ID and surface the [ProtocolError] as an
        // error. This allows the [IsolateDispatcher] to notice that an error
        // has occurred and close out the underlying channel.

        // Protocol errors have already been through [_handleError] in the child
        // isolate, so we just send them as-is and close out the underlying
        // channel.
```

First, the motivation to remove the intermediate `StreamController` is to eliminate two unnecessary performance overheads:
1. To eliminate the overhead of deserializing `ProtocolError` packet, passing it through `addError`, and re-serializing it back to packet. - We don't actually care about the `ProtocolError` object, all we care is that the category 2 indicate the packet contains a ProtocolError.
2. To eliminate the overhead of allocating new `StreamController` for every single compilation.

In regarding of why the `receivePort.listen` is moved to `IsolateDispatcher`, IMO logically how the message is sent is a contract between `IsolateDispatcher` and `CompilationDispatcher`, `ReusableIsolate` is should purely be a passthrough helper just to help with creating and managing isolate for repeated use.

nex3: I think you're right that putting the decoding logic here makes the most sense. But I'm not a big fan of giving the caller responsibility for maintaining a single stream subscription across multiple borrows, and I'm not very worried about the overhead of an extra `StreamController`. Can we make `.borrow()` return a subscription that just forwards a view of the `Stream<object?>` which closes automatically once `.return()` is called?

nex3 (changes requested):

lib/src/embedded/isolate_dispatcher.dart
ntkme: Not precisely a new `Stream` each time, but I made `.borrow` take a new `onData` function each time, which behaves somewhat like the previous implementation but without creating new `StreamController`.

nex3 (approved):

nex3:
Thanks for working on this and helping reduce flakiness!

PR #2317, opened 2024-08-22
Add more statement parsing for `sass-parser`

nex3:


pkg/sass-parser/lib/src/stringifier.ts
```diff
@@ -67,7 +70,67 @@ export class Stringifier extends PostCssStringifier {
     )(statement, semicolon);
   }
 
+  private ['debug-rule'](node: DebugRule, semicolon: boolean): void {
+    this.builder(
+      '@debug' +
+        (node.raws.afterName ?? ' ') +
```
Goodwine, line 76: I just realized that setting `afterName = ''` in the case of `@debug foo` would result in `@debugfoo`. I was worried about that, but then I tried it on postcss and that's the same behavior! o_O

Interestingly, in postcss, if I set `aftername = null`, this results in `@debugnullfoo` 🤔

pkg/sass-parser/lib/src/stringifier.ts
```diff
@@ -67,7 +70,67 @@ export class Stringifier extends PostCssStringifier {
...
+      node,
+      '@for' +
+        (node.raws.afterName ?? ' ') +
+        node.params +
+        (node.raws.between ?? '')
+    );
+  }
+
   private atrule(node: GenericAtRule, semicolon: boolean): void {
+    // If the @at-root shorthand, stringify `@at-root {.foo {...}}` as
```
Goodwine, line 116: It is a bit awkward to read, but I'm not sure how to word it differently 🤔

Goodwine (changes requested):

pkg/sass-parser/lib/src/statement/index.ts
```diff
@@ -30,14 +37,29 @@ export type AnyStatement = Root | Rule | GenericAtRule;
...
 /**
  * All Sass statements that are also at-rules.
  *
  * @category Statement
  */
-export type AtRule = GenericAtRule;
+export type AtRule = DebugRule | EachRule | ErrorRule | ForRule | GenericAtRule;
+
+/**
+ * All Sass statements that are comments.
```
Goodwine, line 58: Is this plural because we will add a separate type for silent comments?

Goodwine (approved):

pkg/sass-parser/lib/src/statement/index.ts
nex3: Exactly.

PR #2318, opened 2024-08-23
Ignore new `unreachable_switch_default` warning.

stereotype441:
The Dart analyzer will soon be changed so that if the `default` clause of a `switch` statement is determined to be unreachable by the exhaustiveness checker, a new warning of type
`unreachable_switch_default` will be issued. This parallels the behavior of the existing `unreachable_switch_case` warning, which is issued whenever a `case` clause of a `switch` statement is determined to be unreachable.

In the vast majority of cases, the most reasonable way to address the warning is to remove the unreachable `default` clause. However, in a few rare cases, the `default` clause must be kept, due to the fact that flow analysis is not as sophisticated as exhaustiveness checking (see https://github.com/dart-lang/language/issues/2977 for details).

Two of these rare cases crop up in dart-sass. This change adds `ignore` comments to avoid a spurious warning, and adds a comment explaining why the `default` clause needs to be kept.

### DIFF
--- a/lib/src/functions/map.dart
+++ b/lib/src/functions/map.dart
@@ -61,7 +61,15 @@ final _set = BuiltInCallable.overloadedFunction("set", {
         throw SassScriptException("Expected \$args to contain a value.");
       case [...var keys, var value]:
         return _modify(map, keys, (_) => value);
-      default:
+      default: // ignore: unreachable_switch_default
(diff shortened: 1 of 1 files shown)


nex3 (approved):
Thanks for the proactive mitigation!

PR #2323, opened 2024-08-28
Delete unreachable `default` clause.

stereotype441:
The Dart analyzer will soon be changed so that if the `default` clause of a `switch` statement is determined to be unreachable by the exhaustiveness checker, a new warning of type
`unreachable_switch_default` will be issued. This parallels the behavior of the existing `unreachable_switch_case` warning, which is issued whenever a `case` clause of a `switch` statement is determined to be unreachable. For details see
https://github.com/dart-lang/sdk/issues/54575.

This PR deletes an unreachable `default` clause from `dart-sass` now, to avoid a spurious warning when the analyzer change lands.


stereotype441:
@nex3 Sorry for not bundling this with the very similar https://github.com/sass/dart-sass/pull/2318. I'm working my way through a google3 LSC, and I failed to notice that these two changes both affected the same open source repo.

PR #2333, opened 2024-09-05
Add common Sass TypeScript configuration to the analysis package

nex3:


pkg/sass-parser/package.json
```diff
@@ -37,12 +37,14 @@
   "devDependencies": {
     "@types/jest": "^29.5.12",
     "copyfiles": "^2.4.1",
+    "eslint-plugin-import": "^2.30.0",
     "expect": "^29.7.0",
     "gts": "^5.0.0",
```
stof, line 42: is it needed to still have a direct dependency on ègts` or is the transitive dependency enough (`sass-analysis` will use its own `gts` dependency)

analysis/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json"
```
stof, line 2: does this work if `gts` is hoisted differently ? It should probably use `gts/tsconfig-google.json` instead, relying on the node-style resolution of the package

stof (commented):
should the eslint config be migrated to flat config first instead of exposing a config in the old format ? Eslint 9 does not support legacy configs anymore.

nex3:
Unfortunately, it seems that gts is _still_ not compatible with eslint flat configs (https://github.com/google/gts/issues/830). I don't think we can land this without either that or moving away from GTS entirely.

Issue #1805, opened 2022-09-21
Implement color spaces


nex3:
See https://github.com/sass/sass/issues/2831

stof:
@nex3 what is the state of this implementation ? Color Level 4 is now coming to all browsers (shipped in Safari and Chrome; available behind a flag in Firefox which is enabled by default in Firefox Nightly so likely to ship in a few weeks) so it would be great to have access to them in Sass.

nex3:
@mirisuzanne and I are still hammering out the details. A draft implementation is available on the `features.color-4` branch, but the spec keeps changing out from under us so it's been difficult to nail down 100% final behavior. (It's paradoxically _easier_ for browsers to change their edge case behavior than for us, since browsers have the ability to see whether something is used in the wild.)

The other big blocker here is getting tests written/updated for every corner of the (very large) proposal. Although the implementation is likely _mostly_ correct, I'm sure I got some corners wrong in ways we won't know until we test it.

stof:
@nex3 the implementation should be marked as done

PR #2341, opened 2024-09-12
Merge feature.color-4 into main

nex3:
See https://github.com/sass/dart-sass/issues/1805
See https://github.com/sass/sass/pull/3930
See https://github.com/sass/sass-spec/pull/2013
[skip sass-embedded]


lib/src/value/color/channel.dart
```diff
@@ -0,0 +1,106 @@
...
+  @internal
+  const ColorChannel(this.name,
+      {required this.isPolarAngle, this.associatedUnit});
+
+  /// Returns whether this channel is [analogous] to [other].
+  ///
+  /// [analogous]: https://www.w3.org/TR/css-color-4/#interpolation-missing
+  bool isAnalogous(ColorChannel other) => switch ((name, other.name)) {
+        ("red" || "x", "red" || "x") ||
+        ("green" || "y", "gren" || "y") ||
```
jathak, line 45:
```suggestion
        ("green" || "y", "green" || "y") ||
```

test/dart_api/value/color_test.dart
```diff
@@ -2,6 +2,7 @@
 // MIT-style license that can be found in the LICENSE file or at
 // https://opensource.org/licenses/MIT.
 
+@Skip("TODO(nweiz): Update these for the new Color API")
```
jathak, line 5: Should this have a tracking bug?

jathak (approved):

test/dart_api/value/color_test.dart
nex3: Oops! I've just updated these in this PR, PTAL

jathak (approved):

nex3:
I think the CI is going weird here because GitHub can't tell the difference between the PR actions and the normal branch actions, and the branch actions aren't handling the link with the language repo properly. I'm just going to merge and clean up any failures after the fact.

stof:
Actually, GitHub can differentiate them. But the dropdown showing the list has a max-width and truncates the names of the statuses when they are too long to fit, and the info to distinguish the status being set by `(push)` or `(pull_request)` is at the end of the name.

PR #2347, opened 2024-09-16
Bump chokidar to v4

### LINKED ISSUE #2345: Upgrade to chokidar 4
Chokidar v4 is out now. It drops a dozen dependencies, so is much smaller and lighter

ntkme:
Chokidar v4 targets `node >=14`, which is the same as the current sass package, so there is no breaking change for end users. See: https://github.com/paulmillr/chokidar?tab=readme-ov-file#upgrading

- Closes #2345

### DIFF
--- a/lib/src/io/js.dart
+++ b/lib/src/io/js.dart
@@ -252,8 +252,7 @@ Future<Stream<WatchEvent>> watchDir(String path, {bool poll = false}) {
   if (!isNodeJs) {
     throw UnsupportedError("watchDir() is only supported on Node.js");
   }
-  var watcher = chokidar.watch(
-      path, ChokidarOptions(disableGlobbing: true, usePolling: poll));
+  var watcher = chokidar.watch(path, ChokidarOptions(usePolling: poll));
 
   // Don't assign the controller until after the ready event fires. Otherwise,
   // Chokidar will give us a bunch of add events for files that already exist.
--- a/lib/src/js/chokidar.dart
+++ b/lib/src/js/chokidar.dart
@@ -12,10 +12,9 @@ class Chokidar {
 @JS()
 @anonymous
 class ChokidarOptions {
-  external bool? get disableGlobbing;
   external bool? get usePolling;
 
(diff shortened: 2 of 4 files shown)


jathak (approved):

nex3:
The new Chokidar version has caused a couple end user issues (#2362, #2373). It may be worth considering rolling it back until Chokidar v4 is more stable. @ntkme, is there a particular reason you wanted to do this update, or was it just general dependency hygiene?

ntkme:
v3 was very flaky in our GitHub CI, and I was mainly trying to see if v4 is less flaky. It seems like that it’s less flaky in CI but is giving end users lots of trouble.

Maybe we should revert this. On the other hand, maybe we should consider other options like https://www.npmjs.com/package/@parcel/watcher, although it doesn’t have a real “poll” mode.

nex3:
The lack of a polling watcher isn't an insurmountable problem, but it would require some engineering effort to deal with (possibly by moving the Dart watcher package's `PollingDirectoryWatcher` to a `dart:io`-independent library and adding the ability to pass in a [custom filesystem](https://pub.dev/packages/file)). Otherwise, the Parcel watcher does look good. But maybe it's better to just go back to chokidar v3 for now until we can put in that work.

ntkme:
> running brute-force Parcel queries in a loop by hand

I actually did a POC for that at some point and concluded it was flaky and kind of dirty (with what it supports as of now). The problem is that the “query” mode requires user to first write a snapshot to disk, and later “query” the current state against the previous snapshot. This introduces two ugly issues: 1. It does not support snapshot in memory, meaning each directory watched needs at least one temporary file on disk for snapshot, which needs clean up. 2. It does not have atomic “query and update snapshot”, meaning depends on whether we update snapshot or query first at each interval, we either risks double firing or missing events due to non-atomic operation.

Other than not having a good polling solution, pracel/watcher has been pretty good in the POC I did.

nex3:
Ah, that's too bad. We may need to rely on the Dart logic in that case.

Issue #2335, opened 2024-09-05
`unifyCompound` is not commutative when applied to pseudo classes of pseudo elements


mattrberry:
The following produces `.y.x::scrollbar:vertical` as expected

```dart
print(unifyCompound(
  CompoundSelector.parse('.x'),
  CompoundSelector.parse('.y::scrollbar:vertical'),
));
```

However, swapping the operands produces `.x.y:vertical::scrollbar`

```dart
print(unifyCompound(
  CompoundSelector.parse('.y::scrollbar:vertical'),
  CompoundSelector.parse('.x'),
));
```

Goodwine:
To clarify, because I was confused, the issue is not that `.x` and `.y` switched places. The issue is that `:vertical` and `::scrollbar` switched places

PR #2350, opened 2024-09-18
unifyComound() and unifyComplex() no longer move pseudo-classes across pseudo-element boundaries

Goodwine:
Fixes https://github.com/sass/dart-sass/issues/2335

See https://github.com/sass/sass-spec/pull/2018

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,16 @@
 ## 1.79.4
 
-* `selector.unify()` now preserves the original order of pseudo-classes and
-  pseudo-elements within individual selectors when combining them.
+* Changes to how `selector.unify()` and `@extend` combine selectors:
+
+    * The relative order of pseudo-classes (like `:hover`) and pseudo-elements
+      (like `::before`) within each original selector is now preserved when
+      they're combined.
+
+    * Pseudo selectors are now consistently placed at the end of the combined
+      selector, regardless of which selector they came from. Previously, this
+      reordering only applied to pseudo-selectors in the second selector.
+
+    * Prioritizes components from the first selector in the output.
 
 ### JS API
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@
       selector, regardless of which selector they came from. Previously, this
       reordering only applied to pseudo-selectors in the second selector.
 
-    * Prioritizes components from the first selector in the output.
-
 ### JS API
 
 * Fix a bug where passing `green` or `blue` to `color.change()` for legacy
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.79.4
+## 1.79.5
 
 * Changes to how `selector.unify()` and `@extend` combine selectors:
 
@@ -9,6 +9,8 @@
     * Pseudo selectors are now consistently placed at the end of the combined
       selector, regardless of which selector they came from. Previously, this
       reordering only applied to pseudo-selectors in the second selector.
+
+## 1.79.4
 
 ### JS API
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 1.79.4
 
+* `selector.unify()` now preserves the original order of pseudo-classes and
+  pseudo-elements within individual selectors when combining them.
+
 ### JS API
 
 * Fix a bug where passing `green` or `blue` to `color.change()` for legacy
--- a/lib/src/extend/functions.dart
+++ b/lib/src/extend/functions.dart
@@ -12,7 +12,6 @@
 library;
 
 import 'dart:collection';
-import 'dart:math';
 
 import 'package:collection/collection.dart';
 import 'package:source_span/source_span.dart';
--- a/lib/src/extend/functions.dart
+++ b/lib/src/extend/functions.dart
@@ -105,7 +105,7 @@ List<ComplexSelector>? unifyComplex(
 /// Returns a [CompoundSelector] that matches only elements that are matched by
 /// both [compound1] and [compound2].
 ///
-/// The [span] will be used for the new unified selector.
+/// The [compound1]'s `span` will be used for the new unified selector.
 ///
 /// This function ensures that the relative order of pseudo-classes (`:`) and
 /// pseudo-elements (`::`) within each input selector is preserved in the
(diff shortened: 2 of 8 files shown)


pkg/sass_api/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 12.0.2
+
+* `unifyComplex()` and `unifyCompound()` now preserve the original order of
+  pseudo-classes and pseudo-elements within individual selectors when combining
+  them.
```
nex3, line 5: Since this isn't an API that's specific to `sass_api`, just put "No user-visible changes" here.

lib/src/extend/functions.dart
```diff
@@ -108,16 +107,44 @@ List<ComplexSelector>? unifyComplex(
 ///
 /// The [span] will be used for the new unified selector.
 ///
+/// This function ensures that the relative order of pseudo-classes (`:`) and
+/// pseudo-elements (`::`) within each input selector is preserved in the
+/// resulting combined selector.
+///
+/// This function enforces a general rule that pseudo-classes (`:`) should come
```
nex3, line 114: Is this rule just applicable to pseudo-classes, or is it be applicable to _all_ simple selectors? It doesn't look like this function has any special handling for pseudo selectors.

nex3 (changes requested):

lib/src/extend/functions.dart
Goodwine: done, I added more test cases to catch bugs like.. `::foo.x` and `.x` which resulted in `.x::foo.x` 😅. It should now be producing the correct combined selector

CHANGELOG.md
```diff
@@ -1,7 +1,16 @@
...
+
+    * The relative order of pseudo-classes (like `:hover`) and pseudo-elements
+      (like `::before`) within each original selector is now preserved when
+      they're combined.
+
+    * Pseudo selectors are now consistently placed at the end of the combined
+      selector, regardless of which selector they came from. Previously, this
+      reordering only applied to pseudo-selectors in the second selector.
+
+    * Prioritizes components from the first selector in the output.
```
nex3, line 17: "Prioritizes" in what sense?

lib/src/extend/functions.dart
```diff
@@ -106,19 +104,63 @@ List<ComplexSelector>? unifyComplex(
...
+  var pseudoElementFound = false;
+
+  // Compound selectors may not contain more than one pseudo-element selectors.
+  if (compound1.components
+          .followedBy(compound2.components)
+          .whereType<PseudoSelector>()
+          .where((pseudo) => pseudo.isElement)
+          .toSet()
+          .length >
+      1) {
```
nex3, line 136: Unification can be part of a tight loop in `@extend`-heavy stylesheets, at which point allocating a bunch of new objects gets expensive. Consider checking whether both compound selectors even have pseudo-elements at all before doing this check.

lib/src/extend/functions.dart
```diff
@@ -106,19 +104,63 @@ List<ComplexSelector>? unifyComplex(
...
 CompoundSelector? unifyCompound(
     CompoundSelector compound1, CompoundSelector compound2) {
-  var result = compound2.components;
-  for (var simple in compound1.components) {
-    var unified = simple.unify(result);
-    if (unified == null) return null;
-    result = unified;
+  var result = <SimpleSelector>[];
+  var pseudoResult = <SimpleSelector>[];
+  var pseudoElementFound = false;
```
nex3, line 127: Declare these closer to where they're used.

lib/src/extend/functions.dart
```diff
@@ -106,19 +104,63 @@ List<ComplexSelector>? unifyComplex(
...
+    // first components into the second components.
+    if (simple == null) {
+      pseudoElementFound = false;
+      continue;
+    }
+
+    // All pseudo-classes are unified separately after a pseudo-element to
+    // preserve their relative order with the pseudo-element.
+    if (pseudoElementFound && simple is PseudoSelector) {
+      var unified = simple.unify(pseudoResult);
```
nex3, line 152: I'm really nervous about the way this _re-runs_ unification for `compound2`, both from a performance and a correctness perspective. For example:

```scss
.foo.foo.bar {x: y}
.baz {@extend .bar}
```

now produces

```css
.foo.foo.bar, .foo.baz {
  x: y;
}
```

with the duplicated `.foo` removed. I think a better way to handle it would be to ensure that `simple.unify()` injects itself in the correct place relative to an existing pseudo-element.

nex3 (changes requested):

CHANGELOG.md
Goodwine: Ah, interesting... So, if you notice, I changed the order of which selector gets unified into the other. It used to be that the first selector was unified into the other in `unifyCompound`. For example:

```
print(unifyCompound(
  CompoundSelector.parse('.x'),
  CompoundSelector.parse('.y'),
));
// > .y.x
```
And I didn't like that 😅 

But apparently `selector.unify()` doesn't have this issue 🤔 [[playground](https://sass-lang.com/playground/#eJwzNHQoLU5VUCpOLC62Kk7NSU0uyS9SsuZySElNKk1XgInoleZlplVqqOtVqOsoqOtVWlkVJxfl5+QkJRZZlaUWlWQmJ+aoawIA/RkaYQ==)]

Since that didn't change behavior, I deleted this line.

lib/src/extend/functions.dart
Goodwine: Does allocation happen with `toSet()`? IIUC the other functions are lazy. With that in mind, I'll check for the `length` being > 1 in the first place before doing `toSet()`. Does that sound reasonable?

lib/src/extend/functions.dart
Goodwine: When I ran this before, I noticed that there were duplicate selectors, and kinda forgot about specificity! I'll change this back :)

lib/src/extend/functions.dart
Goodwine: actually, since I removed the tests where one selector contains `::x::y`, then I think we don't need this. The consequence is that `selector.unify('::x::y', 'z')` will produce `z::x::y` instead of `null`, but the input selectors don't make sense.. so I think that's OK?

nex3: Yeah, that's fine. That's the sort of thing we'd only really expect to be accurate if we started parsing pseudo-elements as combinators.

nex3 (approved):

Issue #2353, opened 2024-09-18
color.channel can not auto convert between spaces and wrong docs about it


Mister-Hope:
Reproduction: https://sass-lang.com/playground/#eJwzNHAoLU5VUCpOLC62Ss7PyS9SsubiUgGzrBSU00zTzNPSgCIOKalJpekKYHG95IzEvLzUHA2IMh0F9fKMzJLUvNTiYnVNawAzfhqY

```scss
@use "sass:color";

$color: #f5f7ff;

@debug color.channel($color, 'whiteness');
```

```

$channel: Color #f5f7ff has no channel named whiteness.
  ╷
5 │ @debug color.channel($color, 'whiteness');
  │        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  - 5:8  root stylesheet
```

I don't think this is the expected behavior.

The docs says:

> ⚠️ Heads up!
>
> Because color.whiteness() is redundant with [color.channel()](https://sass-lang.com/documentation/modules/color/#channel), it’s no longer recommended. Instead of color.whiteness($color), you can write [color.channel($color, "whiteness")](https://sass-lang.com/documentation/modules/color/#channel).

However, the correct name should be lightness or blackness on the [color space docs](https://sass-lang.com/documentation/values/colors/). Something worth noting is that the deprecation also requires users to convert color space manually, as `color.channel` won't do that. An exmaple here:

```scss
@use "sass:color";

$color: #f5f7ff;

// $channel: Color #f5f7ff has no channel named lightness.
@debug color.channel($color, 'lightness');
// works and return 98.0392156863%
@debug color.channel(color.to-space($color, hsl), 'lightness');
```

So I believe at least the docs needs an update.

If the current behavior of `color.channel` is a bug, and sass team believes a channel name could be returned with any legacy color space by automatically converted to a correct space behind the sence, then only whiteness and blackness should be corrected in docs.

Otherwise, every deprecation about `color.channel` should be corrected, like `color.red($color)` must be replaced by `color.channel(color.to-space($color, rgb), 'red')`, unless the developer is sure that `$color` MUST be in rgb space

nex3:
[...]

PR #2354, opened 2024-09-19
Add a `$space` parameter to the channel function deprecation

nex3:
Closes sass/dart-sass#2353
See sass/sass-spec#2019

### DIFF
--- a/pkg/sass-parser/package.json
+++ b/pkg/sass-parser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sass-parser",
-  "version": "0.2.0",
+  "version": "0.2.1-dev",
   "description": "A PostCSS-compatible wrapper of the official Sass parser",
   "repository": "sass/sass",
   "author": "Google Inc.",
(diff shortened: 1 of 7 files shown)


pkg/sass_api/pubspec.yaml
```diff
@@ -2,15 +2,15 @@ name: sass_api
 # Note: Every time we add a new Sass AST node, we need to bump the *major*
 # version because it's a breaking change for anyone who's implementing the
 # visitor interface(s).
-version: 12.0.1
+version: 12.0.2
```
Goodwine, line 5: since the regular sass version is `-dev`, should this be `-dev` too?

pkg/sass-parser/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sass-parser",
-  "version": "0.2.0",
+  "version": "0.2.1-dev",
```
Goodwine, line 3: This this supposed to change?

Goodwine (approved):

pkg/sass-parser/package.json
nex3: Yeah, just like `sass_api` this is released in lockstep with the main `sass` package so it needs to get its version bumped at the same time.

PR #2355, opened 2024-09-19
Add documentation link to legacy-js-api warning

jathak:


nex3 (approved):
I think you'll also need to remove `-dev` from the `pkg/` versions.

PR #2359, opened 2024-09-20
add --random-seed option

Joshix-1:
This is an important change to make the output of dart-sass reproducible.

### DIFF
--- a/lib/src/executable/compile_stylesheet.dart
+++ b/lib/src/executable/compile_stylesheet.dart
@@ -16,6 +16,7 @@ import '../importer/filesystem.dart';
 import '../io.dart';
 import '../stylesheet_graph.dart';
 import '../syntax.dart';
+import '../util/random.dart' show setRandomSeed;
 import '../utils.dart';
 import '../visitor/serialize.dart';
 import 'options.dart';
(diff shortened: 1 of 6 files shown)


lib/src/executable/compile_stylesheet.dart
```diff
@@ -16,6 +16,7 @@ import '../importer/filesystem.dart';
 import '../io.dart';
 import '../stylesheet_graph.dart';
 import '../syntax.dart';
+import '../util/random.dart' show setRandomSeed;
```
nex3, line 19: Style nit: typically in Dart you don't explicitly `show` the names you use.

nex3 (changes requested):
I think this is a good concept, but it doesn't take into account running the compiler from the Dart or JS APIs or through the embedded protocol. This will require a proper [proposal](https://github.com/sass/sass/blob/main/CONTRIBUTING.md).

Joshix-1:
> This will require a proper [proposal](https://github.com/sass/sass/blob/main/CONTRIBUTING.md).
Why? I don't see this as a new sass feature, as it doesn't really introduce anything new. The "random" algorithm seems to be an implementation detail. This only adds a flag to configure the random algorithm a bit.

nex3:
Anything that's configurable via the CLI also needs to be configurable through the API, and any changes to the JS API need a language proposal because it's part of the Sass specification.

nex3:
Since it's been three weeks, I'm going to close this out. We may end up resurrecting it if/when we go through the full proposal process for adding a new configuration option.

PR #2372, opened 2024-10-01
Fix srgb to linear

ntkme:
In the edge case of `channel = 0.04045` the error is more than 1e⁻⁹, much higher than the required precision in Sass.

https://www.w3.org/TR/css-color-4/#color-conversion-code
https://github.com/color-js/color.js/blob/40e7a059c639bafde14504627e62791588c63100/src/spaces/srgb.js#L38


ntkme:
Why this PR matters in the edge case with exact value at cutoff point for the current implementation, and possible improvement: https://github.com/color-js/color.js/issues/596

PR #2375, opened 2024-10-02
Use correct notation for multiple denominator units

Goodwine:
See: https://physics.nist.gov/cuu/Units/checklist.html

Tests: https://github.com/sass/sass-spec/pull/2027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 * Fix a slight inaccuracy case when converting to `srgb-linear` and
   `display-p3`.
 
+* **Potentially breaking bug fix:** `math.unit()` now joins denominator units
+  with a division operator (`/`) instead of a multiplication operator (`*`). For
+  example, `px/em/em` instead of `px/em*em`.
+
 ### JS API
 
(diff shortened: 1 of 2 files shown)


nex3:
I'm not entirely convinced this is better. We can't realistically match the SI standards here—I don't think users would really understand `px*em⁻¹*rad⁻¹*s⁻¹` even if we could rely on Unicode characters rendering well everywhere. `px*em/rad/s` is explicitly called out as *not* recommended. We could do `px*em/(rad*s)`, which does seem to be one of the allowed SI formats and is more clear to a naive reader.

Goodwine:
Oops, I only skimmed through the doc and didn't actually read it lol

I like using parentheses in this case, I'll update the PR

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 1.79.6
```
nex3, line 1: We actually haven't released 1.79.5 yet, so you might as well put this in there too.

nex3 (approved):

CHANGELOG.md
Goodwine: oks, I will do that then :)

PR #2376, opened 2024-10-03
Add `BooleanExpression` and `NumberExpression` to sass-parser

Goodwine:


pkg/sass-parser/lib/src/expression/number.ts
```diff
@@ -0,0 +1,100 @@
...
+import {Expression} from '.';
+
+/**
+ * The initializer properties for {@link NumberExpression}.
+ *
+ * @category Expression
+ */
+export interface NumberExpressionProps {
+  value: number;
+  unit?: string;
```
Goodwine, line 19: FYI - I was thinking of adding numerator/denominator units (this is what prompted the other PR https://github.com/sass/dart-sass/pull/2375)

However, a user can never write `1px*px`, or in other words, Sass/SCSS code can never have a complex unit. It is used by Sass internally, yes, but it's not valid Sass code. We can only ever have a single unit here, or no unit at all. Is this correct?

pkg/sass-parser/lib/src/expression/number.ts
```diff
@@ -0,0 +1,100 @@
...
+  raws?: NumberExpressionRaws;
+}
+
+/**
+ * Raws indicating how to precisely serialize a {@link NumberExpression}.
+ *
+ * @category Expression
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface -- No raws for a number expression yet.
+export interface NumberExpressionRaws {}
```
Goodwine, line 29: For both Booleam and Number.. I don't think we have raws, unless we want to have a `before`. But I based this on `StringExpression` which doesn't. I left them empty in here to be consistent, but it smells of YAGNI.

Should I remove these if they are empty?

pkg/sass-parser/lib/src/expression/number.ts
nex3: Yep, that's right. No need for complex unit literals.

pkg/sass-parser/lib/src/expression/number.ts
nex3: For `NumberExpression`, we should have a `{raw: string, value: string}`-style raw for the number's value, because the same parsed number can be represented in different ways. For example, `1` can be represented as `1.0`, `01`, and so on.

For `BooleanExpression`, as I mentioned in chat, we may want to add general support for `ExpressionRaws.before`. Let's keep this in place for now, and we can remove it later if we decide it won't be useful.

nex3 (changes requested):

pkg/sass-parser/lib/src/expression/number.ts
Goodwine: As discussed in chat, I'll leave the `before`/`after` for all expressions as out of scope for this PR, and I only added the number raw value.

That said, I wasn't sure why is `{raw: string, value: string}` needed. Can this use `{value: string}` only as Raws? or what case do you have in mind where `raw` and `value` (in `raws`) would have different values?

nex3: Sorry, for a number it should be `{raw: string, value: number}`. The point is to only apply the raw string if the value matches the actual value of the expression. For example, suppose you have a PostCSS plugin that rounds all numbers to 3 decimal places. The number expression would be

```js
{
  sassType: "number",
  value: 0.1234567,
  raws: {raw: ".1234567", value: 0.1234567}
}
```

Since this plugin doesn't care about formatting, it doesn't bother setting raws. It just sets the `number.value = 0.123`. The serializer then checks whether `number.raws.value === number.value`, and if it doesn't it ignores the raw. This way, it avoids overwriting the modified value with the original raw.

pkg/sass-parser/lib/src/expression/number.test.ts
```diff
@@ -0,0 +1,192 @@
+// Copyright 2024 Google Inc. Use of this source code is governed by an
+// MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+import {Value} from 'sass';
```
nex3, line 5: Please fix.

nex3 (changes requested):

pkg/sass-parser/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 0.2.6
 
-* No user-visible changes.
+* Add `BooleanExpression` and `NumberExpression`.
```
Goodwine, line 3: I can update this to 0.2.7 if this version has already come out

nex3: Yeah, I just cut a release for it.

pkg/sass-parser/lib/src/expression/boolean.test.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    describeNode('parsed', () => utils.parseExpression('true'));
+
+    describeNode(
+      'constructed manually',
+      () =>
+        new BooleanExpression({
+          value: true,
+        })
```
nex3, line 32: Nit:

```suggestion
      () => new BooleanExpression({value: true})
```

Prettier won't do this for you, unfortunately. Also below.

pkg/sass-parser/lib/src/expression/boolean.test.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    describeNode('constructed from ExpressionProps', () =>
+      utils.fromExpressionProps({
+        value: false,
+      })
+    );
+  });
+
+  describe('assigned new', () => {
+    beforeEach(() => void (node = utils.parseExpression('true')));
```
nex3, line 74: Not much reason to have this be a `describe()` with its own `beforeEach()` rather than just a single `it()`.

pkg/sass-parser/lib/src/expression/number.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+   * The numeric value for the raw string that represents the number.
+   *
+   * This `value` is compared with {@link NumberExpression.value}. If they
+   * match, then the {@link raw} string is output when stringifying the
+   * expression. Otherwise, the {@link raw} string is ignored, and the
+   * {@link NumberExpression.value} is output.
+   */
+  value?: number;
```
nex3, line 46: These should be in a single object, not at the root of the raws. Once https://github.com/sass/dart-sass/pull/2389 lands, you can use `RawWithValue` for this.

nex3 (changes requested):

pkg/sass-parser/lib/src/expression/number.test.ts
```diff
@@ -0,0 +1,210 @@
...
+          }).toString()
+        ).toBe('123'));
+
+      it('with a the same value as the expression but falsy format', () =>
+        expect(
+          new NumberExpression({
+            value: 123,
+            raws: {number: {raw: '', value: 123}},
+          }).toString()
+        ).toBe(''));
```
nex3, line 147: I don't know that these are really worth testing. The former should be forbidden by TypeScript and there's not a strong reason to think the latter would be an issue.

pkg/sass-parser/lib/src/expression/number.ts
```diff
@@ -0,0 +1,112 @@
...
+export interface NumberExpressionRaws {
+  /**
+   * The raw string representation of the number.
+   *
+   * Numbers can be represented with or without leading and trailing zeroes, and
+   * use scientific notation. For example, the following number representations
+   * have the same value: `1e3`, `1000`, `01000.0`.
+   */
+  // TODO: Replace with RawWithValue<T> when #2389 lands.
+  number?: {raw?: string; value?: number};
```
nex3, line 37: I'd name this `value` to match the property on `NumberExpression`.

Also, `raw` and the inner `value` should be required.

nex3 (approved):

pkg/sass-parser/lib/src/expression/number.ts
Goodwine: Yeah, I wasn't too sure and I ended up changing it. I think it should be `value` as you suggest to match `NumberExpression` (both here and in the sass_api), but I was getting confused with `raws.value.value` so I changed it. I'll change it back to `value` and we can revisit this later if it's too confusing.

PR #2379, opened 2024-10-09
Partially replace chokidar with @parcel/watcher

ntkme:
See discussion from #2347.

This PR uses `@parcel/watcher` for watch with `polling: false`, and uses `chokidar` v4 for watching with `polling: true`, in hope to address some short comings of both chokidar v3 and chokidar v4.


nex3:
I'm not thrilled about keeping on a hefty dependency _just_ to support polling, but I suppose this is still better than the status quo.

nex3 (approved):

PR #2382, opened 2024-10-11
Update QEMU

ntkme:
So far riscv64 release has been a bit flaky after we switched to native aot build on QEMU. Attempt to upgrade QEMU to see if it will help make these more stable.

https://github.com/sass/dart-sass/actions/runs/10912255003/job/30287343660
https://github.com/sass/dart-sass/actions/runs/11283989988/job/31384720406

### DIFF
--- a/.github/workflows/build-linux-musl.yml
+++ b/.github/workflows/build-linux-musl.yml
@@ -35,6 +35,8 @@ jobs:
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
+        with:
+          image: tonistiigi/binfmt:master
 
       - name: Compile Protobuf
         run: |
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -35,6 +35,8 @@ jobs:
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
+        with:
+          image: tonistiigi/binfmt:master
 
       - name: Compile Protobuf
         run: |


ntkme:
@nex3 As for yesterday's failed release, we can still try re-run failed jobs. Can you please help do that? https://github.com/sass/dart-sass/actions/runs/11283989988

nex3:
I've restarted the old workflow (odd that it didn't send me an email about the failure).

Can you help me understand exactly what this change is doing? Is docker/setup-qemu-action out-of-date in a way that tonistiigi/binfmt isn't?

ntkme:
https://github.com/docker/setup-qemu-action?tab=readme-ov-file#inputs

By default `setup-qemu-action` uses `tonistiigi/binfmt:latest`, which has been staled at QEMU `7.0.0` for years due to concerns over some very minor potential breaking change. By switching to `tonistiigi/binfmt:master`, it gives us QEMU `8.1.5` as of writing.

ntkme:
@nex3 I just check the re-run and saw it was already taking way longer than expected. It might be stuck again. We can try stop the current run and re-run it one more time. Or maybe retry with the next release on QEMU 8.1.5 and see if that's better.

I also found someone else hitting this issue: https://github.com/dart-lang/sdk/issues/56671#issuecomment-2343126326

Note the linked issue is mainly about a different crash, however, in that specific linked comment the author mentioned:

> I'd note that the same sequence (and more) runs inside of Docker buildx (hence using QEMU) every time we release binaries, and whilst it hangs far too frequently it doesn't crash in this manner.

So this does seems to be an issue either with QEMU or dart-sdk.

nex3:
I think this is causing https://github.com/sass/dart-sass/actions/runs/11375079152/job/31645316320—unless you know of a quick fix, we may have to disable riscv64 builds for now.

ntkme:
@nex3 Can you please retry it? It looks like a docker.io server issue. Maybe docker.io rate limit or simply a server glitch.

nex3:
Okay, yeah, it looks like the retry is running okay.

PR #2387, opened 2024-10-14
Fix GitHub Actions config

jathak:
This updates our dependabot config to be compatible with new format restrictions and passes `CHROME_DEVEL_SANDBOX` when running browser tests to avoid a new sandbox that's incompatible with Ubuntu (see [here](https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md#option-3_the-safest-way))

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -304,9 +304,10 @@ jobs:
       with:
         dart-sdk: ${{ matrix.dart_channel }}
         github-token: ${{ github.token }}
-
+    
     - run: dart run grinder pkg-npm-dev
       env: {UPDATE_SASS_SASS_REPO: false}
+    - run: sudo chmod 4755 /opt/google/chrome/chrome-sandbox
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -307,10 +307,12 @@ jobs:
 
     - run: dart run grinder pkg-npm-dev
       env: {UPDATE_SASS_SASS_REPO: false}
+    - run: sudo chmod 4755 /opt/google/chrome/chrome-sandbox
     - name: Run tests
       run: dart run test -p chrome -j 2
       env:
         CHROME_EXECUTABLE: chrome
+        CHROME_DEVEL_SANDBOX: /opt/google/chrome/chrome-sandbox
```
nex3, line 315: Can you link the relevant Chrome documentation here as well?

nex3 (approved):

Issue #2390, opened 2024-10-17
Parse-time deprecation messages are not properly rate limited


ntkme:
```
$ npx sass --version
1.80.0 compiled with dart2js 3.5.3

$ npx sass --no-verbose node_modules/bootstrap/scss/bootstrap.scss 2>&1 | grep 'Sass @import rules are deprecated and will be removed in Dart Sass 3.0.0.' | wc -l
      87
```

When compiling latest bootstrap, the same `@import` deprecation is printed 87 times, where it should only be printed 5 times and then get summarized.

ntkme:
https://github.com/sass/dart-sass/blob/ab19f94b144ca51c62416e1a6aca6c9a90dcbd23/lib/src/import_cache.dart#L360-L379

@jathak This seems to be a deliberate choice but I'm afraid that this is way too noisy for large framework like bootstrap.

nex3:
I agree that we should make sure this case uses a repetition-limiting logger. Probably from the CLI, we should manually construct a shared logger—or if we're running in watch mode, a separate shared logger for each batch of events.

ntkme:
> Probably from the CLI, we should manually construct a shared logger—or if we're running in watch mode, a separate shared logger for each batch of events.

In fact, we have the very same issue in both CLI and API due to the `DeprecationProcessingLogger` inside `ImportCache` is hard coded with `limitRepetition: false`. The JS API specifically mentions the behavior of `verbose` option here: https://sass-lang.com/documentation/js-api/interfaces/options/#verbose

> By default, Dart Sass will print only five instances of the same deprecation warning per compilation to avoid deluging users in console noise. If you set verbose to true, it will instead print every deprecation warning it encounters.

nex3:
After consulting with @jathak, the current plan here is to move away from emitting any deprecation warnings at parse time at all. Instead, the parser will accumulate a list of deprecations and attach it to a hidden field in the `Stylesheet` object, and the evaluator will then emit those warnings when the stylesheet is loaded. This will allow us to avoid needing to have the `ImportCache` handle a logger at all.

As a side effect of this, we can (Dart-level) deprecate the `logger` parameter to the `ImportCache` as well as various parse functions.

PR #2391, opened 2024-10-17
Always limit the repetition of loggers

nex3:
See sass/dart-sass#2390. This is definitely not the correct long-term
solution—we only want to limit repetition in the context of a single
batch of compilations, and this doesn't always respect `--verbose`—but
it's way better to display too few deprecations than too many.

See https://github.com/sass/sass-spec/pull/2030

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.80.1
+
+* Fix a bug where repeated deprecation warnings were not automatically limited.
+
 ## 1.80.0
 
 * `@import` is now officially deprecated, as are global built-in functions that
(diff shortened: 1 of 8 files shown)


nex3:
I'm planning to land this TBR and cut a release tonight so we don't flood users with deprecation warnings for `@import`.

nex3:
@ntkme Since you seem to be around, can you give this and https://github.com/sass/sass-spec/pull/2030 a quick check so they have some approval?

ntkme (approved):
This PR looks good to me. I have a question on the spec change, which I commented there.

jathak (commented):
LGTM

PR #2398, opened 2024-10-18
Fix `@import url("...")` in plain CSS

### LINKED ISSUE #2397: type 'InterpolatedFunctionExpression' is not a subtype of type 'StringExpression'
After upgrading from 1.79.4 to 1.79.5, I am getting the following error in my Rails app when running `yarn build:css`:

```
Unexpected exception:
TypeError: Instance of 'InterpolatedFunctionExpression': type 'InterpolatedFunctionExpression' is not a subtype of type 'StringExpression'


node_modules/sass/sass.dart.js 2314:43   Object.wrapException
node_modules/sass/sass.dart.js 3783:15   Object._failedAsCheck
node_modules/sass/sass.dart.js 3772:9    Rti._generalAsCheckImplementation
node_modules/sass/sass.dart.js 57728:39  CssParser._cssImportRule$1
node_modules/sass/sass.dart.js 57697:22  CssParser.atRule$2$root
node_modules/sass/sass.dart.js 59839:22  CssParser._statement$1$root
node_modules/sass/sass.dart.js 63051:17  StylesheetParser_parse__closure.call$0
node_modules/sass/sass.dart.js 59368:30  CssParser.statements$1
node_modules/sass/sass.dart.js 63035:23  StylesheetParser_parse_closure.call$0
node_modules/sass/sass.dart.js 58609:25  CssParser.wrapSpanFormatException$1$1
error Command failed with exit code 255.
```

I have confirmed that I get the same error for any version > 1.79.4.

nex3:
Closes #2397
See sass/sass-spec#2033

### DIFF
--- a/lib/src/parse/css.dart
+++ b/lib/src/parse/css.dart
@@ -7,6 +7,7 @@ import 'package:string_scanner/string_scanner.dart';
 
 import '../ast/sass.dart';
 import '../functions.dart';
+import '../interpolation_buffer.dart';
 import 'scss.dart';
 
 /// The set of all function names disallowed in plain CSS.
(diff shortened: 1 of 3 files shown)


lib/src/parse/css.dart
```diff
@@ -86,16 +87,38 @@ class CssParser extends ScssParser {
   ImportRule _cssImportRule(LineScannerState start) {
     var urlStart = scanner.state;
     var url = switch (scanner.peekChar()) {
-      $u || $U => dynamicUrl() as StringExpression,
+      $u || $U => switch (dynamicUrl()) {
+          StringExpression string => string.text,
+          InterpolatedFunctionExpression(
```
ntkme, line 92: I think this would work for most of the cases. However, this would attempt to parse a "interpolated" string. In pure css, technically I can do `@import url("a.css#{url-with-a-weird-hash}");` and it should still be valid, but I suppose this will throw interpolation is not allowed in css?

ntkme: In other words, strings in plain css should probably never be parsed as interpolated, even if they contain sass interpolation syntax.

nex3: This code isn't actually parsing anything, it's just handling the various formats that the `dynamicUrl()` function can return. But yes, interpolations will throw errors in plain CSS mode even when they are technically valid CSS—largely because LibSass incorrectly interpreted `.css` files as SCSS, so we want to highlight places where that's not working. It might be worth revisiting that decision for Dart Sass 3.0.0, but probably not for 2.0.0.

PR #2399, opened 2024-10-20
Enable AOT build for android-riscv64

ntkme:
This PR replaces `qemu-user-static` binaries distributed by `tonistiigi/binfmt` with `fedora` distributed ones.

To properly run android riscv64 QEMU >= 8.2.0 is needed, but `tonistiigi/binfmt:master` has been stuck at QEMU 8.1.5 for a very long time (8.2.0 was released on 20 Dec 2023): https://github.com/tonistiigi/binfmt/issues/197. Instead, current stable fedora 40 has QEMU 8.2.7 and upcoming fedora 41 has QEMU 9.1.0. Both version runs dart for android riscv64 without issues.

In addition, the new QEMU version fixed the bug that dart arm linux-musl would get stuck, so that this PR changes it back to run all linux builds with QEMU for simplicity.

Although dart-sdk still has no support for [cross compiling AOT snapshot](https://github.com/dart-lang/sdk/issues/28617), with this PR we would reach complete coverage for AOT on all supported platforms.


ntkme:
Screenshot 2024-10-20 at 2 16 33 PM

jathak (approved):

PR #2403, opened 2024-10-23
Add sass-parser support for variable declarations and `@warn` rules

nex3:


pkg/sass-parser/lib/src/statement/variable-declaration.test.ts
```diff
@@ -0,0 +1,619 @@
...
+    ): void {
+      describe(description, () => {
+        beforeEach(() => (node = create()));
+
+        it('has a type', () => expect(node.type.toString()).toBe('decl'));
+
+        it('has a sassType', () =>
+          expect(node.sassType.toString()).toBe('variable-declaration'));
+
+        it('has no namespace', () => expect(node.namespace).toBe('baz'));
```
Goodwine, line 103:
```suggestion
        it('has a namespace', () => expect(node.namespace).toBe('baz'));
```

pkg/sass-parser/lib/src/statement/variable-declaration.ts
```diff
@@ -0,0 +1,228 @@
...
+  /** @hidden */
+  toString(): string {
+    return (
+      this.prop +
+      (this.raws.between ?? ': ') +
+      this.expression +
+      (this.raws.flags?.value?.guarded === this.guarded &&
+      this.raws.flags?.value?.global === this.global
+        ? this.raws.flags.raw
+        : (this.guarded ? ' !default' : '') + (this.global ? ' !global' : '')) +
```
Goodwine, line 219: Should there be documentation on the order in which these are rendered in the case where both are true but raws are empty or mismatch?

Goodwine (approved):

pkg/sass-parser/lib/src/statement/variable-declaration.ts
nex3: Nah, I don't think documenting the default rendering is important when it's easy to control if you want to.

PR #2410, opened 2024-10-25
Add sass-parser support for the `@while` rule

nex3:


pkg/sass-parser/lib/src/stringifier.ts
```diff
@@ -165,18 +166,22 @@ export class Stringifier extends PostCssStringifier {
...
+
   /** Helper method for non-generic Sass at-rules. */
   private sassAtRule(node: postcss.AtRule, semicolon?: boolean): void {
     const start =
       '@' +
       node.name +
       (node.raws.afterName ?? ' ') +
-      node.params +
-      (node.raws.between ?? '');
+      node.params;
```
Goodwine, line 179: I'm confused as to why `this.block()` doesn't need `between` to be produced first. Does it already do that

Goodwine: oh, it does!

Goodwine (changes requested):
I think you can run `npm run fix` to get rid of the reported issues. You may have to run it again if you submit after https://github.com/sass/dart-sass/pull/2415

Goodwine (approved):
Since I submitted the PR that updated `gts`, you should rerun `npm run fix`

PR #2413, opened 2024-10-26
Implement `sass --embedded` in pure JS mode

### LINKED ISSUE #2325: Implement `sass --embedded` in pure JS mode
There has been a few requests for BSD support on the sass-embedded-host-ruby:

- https://github.com/sass-contrib/sass-embedded-host-ruby/issues/228
- https://github.com/sass-contrib/sass-embedded-host-ruby/issues/227
- https://github.com/nanoc/nanoc/issues/1710
- https://github.com/jekyll/jekyll/issues/9462
- https://github.com/jekyll/jekyll/issues/9493

The most reasonable solution seems to be implementing `sass --embedded` in pure JS mode so that we can run embedded compiler using Node.js runtime on platforms that Dart runtime has no support. Aside from BSD, this will also allow users of some minor CPU architecture to run embedded host. 

The pure JS implementation should be very similar to the current Dart implementation:

- Use synchronous compilation only for better performance.
- Use Node.js worker threads to replace Dart isolates.
- Use `SyncMessagePort` from embedded-host-node to replace `Mailbox`.

ntkme:
Closes #2325.

## Implementation

The actual isolate dispatcher and compilation dispatcher are nearly unchanged. However, I had to replace isolate with worker communication, and mock tons of small things that do not work on node.

## Testing

- All Dart embedded tests are passing. - GitHub CI has been updated to run these in this PR.
- All JS API tests are passing. - GitHub CI has been updated to run these in this PR.
- All Ruby API tests are passing.

## Current Status

Functionally it's ready to merge. Cosmetically it's waiting on dart-sass (depends on node_interop) to migrate off the legacy js interop to avoid the confusion of mixing legacy and new js interop, which is currently blocked by https://github.com/dart-lang/sdk/issues/63403

Because FreeBSD support has been requested a lot for ruby's sass-embedded gem, this patch has been bundled within sass-embedded gem.


ntkme:
This creates some intertwined dependencies between `sass`'s dart2js code and `sass-embedded`'s typescript code. I would like to untangle it, but with the current repo structure it's difficult.

I would like to purpose that we merge embedded-host-node as a child package into dart-sass mono repo, just like sass-parser is a child package. This will make it more flexible to bundle and ship artifacts that uses dart2js and typescript together. @nex3 What do you think?

nex3:
Some initial thoughts:

* Are we confident this is simpler than writing an embedded compiler implementation directly in TS using the `sass` package? That would duplicate some of the dispatcher logic, but it would mitigate a lot of the JS interop overhead. (I'm not saying that's definitely the right way, just that we should at least talk through the possibility.)

* The infrastructure here could make https://github.com/sass/dart-sass/issues/868 easier.

* Can you provide a list of the "small things that do not work on node"?

* I can start the process of pulling out `sync-message-port`.

ntkme:
> Are we confident this is simpler than writing an embedded compiler implementation directly in TS using the sass package?

My answer is yes... To begin with, currently the compilation dispatcher (isolate/worker) cannot be implemented in pure JS due to lack of feature. Specifically, no access to "formatted" log for customized logger and `containingUrlWithoutMarking`. Because of that, we have to use the Dart API, or export the a custom Dart function to JS and run it in JS.

My first implementation is exact a customized exported function to start the compilation worker, and then a TS implementation for the worker dispatcher. The history still exist on GitHub:
- https://github.com/sass/dart-sass/commit/7084da7fa261bd6d640ed44632e4a81af8d8bc51
- https://github.com/sass/embedded-host-node/commit/ec05f2d7f9d57f987475f6d6345799001fae53b1

It was not difficult to get that to a functional state on basically level, but getting two packages connecting to each other and get testing setup working was extremely painful. The performance was extremely bad, due to lack of a `ResourcePool` thus allowing unlimited number of threads. Spawning node worker thread itself is of slow, and cost relatively high amount of resources. There is a test in js-api-spec that runs 1000 concurrent compilation, which started nearly 1000 threads and my whole system was suffering to the point it started to cause unrelated tests to timeout. Finally, although the API specs are mostly passing, the `dart test` was having tons of failures for handling `ProtocolError`, which is extremely difficult to completely align the behavior.

That's the point I decided to completely rewrite it and just switch to dart2js and reuse the existing logic. 

> That would duplicate some of the dispatcher logic, but it would mitigate a lot of the JS interop overhead.

To be honest, I don't think JS interop overhead matter that much here, because in the end what's the actual compute heavy Sass compilation happens on dart2js code anyways. The only thing that *might* have a meaningful performance impact is protobuf-es vs protobuf.dart running on dart2js, but protobuf-es is known to be slow: https://github.com/bufbuild/protobuf-es/issues/333, so probably also doesn't matter.

---

Finally, from a long term code maintainability perspective, it is easier to maintain one implementation instead of two.

ntkme:
> Can you provide a list of the "small things that do not work on node"?

By "do not work on node" I meant the dart code do not work out of box without an alternative implementation. Basically every thing in `lib/src/embedded/js/` folder:

- `stdin` / `stdout` -> `process.stdin` / `process.stdout` wrapped with `StreamController`
- `Isolate` -> `Worker`
  - The way they spawn are different: Isolates can partially share code memory stack within the same isolate group, they are lunched with a entrypoint function that does not need to re-JIT all the code, see: https://github.com/dart-lang/sdk/issues/36097. On the other hand, workers do not share code memory, that each worker load a entrypoint script - the sass.js and its dependencies are quiet heavy, thus spawning worker is a pretty high cost. This cost may potentially be reduced a tiny bit if we can split out a multiple entrypoints and lazily load dependencies not used in specific code path. However, given `sass.js` is several megabytes in size, it will not help much because it will only make the main thread load faster, but the worker thread will still suffer slow start. 
- `Isolate.exit` -> `process.exit`
- `exitCode` -> `process.exitCode`
  - `exitCode` in dart is a shared variable across all isolate, but in `process.exitCode` in node is worker thread local. To get any `exitCode` on main thread we have to listen on worker's exit event, this means: in dart we can eagerly exit on failure, but on node we have to wait for worker to transfer their exitCode to main thread and then set the exit code in main thread.
- A wrapper interface for Mailbox and SyncMessagePort to look the same for compilation isolate.

ntkme:
Here is some benchmark results on 1.80.5 native binary vs pkg-npm-release.

I tested three cases. All of them on a single persisted compiler, lines that marked with `/MT` means multi-threaded (10 threads).

1. Minimal Sass Input - compile `a {b: c}` 10000 times.
2. High Round Trips - invoke host function 10 times per compilation, and run 1000 compilations.
3. Bootstrap - compile bootstrap 50 times.

On single-threaded test Dart is much faster than pure JS, especially in the compute heavy bootstrap test. However, in nearly all multi-threaded test, pure JS is not far from Dart, which is actually great.

I think the previous performance issues I had 

```
sass-embedded	1.80.5	(Embedded Host)	[Ruby]
dart-sass	1.80.5	(Sass Compiler)	[JavaScript]
--- Minimal Sass Input ---
                       user     system      total        real
sass-embedded      0.304688   0.081796   0.386484 (  2.195485)
sass-embedded/MT   0.546339   0.260653   0.806992 (  1.260589)
--- High Round Trips ---
                       user     system      total        real
sass-embedded      0.268202   0.084800   0.353002 (  2.135437)
sass-embedded/MT   0.316647   0.207332   0.523979 (  0.558180)
--- Bootstrap ---
                       user     system      total        real
sass-embedded      0.013900   0.006100   0.020000 ( 27.595710)
sass-embedded/MT   0.023358   0.012144   0.035502 (  3.378303)
```

```
sass-embedded	1.80.5	(Embedded Host)	[Ruby]
dart-sass	1.80.5	(Sass Compiler)	[Dart]
--- Minimal Sass Input ---
                       user     system      total        real
sass-embedded      0.255831   0.066576   0.322407 (  1.207566)
sass-embedded/MT   0.323200   0.218337   0.541537 (  0.578477)
--- High Round Trips ---
                       user     system      total        real
sass-embedded      0.197848   0.069268   0.267116 (  0.840963)
sass-embedded/MT   0.256294   0.226492   0.482786 (  0.521811)
--- Bootstrap ---
                       user     system      total        real
sass-embedded      0.013402   0.005420   0.018822 (  7.295685)
sass-embedded/MT   0.015503   0.008968   0.024471 (  2.538047)
```

<details>
<summary>
Benchmark code
</summary>

``` ruby
require 'benchmark'
require 'sass-embedded'

puts Sass.info

BOOTSTRAP = Gem::Specification.find_by_name('bootstrap').gem_dir + '/assets/stylesheets/_bootstrap.scss'

def run(n:, thread: 10, use_thread: false)
  if use_thread
    queue = Queue.new
    n.times do
      queue << nil
    end
    list = []
    thread.times do
      list << Thread.new do
        loop do
          queue.pop(true)
          yield
        rescue ThreadError
          break
        end
      end
    end
    list.each(&:join)
    queue.close
  else
    n.times do
      yield
    end
  end
end

def sass_embedded_minimal(use_thread: false)
  run(n: 10000, use_thread:) do
    Sass.compile_string('a {b: c}', style: :compressed, logger: Sass::Logger.silent).css
  end
end

def sass_embedded_round_trips(use_thread: false)
  run(n: 1000, use_thread:) do
    Sass.compile_string('@for $i from 1 through 10 { a:nth-child(#{$i}) { b: foo(); } }', functions: { 'foo()' => ->(_) { Sass::Value::String.new('test') } }, style: :compressed, logger: Sass::Logger.silent).css
  end
end

def sass_embedded_bootstrap(use_thread: false)
  run(n: 50, use_thread:) do
    Sass.compile(BOOTSTRAP, style: :compressed, logger: Sass::Logger.silent).css
  end
end

puts '--- Minimal Sass Input ---'
Benchmark.bm(16) do |x|
  GC.start

  x.report('sass-embedded') do
    sass_embedded_minimal
  end

  GC.start

  x.report('sass-embedded/MT') do
    sass_embedded_minimal(use_thread: true)
  end
end

puts '--- High Round Trips ---'
Benchmark.bm(16) do |x|
  GC.start

  x.report('sass-embedded') do
    sass_embedded_round_trips
  end

  GC.start

  x.report('sass-embedded/MT') do
    sass_embedded_round_trips(use_thread: true)
  end
end

puts '--- Bootstrap ---'
Benchmark.bm(16) do |x|
  GC.start

  x.report('sass-embedded') do
    sass_embedded_bootstrap
  end

  GC.start

  x.report('sass-embedded/MT') do
    sass_embedded_bootstrap(use_thread: true)
  end
end
```
</details>

nex3:
This PR _is_ on my radar, it's just been a very review-heavy few weeks for me and this is unfortunately on the low end of the priority spectrum.

ntkme:
GitHub Actions revealed an interesting race condition - running `postMessage()` immediately followed by `process.exit()` in the worker may not exit immediately, and while the worker has not finished exiting, the main thread could call `worker.terminate()` based on the received message, which will cause the worker to have `exitCode = 1` instead of whatever `exitCode` was set by worker.

To overcome this issue, I decided to use the first byte we added to the message buffer to explicitly send the `exitCode` to the main thread, so we don't need to wait for the unreliable async exit event from worker.

ntkme:
Ok, one more quirk for node worker. The way stdout/stderr stream in the worker thread works is that it post message to the main thread and then the main thread writes the message asynchronously. So we cannot eagerly exit the main isolate, or otherwise the error message that the worker printed to stderr might get lost.

ntkme:
@nex3 I can see that the team is focusing most of the time on the postcss sass parser and this pull request has low priority.

However, I would like to get this done in next few month before dart 3.8 become stable so that we can offer this as a replacement for dart ia32. Please take a look when you get a chance and let me know if you have any questions.

verm:
Would really be great to see this merged to have embedded support on FreeBSD even if in pure JS mode.  Thank you!

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -107,8 +106,12 @@ class IsolateDispatcher {
...
       },
       onDone: () {
-        _closed = true;
-        _allIsolates.stream.listen((isolate) => isolate.kill());
+        if (_gracefulShutdown) {
+          _closed = true;
+          _allIsolates.stream.listen((isolate) => isolate.kill());
+        } else {
+          exit(exitCode);
+        }
```
nex3, line 114: Since graceful shutdown is disabled on the Dart VM, this is changing the current behavior.

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -156,9 +159,14 @@ class IsolateDispatcher {
           _inactiveIsolates.add(isolate);
           resource.release();
           _channel.sink.add(packet);
-        case 2:
+        default:
           _channel.sink.add(packet);
-          exit(exitCode);
+          exitCode = category;
+          if (_gracefulShutdown) {
+            _channel.sink.close();
```
nex3, line 166: Shouldn't this also send kill messages to the other isolates?

lib/src/embedded/isolate_dispatcher.dart
```diff
@@ -188,7 +196,11 @@ class IsolateDispatcher {
       compilationId ?? errorId,
       handleError(error, stackTrace, messageId: messageId),
     );
-    _channel.sink.close();
+    if (_gracefulShutdown) {
+      _channel.sink.close();
+    } else {
+      exit(exitCode);
+    }
```
nex3, line 203: This is also changing the current behavior for the Dart VM.

lib/src/embedded/js/reusable_isolate.dart
```diff
@@ -0,0 +1,103 @@
...
+import 'dart:typed_data';
+
+import 'js.dart';
+import 'sync_message_port.dart';
+import 'worker_threads.dart';
+
+/// The entrypoint for a [ReusableIsolate].
+///
+/// This must return a Record of filename and argv for creating the Worker.
+typedef ReusableIsolateEntryPoint = (String, JSArray<JSAny?>) Function();
```
nex3, line 16: Two implementations of the same library should have exactly the same public interface—they shouldn't differ in which types they export.

lib/src/embedded/js/executable.dart
```diff
@@ -0,0 +1,27 @@
...
+void main(List<String> args) {
+  if (parseOptions(args)) {
+    if (isMainThread) {
+      IsolateDispatcher(StreamChannel.withGuarantees(stdin, stdout,
+                  allowSinkErrors: false)
+              .transform(lengthDelimited))
+          .listen();
+    } else {
+      var port = workerData! as MessagePort;
+      isolateMain(JSSyncReceivePort(port), JSSendPort(port));
```
nex3, line 24: The distinction between this `isolateMain` and the one that's passed to `ReusableIsolate.spawn()` is very confusing. Given that _neither_ is consistently spawned as the main method of a new worker, it would be clearer to rename both to something that more explicitly describes what they do.

lib/src/embedded/isolate_dispatcher.dart
nex3, line 1: The term "isolate" is very Dart-VM-specific. Now this code is generic across multiple implementations, I think we should use the more general term "worker" instead.

nex3 (commented):
Finally managed to carve out some time for this. I haven't done a complete review yet, but this should be enough to get you started.

.github/workflows/test.yml
```diff
@@ -168,19 +184,13 @@ jobs:
...
 
       - name: Version info
-        run: |
-          path=embedded-host-node/dist/lib/src/vendor/dart-sass/sass
-          if [[ -f "$path.cmd" ]]; then "./$path.cmd" --version
-          elif [[ -f "$path.bat" ]]; then "./$path.bat" --version
-          elif [[ -f "$path.exe" ]]; then "./$path.exe" --version
-          else "./$path" --version
-          fi
+        run: node dist/bin/sass.js --version
```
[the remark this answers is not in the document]
ntkme: Yes, it works in both mode - this is the launcher wrapper script: https://github.com/sass/embedded-host-node/blob/main/bin/sass.ts

lib/src/embedded/isolate_dispatcher.dart
ntkme: "Graceful" here does not mean wait for all inflight request to be completely, - it means wait for child isolates/workers to shutdown first before shutting down main isolate/thread.

In other words, it's only "graceful" for the dart or node vm. This is only required for node, see this earlier comment for explanation: https://github.com/sass/dart-sass/pull/2413#issuecomment-2480235837

From a end user perspective nothing changes, - we have always been "non-graceful".

lib/src/embedded/isolate_dispatcher.dart
ntkme: That is handled in `onDone` hook for the `_channel`, so after sink is closed, `kill` is sent.

lib/src/embedded/isolate_dispatcher.dart
ntkme: Actually it's not changing any behavior for end users, see https://github.com/sass/dart-sass/pull/2413#discussion_r2023894593

lib/src/embedded/js/reusable_isolate.dart
ntkme: I understand this is ugly, but I cannot think of a better way to abstract this - the fundamental problem is that the way dart launches isolates and js launches workers are way too different. In dart VM you just launch the isolate with a reference to a function, but in JS you need to give it an entrypoint script filename, and arguments.

My end goal was that for `ReusableIsolate` class itself to share same interface, even if the type of entrypoint argument is different.

Let me know if you have a better idea.

lib/src/embedded/executable.dart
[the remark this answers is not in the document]
ntkme: Done

lib/src/embedded/executable.dart
```diff
@@ -3,40 +3,19 @@
...
-        "details.",
-      );
-      // USAGE error from https://bit.ly/2poTt90
-      exitCode = 64;
-      return;
+  if (parseOptions(args)) {
+    IsolateDispatcher(
+            StreamChannel.withGuarantees(stdin, stdout, allowSinkErrors: false)
+                .transform(lengthDelimited),
+            gracefulShutdown: false)
```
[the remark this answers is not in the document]
ntkme: Added a comment.

lib/src/embedded/js/executable.dart
ntkme: I added a section and some diagrams in `lib/src/embedded/README.md` to explain how this convoluted process works.

lib/src/embedded/compilation_dispatcher.dart
```diff
@@ -408,16 +409,16 @@ final class CompilationDispatcher {
...
+    // the [IsolateDispatcher] knows whether to treat this isolate as inactive
+    // or close out entirely.
     var packet = Uint8List(
       1 + _compilationIdVarint.length + protobufWriter.lengthInBytes,
     );
     packet[0] = switch (message.whichMessage()) {
       OutboundMessage_Message.compileResponse => 1,
-      OutboundMessage_Message.error => 2,
-      _ => 0,
+      OutboundMessage_Message.error => exitCode,
```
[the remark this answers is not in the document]
ntkme: I decided to just use a second byte for this to avoid the conflict.

nex3 (commented):
Another partial review, but I think we're making good progress!

lib/src/embedded/js/io.dart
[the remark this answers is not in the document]
ntkme: Well, I was trying to implement this with new js interop from dart 3.3 without relying on the node_interop package which heavily depends on deprecated `dart:js_util`, which will start to break at some point.

For `exitCode` I can use the existing one we have.

What about `Stdin`, `Stdout`, `Stderr`? These are effectively wrapper classes on node stdio to create an interface that's same as the dart built-in classes while it only partially implement necessarily methods we need on those interface. @nex3 What's your preference here? Create another interface that wraps the original dart interface?

lib/src/embedded/js/js.dart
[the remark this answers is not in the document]
ntkme: Again, this is a saparate file because it's new js interop from dart 3.3, which I found a bit awkward to put it together with the classic interop code in the same place. Any recommendation on how to deal with it?

ntkme:
Applied most of the feedbacks. However, I have a awkward feeling about mixing classic js interop and new js interop in the same file, which is why I have lots of things only in embedded folder instead of in the generic folder.

@nex3 I haven't changed these yet, as I want to hear what's your thought on this.

ntkme:
Another note: The reason I tried to avoid classic interop and node_introp is that I couldn't get `channel.port2` passed to `Worker()` constructor correctly with node_interop. Neither `allowInterop()` nor explicit casting worked, and it was a nightmare trying to debug what went wrong. I looked at node_interop repo trying to see if there is any example for it and there is no actual usage of transferList anywhere.

The new dart 3.3 js interop on the other hands was way more straight forward to use and debug. I got `Worker()` working on my first attempt.

ntkme:
@nex3 Pinging again regarding how to deal with legacy js package vs new js_introp:

- The legacy js package has been officially marked discontinued.
- Both the legacy js package and new js_introp provides `@JS` annotations. This issue is addressed in 0.7.0 of legacy js:
   > Breaking Change: Moved annotations to single location in dart:_js_annotations. This makes it easier to avoid collisions with dart:js_interop's `@JS`, as you will now see a static error instead of it resolving in favor of the package definition. However, since this may result in breakages in previously working code, this is treated as a breaking change.
- Ideally `node_interop` should be rewritten with new `js_interop` but nothing is happening. On the other hand, `node_interop` cannot be easily upgraded to js 0.7.x due to js 0.7.x requires dart >=3.1.x, while `node_interop`'s dev dependencies `build_node_compilers` do not support dart 3.x at all. This means `node_interop` is effectively stuck at js 0.6.x as of now, which means we have `@JS` conflict that we cannot mix old js and new js_interop code in the same file.

So honestly I think the way js interop is handled in this PR is already at best effort. - That new code only uses js_interop and it's currently isolated from the legacy js code.

Any more thoughts? Would you mind take another look at this?

nex3:
Sorry that this has taken me so long to finish reviewing, there have been a lot of things to juggle this year.

I'm working on a major change (https://github.com/sass/dart-sass/tree/js-interop) to move Dart Sass to the new JS interop system. This includes creating a new package for defining typings and wrappers for the JS core library (as well as possibly pushing some changes upstream to `dart:js_interop`) as well as updating `node_interop` which will probably include dropping its compiler dependency. You can see the work in progress on each of these at https://github.com/sass/dart_js_core and  https://github.com/pulyaevskiy/node-interop/tree/js_interop.

At this point it may make sense to wait on landing this PR until that work is finished. We're already closing in on that, so it shouldn't be too much more time.

karolyi:
Rooting for this, as I can't use embedded sass on FreeBSD.

verm:
> Rooting for this, as I can't use embedded sass on FreeBSD.

Me too as I use it on FreeBSD and it works great.

nex3:
The JS interop work is close to being done, at which point this will be unblocked.

verm:
Would just like to add on top of using this to get Discourse on FreeBSD I just got Vikunja working.  Both work flawlessly thank you for this change it's been a huge help.

ntkme:
@nex3 Any updates on the `js_interop` migration? I saw that you opened quite a lot issues in dart-sdk in this area - is there any blocker?

nex3:
The only real blocker is merging a bunch of my JS type definitions which are currently living in a private package into Dart team owned sources. I'm in discussion with @srujzs about the best way to do that, and I may pitch in to help land it. If you'd like to help as well, we could probably loop you in.

chalin:
Hi @nex. Do you have an ETA for this PR?

lib/src/embedded/worker_dispatcher.dart
```diff
@@ -188,7 +196,11 @@ class IsolateDispatcher {
       compilationId ?? errorId,
       handleError(error, stackTrace, messageId: messageId),
     );
-    _channel.sink.close();
+    if (_gracefulShutdown) {
+      _channel.sink.close();
+    } else {
+      exit(exitCode);
+    }
```
ntkme: Moved this into a helper function, and documented why it's needed.

verm:
Is this still planned to be merged?  I use this frequently on FreeBSD and it's always worked perfectly it would be fantastic to see it merged as part of the base.  Thank you as well @ntkme for maintaining this as well as you do it's greatly appreciated.

nex3:
@verm Yes, but it's blocked on getting Dart Sass working with the new Dart JS interop system, which is itself a long process.

PR #2414, opened 2024-10-26
Lazy load @parcel/watcher and fallback to chokidar

### LINKED ISSUE #2409: Can't run on arm-musl platform since 1.76.5
Context:

* I'm building a Docker container for `arm` (32 bit, not 64), on an Alpine as a base image.
* Alpine uses `musl` instead of `glibc`.
* dart-sass 1.76.5 changed to `@parcel/watcher` for filesystem events.
* `@parcel/watcher` does not currently have a build for the `linux-arm-musl` platform.

Therefore, sass fails on 32 bit arm musl platforms.

> Error: No prebuild or local build of @parcel/watcher found. Tried @parcel/watcher-linux-arm-musl. Please ensure it is installed (don't use --no-optional when installing with npm). Otherwise it is possible we don't support your platform yet.

There are a few possible solutions, I'm not sure which is best:

1. Do nothing. 32-bit is old hat, maybe this is just a "me" problem. 😄 
2. Get `@parcel/watcher` to provide an `arm-musl` build. I've sent them [a PR](https://github.com/parcel-bundler/watcher/pull/192) for that, but it might have problems which is why they've not done it.
3. Allow `sass` to use their cross-platform `watcher-wasm` backend if necessary. I think this will need a code change, it seems to have a different require line.
4. Only throw errors at runtime if watch mode is enabled. In my case, I'm doing a one-off compilation in the Docker build, so there's no need for file watching.

Any thoughts?

ntkme:
- Fixes #2409.

Depends on https://github.com/google/dart_cli_pkg/pull/169

With this change, it no longer fails on loading `sass` on CLI or API if `@parcel/watcher`'s optional dependency is not available. The error about missing optional dependency is printed once per execution only if user actually used `--watch` flag on CLI, and then it fallbacks to chokidar for that whole execution.

To test the new behavior, run `npm install --omit optional` to remove `@parcel/watcher`'s optional depenency. Then run `node build/npm/sass.js` or `node build/npm/sass.js --watch [...]` to see it in action.

I'm bumping dart-sdk requirement to 3.3.0 in order to use new `dart:js_interop`, as the legacy `package:js_interop/js_interop.dart` is giving me lots of troubles. E.g. cannot add non-static helper methods to an `@JS` class etc.

### DIFF
--- a/lib/src/io/js.dart
+++ b/lib/src/io/js.dart
@@ -257,7 +257,30 @@ Future<Stream<WatchEvent>> watchDir(String path, {bool poll = false}) async {
   // Don't assign the controller until after the ready event fires. Otherwise,
   // Chokidar will give us a bunch of add events for files that already exist.
   StreamController<WatchEvent>? controller;
-  if (poll || parcelWatcher == null) {
+  if (parcelWatcher case var parcel? when !poll) {
+    var subscription = await parcel.subscribe(path,
+        (Object? error, List<ParcelWatcherEvent> events) {
+      if (error != null) {
+        controller?.addError(error);
+      } else {
+        for (var event in events) {
+          switch (event.type) {
+            case 'create':
+              controller?.add(WatchEvent(ChangeType.ADD, event.path));
+            case 'update':
+              controller?.add(WatchEvent(ChangeType.MODIFY, event.path));
+            case 'delete':
+              controller?.add(WatchEvent(ChangeType.REMOVE, event.path));
+          }
+        }
+      }
+    });
+
+    return (controller = StreamController<WatchEvent>(onCancel: () {
+      subscription.unsubscribe();
+    }))
+        .stream;
+  } else {
     var watcher = chokidar.watch(path, ChokidarOptions(usePolling: poll));
     watcher
       ..on(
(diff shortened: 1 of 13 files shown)


lib/src/io/js.dart
```diff
@@ -287,7 +287,7 @@ Future<Stream<WatchEvent>> watchDir(String path, {bool poll = false}) async {
 
     return completer.future;
   } else {
-    var subscription = await ParcelWatcher.subscribeFuture(path,
+    var subscription = await parcelWatcher!.subscribe(path,
```
nex3, line 290: Consider swapping the order of these conditions so you can write `if (parcelWatcher case parcel when !poll) {` and avoid a non-null assertion.

pubspec.yaml
```diff
@@ -8,13 +8,14 @@ executables:
...
-  sdk: ">=3.0.0 <4.0.0"
+  sdk: ">=3.3.0 <4.0.0"
 
 dependencies:
   args: ^2.0.0
   async: ^2.5.0
   charcode: ^1.2.0
-  cli_pkg: ^2.8.0
+  cli_pkg:
+    git: https://github.com/google/dart_cli_pkg.git
```
nex3, line 18: The new version has been released as 2.11.0.

nex3 (approved):

nex3:
Actually, shouldn't this also set `@parcel/watcher` as an `optionalDependency` in `package/package.json`?

ntkme:
> Actually, shouldn't this also set @parcel/watcher as an optionalDependency in package/package.json?

Good call. Updated.

PR #2435, opened 2024-11-13
Support `@forward` in sass-parser

nex3:


nex3:
Test failures are caused by https://github.com/bufbuild/buf/issues/3469, but this should be good to review.

pkg/sass-parser/lib/src/statement/forward-rule.test.ts
```diff
@@ -0,0 +1,930 @@
...
+          it('preserves params', () =>
+            expect(clone.params).toBe(
+              '"foo" as bar-* show baz, $qux  with ($zip: "zap")',
+            ));
+        });
+      });
+    });
+  });
+
+  // Can't JSON-serialize this until we implement Configuration.source.span
```
Goodwine, line 924: Is there a github issue we can link to here?

Goodwine (approved):

pkg/sass-parser/lib/src/statement/forward-rule.test.ts
nex3: It's really just part of the general task of implementing `sass-parser`.

PR #2436, opened 2024-11-13
Add support for CSS `round()` with one unitless argument

### LINKED ISSUE #2433: Implement `round()` with no modulus
See https://github.com/sass/sass/issues/3803

nex3:
This also adds deprecation warnings for cases where a CSS calculation
is determined to need to mimic a global Sass function, to match the
global Sass function deprecation.

See sass/sass#3803
Closes #2433

See sass/sass-spec#2037

### DIFF
--- a/lib/src/value/calculation.dart
+++ b/lib/src/value/calculation.dart
@@ -6,6 +6,7 @@ import 'dart:math' as math;
 
 import 'package:charcode/charcode.dart';
 import 'package:meta/meta.dart';
+import 'package:source_span/source_span.dart';
 
 import '../deprecation.dart';
 import '../evaluation_context.dart';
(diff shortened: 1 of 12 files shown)


lib/src/value/calculation.dart
```diff
@@ -584,32 +615,54 @@ final class SassCalculation extends Value {
...
+              "interpreted as the CSS $inLegacySassFunction() calculation. "
+              "This doesn't allow unitless numbers to be mixed with numbers "
+              "with units. If you want to use the Sass function, call "
+              "math.$inLegacySassFunction() instead.\n"
+              "\n"
+              "See https://sass-lang.com/d/import",
+              Deprecation.globalBuiltin);
+          compatible = true;
+        }
+        if (compatible) {
```
pamelalozano16, line 661: Isn't it better to just make `compatible` a `const` and return
```
return operator == CalculationOperator.plus
              ? left.plus(right)
              : left.minus(right);
```
either way? Since `compatible` is set to `true` if it's false.

pamelalozano16 (approved):

lib/src/value/calculation.dart
nex3: `compatible` is only set to true if `inLegacySassFunction != null && left.isComparableTo(right)`.

PR #2446, opened 2024-11-27
Add support for `calc-size()`

nex3:
See sass/sass#3989
See sass/sass-spec#2039

### DIFF
--- a/lib/src/value/calculation.dart
+++ b/lib/src/value/calculation.dart
@@ -604,6 +604,21 @@ final class SassCalculation extends Value {
     }
   }
 
+  /// Creates an `calc-size()` calculation with the given [basis] and [value].
+  ///
+  /// The [basis] and [value] must be either a [SassNumber], a
+  /// [SassCalculation], an unquoted [SassString], or a [CalculationOperation].
+  ///
+  /// This automatically simplifies the calculation. It throws an exception if
+  /// it can determine that the calculation will definitely produce invalid CSS.
+  static SassCalculation calcSize(Object basis, Object? value) {
+    var args = [basis, if (value != null) value];
+    _verifyLength(args, 2);
+    basis = _simplify(basis);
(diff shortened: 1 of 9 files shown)


lib/src/value/calculation.dart
```diff
@@ -604,6 +604,21 @@ final class SassCalculation extends Value {
...
 
+  /// Creates an `calc-size()` calculation with the given [basis] and [value].
+  ///
+  /// The [basis] and [value] must be either a [SassNumber], a
+  /// [SassCalculation], an unquoted [SassString], or a [CalculationOperation].
+  ///
+  /// This automatically simplifies the calculation. It throws an exception if
+  /// it can determine that the calculation will definitely produce invalid CSS.
+  static SassCalculation calcSize(Object basis, Object? value) {
+    var args = [basis, if (value != null) value];
```
pamelalozano16, line 615: Should you check for types here? That they are either a `SassNumber`, a `SassCalculation`,  `SassString`, or a `CalculationOperation`?

lib/src/value/calculation.dart
```diff
@@ -604,6 +604,21 @@ final class SassCalculation extends Value {
     }
   }
 
+  /// Creates an `calc-size()` calculation with the given [basis] and [value].
+  ///
+  /// The [basis] and [value] must be either a [SassNumber], a
```
pamelalozano16, line 609: Value is optional right? So it can also be null

pamelalozano16 (approved):

lib/src/value/calculation.dart
nex3: That's checked by `_simplify()`.

lib/src/value/calculation.dart
nex3: Yes, but that's implicit from the type.

PR #2447, opened 2024-11-28
Enable useDefineForClassFields and fix affected TS files.

### LINKED ISSUE #2419: Remove the tsconfig option `useDefineForClassFields` and fix relevant TS files
The recent update on `gts` included new settings for `tsconfig.json`. Specifically, it turned on `useDefineForClassFields`. And this changes the Typescript behavior in how/when class properties get initialized.

Related: https://github.com/microsoft/TypeScript/issues/45995#issuecomment-2291606365

The task for this Issue with regards to `sass-parser` is to find all properties initialized with `!`, and replace them with a `declare` statement, for example:
```ts
class X {
  // BAD:
  _foo!: string;
}
```
```ts
class X {
  // GOOD:
  declare _foo: string;
}
```
Otherwise, the observed behavior for our PostCSS subnode classes is that all the values get reset after they get initialized to their "default" values.

Goodwine:
Fixes #2419

Turning off this TS compiler option is deprecated, and it will be removed "soon". It's annoying.. but oh well.

### DIFF
--- a/pkg/sass-parser/lib/src/configured-variable.ts
+++ b/pkg/sass-parser/lib/src/configured-variable.ts
@@ -100,7 +100,7 @@ export class ConfiguredVariable extends Node {
    * This is the parsed and normalized value, with underscores converted to
    * hyphens and escapes resolved to the characters they represent.
    */
-  name!: string;
+  declare name: string;
 
   /** The expresison whose value the variable is assigned. */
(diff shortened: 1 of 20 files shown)


nex3 (approved):
Please fix the lint error, then LGTM.

PR #2452, opened 2024-12-02
Expose `NodePackageImporter` in sass_api

jathak:
This will allow the Sass migrator to support `pkg:` URLs.

### DIFF
--- a/pkg/sass_api/pubspec.yaml
+++ b/pkg/sass_api/pubspec.yaml
@@ -2,7 +2,7 @@ name: sass_api
 # Note: Every time we add a new Sass AST node, we need to bump the *major*
 # version because it's a breaking change for anyone who's implementing the
 # visitor interface(s).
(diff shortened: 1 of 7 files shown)


pkg/sass_api/pubspec.yaml
```diff
@@ -2,15 +2,15 @@ name: sass_api
 # Note: Every time we add a new Sass AST node, we need to bump the *major*
 # version because it's a breaking change for anyone who's implementing the
 # visitor interface(s).
-version: 14.2.0
+version: 14.2.1
```
nex3, line 5: This is adding a new feature to `sass_api`, so it should be `14.3.0`

jathak: Whoops. Had it as a minor version originally, but accidentally did a patch version when I re-did it after syncing

nex3 (approved):

PR #2453, opened 2024-12-04
Add sass-parser support for `@return` and `@mixin`

nex3:


pkg/sass-parser/lib/src/statement/mixin-rule.ts
```diff
@@ -0,0 +1,161 @@
...
+    return 'mixin';
+  }
+  set name(value: string) {
+    throw new Error("MixinRule.name can't be overwritten.");
+  }
+
+  get params(): string {
+    return (
+      (this.raws.mixinName?.value === this.mixinName
+        ? this.raws.mixinName!.raw
```
Goodwine, line 100: if `this.raws = {value: 'foo'}`  and `this.mixinName = 'foo'`. couldn't this result in `@mixin undefined()`? And if so, is that OK?

Goodwine (changes requested):

pkg/sass-parser/lib/src/statement/mixin-rule.ts
nex3: That's forbidden by the types. `RawWithValue` requires both `raw` and `value` to be defined.

Goodwine (approved):

PR #2467, opened 2024-12-16
[Indented syntax improvements] Dart implementation

jamesnw:
Blocked until [proposal](https://github.com/sass/sass/blob/main/proposal/indented-syntax-improvements.md) is accepted.

- [ ] sass-spec https://github.com/sass/sass-spec/pull/2041

[skip sass-embedded]


jamesnw:
@nex3 This is ready for review. Here are some questions I wanted to flag as you review. Thanks!

1. `allowNewlines` is set to `true` in some cases, like in parser* functions or in `scss.dart`, where the value is non-impactful. Should this usage be differentiated somehow for clarity? 
2. There are some places, like in `_atRootQuery`, where `allowNewlines` is true because it is inside parentheses. I opted to not use the `_inParentheses` variable except for one place where the consumption did rely on it, in `_supportsCondition`. Would it be more idiomatic to use `_inParentheses`, even if the consumption is in the same function?
3. In the `+` include syntax, I am not supporting whitespace after the `+` as it can be part of a selector. This differs from the `=` mixin syntax- should they be the same?
4. I added paren/bracket context to `almostAnyValue`, with the only change being that they must be matched. That means the error for `a:has(d[)]` will be changed from "Expected identifer" to "Expected `]`". Since `almostAnyValue` is used for more than selectors, should I make this functionality opt-in, dependent on a parameter? Based on the places where this is used, I don't foresee issues with bracket matching, but want to verify.

lib/src/parse/stylesheet.dart
```diff
@@ -805,6 +803,7 @@ abstract class StylesheetParser extends Parser {
   ///
   /// [start] should point before the `@`.
   DebugRule _debugRule(LineScannerState start) {
+    whitespace(allowNewlines: false);
```
nex3, line 806: Because this is a Sass at-rule that's always followed by an expression, we should allow newlines after the name.

lib/src/parse/stylesheet.dart
```diff
@@ -815,19 +814,20 @@ abstract class StylesheetParser extends Parser {
...
 
     var variables = [variableName()];
-    whitespace();
+    whitespace(allowNewlines: false);
     while (scanner.scanChar($comma)) {
-      whitespace();
+      whitespace(allowNewlines: false);
       variables.add(variableName());
-      whitespace();
+      whitespace(allowNewlines: false);
```
nex3, line 826: These should all allow newlines, because none of `@each $var`, `@each $var,` and `@each $var1, $var2` are valid in isolation.

lib/src/parse/stylesheet.dart
```diff
@@ -841,6 +841,7 @@ abstract class StylesheetParser extends Parser {
   ///
   /// [start] should point before the `@`.
   ErrorRule _errorRule(LineScannerState start) {
+    whitespace(allowNewlines: false);
```
nex3, line 844: This should allow newlines.

lib/src/parse/stylesheet.dart
```diff
@@ -886,7 +889,7 @@ abstract class StylesheetParser extends Parser {
       ));
     }
 
-    whitespace();
+    whitespace(allowNewlines: false);
```
nex3, line 892: This should allow newlines since the parameter list is mandatory.

lib/src/parse/stylesheet.dart
```diff
@@ -1012,15 +1021,15 @@ abstract class StylesheetParser extends Parser {
...
       withErrorMessage("Expected variable, mixin, or function name", () {
         if (scanner.peekChar() == $dollar) {
           variables.add(variableName());
         } else {
           identifiers.add(identifier(normalize: true));
         }
       });
-      whitespace();
+      whitespace(allowNewlines: false);
     } while (scanner.scanChar($comma));
```
nex3, line 1033: It may require restructuring the loop, but we should allow newlines after a comma since ending a statement with it is not valid.

lib/src/parse/stylesheet.dart
```diff
@@ -1031,20 +1040,21 @@ abstract class StylesheetParser extends Parser {
...
     var children = this.children(child);
-    whitespaceWithoutComments();
+    whitespaceWithoutComments(allowNewlines: false);
 
     var clauses = [IfClause(condition, children)];
     ElseClause? lastClause;
 
     while (scanElse(ifIndentation)) {
-      whitespace();
+      whitespace(allowNewlines: true);
```
nex3, line 1055: We should _not_ allow newlines here, because `@else` on its own is a complete statement.

lib/src/parse/stylesheet.dart
```diff
@@ -1064,7 +1074,7 @@ abstract class StylesheetParser extends Parser {
   ImportRule _importRule(LineScannerState start) {
     var imports = <Import>[];
     do {
-      whitespace();
+      whitespace(allowNewlines: false);
```
nex3, line 1077: Eventually `@import` will be parsed as a plain-CSS at-rule, so we should be more conservative about allowing newlines to avoid a forwards-incompatibility issue.

lib/src/parse/stylesheet.dart
```diff
@@ -1181,14 +1191,13 @@ abstract class StylesheetParser extends Parser {
...
+                allowEmpty: true, allowSemicolon: true, allowNewlines: true));
             buffer.writeCharCode($rparen);
           }
-
           scanner.expectChar($rparen);
-          whitespace();
+          whitespace(allowNewlines: false);
         } else {
-          whitespace();
+          whitespace(allowNewlines: true);
```
nex3, line 1200: The statement could end here, so we shouldn't allow newlines.

lib/src/parse/stylesheet.dart
```diff
@@ -1443,8 +1458,9 @@ abstract class StylesheetParser extends Parser {
   /// [start] should point before the `@`.
   @protected
   SupportsRule supportsRule(LineScannerState start) {
+    whitespace(allowNewlines: true);
```
nex3, line 1461: Because this is a plain-CSS rule that may eventually have no special parsing, we shouldn't allow newlines here.

lib/src/parse/stylesheet.dart
```diff
@@ -1552,6 +1570,7 @@ abstract class StylesheetParser extends Parser {
   ///
   /// [start] should point before the `@`.
   WarnRule _warnRule(LineScannerState start) {
+    whitespace(allowNewlines: false);
```
nex3, line 1573: This should allow newlines.

lib/src/parse/parser.dart
```diff
@@ -67,23 +67,29 @@ class Parser {
...
 
   // ## Tokens
 
   /// Consumes whitespace, including any comments.
+  ///
+  /// If [allowNewlines] is true, the indented syntax will consume newlines as
+  /// whitespace, in positions when a statement can not end.
   @protected
-  void whitespace() {
+  void whitespace({required bool allowNewlines}) {
```
nex3, line 81: I'd call this `consumeNewlines`, to avoid suggesting that it throws an error when a newline is encountered.

lib/src/parse/stylesheet.dart
```diff
@@ -1729,16 +1752,22 @@ abstract class StylesheetParser extends Parser {
   /// If [until] is passed, it's called each time the expression could end and
   /// still be a valid expression. When it returns `true`, this returns the
   /// expression.
+  ///
+  /// If [allowNewlines] is true, the indented syntax will consume newlines as
+  /// whitespace, in positions when a statement can not end.
```
nex3, line 1757: This should be more explicit that it only refers to the top level of the expression, since expressions can have bracketed sub-expressions that always allow newlines.

lib/src/parse/stylesheet.dart
```diff
@@ -1936,7 +1964,11 @@ abstract class StylesheetParser extends Parser {
 
     loop:
     while (true) {
-      whitespace();
+      whitespace(
+          allowNewlines: allowNewlines ||
+              bracketList ||
+              _inParentheses ||
+              wasInExpression);
```
nex3, line 1971: Why does this trigger on `wasInExpression`?

lib/src/parse/stylesheet.dart
```diff
@@ -2857,6 +2898,17 @@ abstract class StylesheetParser extends Parser {
             buffer.writeCharCode(scanner.readChar());
           }
 
+        case $lparen || $lbracket:
+          var bracket = scanner.readChar();
+          buffer.writeCharCode(bracket);
+          brackets.add(opposite(bracket));
+
+        case $rparen || $rbracket:
+          if (brackets.isEmpty) break loop;
```
nex3, line 2907: Given that we're now explicitly parsing matching brackets here, it's probably best to throw a meaningful error here.

Note that this is technically a breaking change whether or not we throw an error here: `[foo#{"]:is(bar"}) {x: y}` is currently valid, but will become an error here. I think it's vanishingly unlikely that anyone is doing that, though, so it should be fine to break, but we should add a "potentially breaking bug fix" note in the changelog about it.

lib/src/parse/stylesheet.dart
```diff
@@ -2943,7 +2999,7 @@ abstract class StylesheetParser extends Parser {
         case $space || $tab:
           buffer.writeCharCode(scanner.readChar());
 
-        case $lf || $cr || $ff when indented:
+        case $lf || $cr || $ff when indented && !allowNewlines:
```
nex3, line 3002: Shouldn't this also allow newlines when `brackets` isn't empty?

lib/src/parse/stylesheet.dart
```diff
@@ -3304,11 +3364,11 @@ abstract class StylesheetParser extends Parser {
         operator = "and";
       }
 
-      whitespace();
+      whitespace(allowNewlines: true);
       var right = _supportsConditionInParens();
       condition = SupportsOperation(
           condition, right, operator, scanner.spanFrom(start));
-      whitespace();
+      whitespace(allowNewlines: false);
```
nex3, line 3371: Should this be `: inParentheses`?

lib/src/parse/stylesheet.dart
```diff
@@ -3283,16 +3340,19 @@ abstract class StylesheetParser extends Parser {
...
   /// Consumes a `@supports` condition.
-  SupportsCondition _supportsCondition() {
+  ///
+  /// If [inParentheses] is true, the indented syntax will consume newlines
+  /// where a statement otherwise would end.
+  SupportsCondition _supportsCondition({bool inParentheses = false}) {
     var start = scanner.state;
     if (scanIdentifier("not")) {
-      whitespace();
+      whitespace(allowNewlines: true);
```
nex3, line 3349: I think we should _only_ allow newlines here when `inParentheses` is true, since we want to eventually treat `@supports` like any other at-rule. Same below.

nex3 (changes requested):
> @nex3 This is ready for review. Here are some questions I wanted to flag as you review. Thanks!
> 
> 1. `allowNewlines` is set to `true` in some cases, like in parser* functions or in `scss.dart`, where the value is non-impactful. Should this usage be differentiated somehow for clarity?

For classes where it's universally non-impactful, I'd define a local `_whitespace()` function that takes no arguments with a comment explaining that the value isn't relevant.

> 2. There are some places, like in `_atRootQuery`, where `allowNewlines` is true because it is inside parentheses. I opted to not use the `_inParentheses` variable except for one place where the consumption did rely on it, in `_supportsCondition`. Would it be more idiomatic to use `_inParentheses`, even if the consumption is in the same function?

I think how it is now is better.

> 3. In the `+` include syntax, I am not supporting whitespace after the `+` as it can be part of a selector. This differs from the `=` mixin syntax- should they be the same?

There's already an odd inconsistency here where whitespace is allowed after `=` but not after `+`, so I think it's fine to expand that to allowing newlines as well. It's possible that we should deprecate that, but it's not in-scope for this change.

> 4. I added paren/bracket context to `almostAnyValue`, with the only change being that they must be matched. That means the error for `a:has(d[)]` will be changed from "Expected identifer" to "Expected `]`". Since `almostAnyValue` is used for more than selectors, should I make this functionality opt-in, dependent on a parameter? Based on the places where this is used, I don't foresee issues with bracket matching, but want to verify.

It's fine to change this error. Arguably the bracket error is clearer anyway.

lib/src/parse/stylesheet.dart
nex3: If you pass `consumeNewlines: true` into `expressionUntilComma()` in `parentheses()`, I think that's a much more explicit way of representing the logic that parenthesized lists should allow newlines everywhere. That should allow you to remove the `_inParentheses` check as well.

lib/src/parse/stylesheet.dart
```diff
@@ -2857,6 +2899,20 @@ abstract class StylesheetParser extends Parser {
...
 
+        case $lparen || $lbracket:
+          var bracket = scanner.readChar();
+          buffer.writeCharCode(bracket);
+          brackets.add(opposite(bracket));
+
+        case $rparen || $rbracket:
+          if (brackets.isEmpty) {
+            scanner.error(
+                'Unexpected "${String.fromCharCode(scanner.peekChar()!)}".');
```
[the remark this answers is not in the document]
nex3: Yeah, you can write `case ($paren || $rbracket) && var char` and it'll bind it to `char`.

nex3 (changes requested):

lib/src/parse/sass.dart
```diff
@@ -309,37 +312,22 @@ class SassParser extends StylesheetParser {
     return LoudComment(buffer.interpolation(scanner.spanFrom(start)));
   }
 
-  void whitespaceWithoutComments() {
+  void whitespaceWithoutComments({required bool consumeNewlines}) {
     // This overrides whitespace consumption so that it doesn't consume
-    // newlines.
+    // newlines where that would cause a statement to end.
```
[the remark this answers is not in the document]
jamesnw: Updated in ea8b098a

lib/src/parse/parser.dart
```diff
@@ -67,23 +67,29 @@ class Parser {
...
+    whitespace(consumeNewlines: true);
     return scanner.scanChar($colon);
   }
 
   // ## Tokens
 
   /// Consumes whitespace, including any comments.
+  ///
+  /// If [consumeNewlines] is true, the indented syntax will consume newlines as
+  /// whitespace in positions when a statement can't end.
```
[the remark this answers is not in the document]
jamesnw: Updated in ea8b098a

jamesnw (commented):
@nex3 Changes made. I also have an outstanding question here- https://github.com/sass/dart-sass/pull/2467/files/38e05bf45a34f175921f5be1d8efbf4ab2c54a9a#r1896194128

lib/src/parse/stylesheet.dart
nex3: Ah, I must have misread.

CHANGELOG.md
```diff
@@ -5,6 +5,24 @@
 * **Potentially breaking bug fix**: Selectors with unmatched brackets now always
   produce a parser error. Previously, some edge cases like `[foo#{"]:is(bar"}) {a:
   b}` would compile without error, but this was an unintentional bug.
+## 1.83.5-dev
```
nex3, line 8: Get rid of this header.

nex3 (approved):

CHANGELOG.md
jamesnw: @nex3 I removed that. I also removed an extra period at the end of the next point that I noticed. (see https://github.com/sass/dart-sass/pull/2467/commits/7c6d741430ffcd60a0b6cbcd61acf6ff842b41e9)

nex3:
Holding off on merging this until https://github.com/sass/dart-sass/pull/2505 lands so it can get into the release.

PR #2468, opened 2024-12-19
Add support for declarations

nex3:


pkg/sass-parser/lib/src/statement/declaration.test.ts
```diff
@@ -0,0 +1,511 @@
...
+        ).toBe('foo: bar !important'));
+
+      it('ignores value', () =>
+        expect(
+          new Declaration({
+            prop: 'foo',
+            value: 'bar',
+            raws: {value: {value: 'bar', raw: 'BAR'}},
+          }).toString(),
+        ).toBe('foo: bar'));
```
Goodwine, line 309: I'm slightly lost, why is `!important` and `value` ignored?

Goodwine (changes requested):

pkg/sass-parser/lib/src/statement/declaration.test.ts
nex3: Because in the Sass AST, those are fully parsed expressions in their own right, and the expression ASTs determine how they're serialized rather than the statement-level raws.

Goodwine (approved):

PR #2478, opened 2025-01-08
Add support for `@if` and `@else`

nex3:


pkg/sass-parser/lib/src/statement/else-rule.test.ts
```diff
@@ -0,0 +1,300 @@
...
+    function describeNode(description: string, create: () => ElseRule): void {
+      describe(description, () => {
+        beforeEach(() => void (node = create()));
+
+        it('has a name', () => expect(node.name.toString()).toBe('else'));
+
+        it('has no expression', () =>
+          expect(node.elseCondition).toBeUndefined());
+
+        it('has matching params', () => expect(node.params).toBe(''));
```
Goodwine, line 20: Nit: should this be "has no params" or "has empty params"?

pkg/sass-parser/lib/src/statement/index.ts
```diff
@@ -195,6 +206,15 @@ const visitor = sassInternal.createStatementVisitor<Statement>({
   visitForRule: inner => new ForRule(undefined, inner),
   visitForwardRule: inner => new ForwardRule(undefined, inner),
   visitFunctionRule: inner => new FunctionRule(undefined, inner),
+  visitIfRule: inner => {
+    const rules: Statement[] = [new IfRule(undefined, inner)];
+    for (let i = 1; i < inner.clauses.length; i++) {
```
Goodwine, line 211: I'm slightly confused why the zero index gets skipped, I imagine it's used in `IfRule`, but I don't see `.clauses[0]`.

I think a comment here can be helpful to understand why it goes from 1 to N rather than 0 to N.

Goodwine (changes requested):

Goodwine (approved):

PR #2488, opened 2025-01-13
Always use the SDK version of the formatter

nex3:
This helps guard against skew when new formatting logic is released in
the dart_style package before it's available in a stable SDK release.


tool/grind/generate_deprecations.dart
```diff
@@ -6,7 +6,6 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:crypto/crypto.dart';
-import 'package:dart_style/dart_style.dart';
 import 'package:grinder/grinder.dart';
 import 'package:pub_semver/pub_semver.dart';
```
jathak, line 10: It looks like this import is unnecessary (also in tool/grind/synchronize.dart)

jathak (approved):

PR #2501, opened 2025-01-30
Add support for the `@content` rule

nex3:


pkg/sass-parser/lib/src/statement/content-rule.ts
```diff
@@ -0,0 +1,128 @@
...
+import * as sassParser from '../..';
+
+/**
+ * The set of raws supported by {@link ContentRule}.
+ *
+ * @category Statement
+ */
+export interface ContentRuleRaws extends Omit<AtRuleRaws, 'params'> {
+  /**
+   * Whether to content an empty argument list. If the argument list isn't
```
Goodwine, line 25:
```suggestion
   * Whether to include an empty argument list. If the argument list isn't
```
I was confused by the comment, but IIUC this was unintentionally search-replaced

Goodwine (approved):

pkg/sass-parser/lib/src/statement/content-rule.ts
nex3: That is accurate :smile:

PR #2503, opened 2025-01-30
Add support for color literals

nex3:


pkg/sass-parser/lib/src/expression/color.test.ts
```diff
@@ -0,0 +1,249 @@
...
+        new SassColor({
+          space: 'rgb',
+          red: 10,
+          green: 20,
+          blue: 30,
+          alpha: 0.4,
+        }),
+      );
+    });
+  });
```
Goodwine, line 157: nit: merge `describe` into `it` since it's only one test

pkg/sass-parser/lib/src/expression/__snapshots__/color.test.ts.snap
```diff
@@ -0,0 +1,25 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`a color expression toJSON 1`] = `
+{
+  "inputs": [
+    {
+      "css": "@#{#00f}",
```
Goodwine, line 7: Is it intended for `@` to be here?

pkg/sass-parser/lib/src/expression/color.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+export class ColorExpression extends Expression {
+  readonly sassType = 'color' as const;
+  declare raws: ColorExpressionRaws;
+
+  /**
+   * The color represented by this expression. This will always be a color in
+   * the RGB color space.
+   *
+   * THrows an error if this is set to a non-RGB color.
```
Goodwine, line 58:
```suggestion
   * Throws an error if this is set to a non-RGB color.
```

pkg/sass-parser/lib/src/expression/color.ts
```diff
@@ -0,0 +1,121 @@
...
+ * An expression representing a color literal in Sass.
+ *
+ * @category Expression
+ */
+export class ColorExpression extends Expression {
+  readonly sassType = 'color' as const;
+  declare raws: ColorExpressionRaws;
+
+  /**
+   * The color represented by this expression. This will always be a color in
```
Goodwine, line 55: Why will it always be an RGB color? It doesn't seem like it's a limitation on sass_api 🤔

Is this a temporary limitation?

Goodwine (changes requested):

pkg/sass-parser/lib/src/expression/__snapshots__/color.test.ts.snap
nex3: Yeah, because we didn't have variable statements early on the way we parse expressions is by creating an interpolation in an unknown at-rule.

pkg/sass-parser/lib/src/expression/color.ts
nex3: Because Sass has no literal syntactic representation of non-RGB colors. Every other color type is returned by function calls.

Goodwine (approved):

PR #2509, opened 2025-02-11
Clean some comments in the code

stof:
The comments in environments is not relevant since the refactoring of `ConfiguredValue` introducing the `implicit` factory. `null` is not passed to the factory (the factory does that behavior internally for all implicit values instead)


nex3 (approved):

nex3:
Oops, didn't notice the tests weren't passing. Don't forget to automatically re-run the synchronization script!

PR #2513, opened 2025-02-12
Fix dart analyze and format on 3.7.0

ntkme:
This PR fixes the following:

```
warning • lib/src/deprecation.dart:185:43 • A value for optional parameter 'description' isn't ever given. Try removing the unused parameter. • unused_element_parameter
   info • lib/sass.dart:6:9 • Library names are not necessary. Remove the library name. • unnecessary_library_name
   info • pkg/sass_api/lib/sass_api.dart:6:9 • Library names are not necessary. Remove the library name. • unnecessary_library_name
```

There are a few `deprecated_member_use` on `dart:js_util`, but migrate to `dart:js` is not a small task so that I'm not touching those for now.

---

For `dart format` this PR adds `dart pub get` before running `dart format` in CI to make sure the language version is correctly detected and a consistent format is produced.


ntkme:
@nex3 I just found out that with dart 3.7, if you start with a clean clone of the repo (or run `git clean -dffx`), it formats in one way. Then, after running `dart pub get`, it formats in another way...

ntkme:
It's mentioned in the changelog: https://github.com/dart-lang/sdk/blob/main/CHANGELOG.md#dart-format

> In order to correctly determine the language version of each file it formats, dart format (like other dart commands) looks for a package_config.json file surrounding the files being formatted. This means that you need to run dart pub get before formatting code in your package. If you have format checks in your continuous integration server, you'll want to make sure it runs dart pub get too.

tool/grind.dart
```diff
@@ -141,16 +150,24 @@ void all() {}
 
 @Task('Run the Dart formatter.')
 void format() {
-  run('dart', arguments: ['format', '--fix', '.']);
+  run('dart', arguments: ['format', '.']);
+  run('dart', arguments: ['fix', '--apply', '.']);
```
ntkme, line 154: I think it's better to apply the fix first and then format?

```suggestion
  run('dart', arguments: ['fix', '--apply', '.']);
  run('dart', arguments: ['format', '.']);
```

nex3: There are arguments both ways. Fixing first will properly handle lines that are made longer by fixed, while formatting first will correctly handle things like single-line `if` statements that are pushed onto multiple lines by the formatter. Maybe we should just do fix-format-fix, since the second fix shouldn't ever cause a need for a second format, although that's going to be wasted work most of the time.

PR #2519, opened 2025-02-20
Add support for function call expressions

nex3:


pkg/sass-parser/lib/src/expression/function.test.ts
```diff
@@ -0,0 +1,316 @@
+// Copyright 2024 Google Inc. Use of this source code is governed by an
```
Goodwine, line 1:
```suggestion
// Copyright 2025 Google Inc. Use of this source code is governed by an
```

pkg/sass-parser/lib/src/expression/function.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      utils.fromExpressionProps({
+        namespace: 'baz',
+        name: 'foo',
+        arguments: [{text: 'bar'}],
+      }),
+    );
+  });
+
+  describe('if()', () => {
+    beforeEach(() => void (node = utils.parseExpression('if(cond, bar, baz)')));
```
Goodwine, line 91: TIL `if()` is a function in Sass! 🤯

Goodwine (approved):

pkg/sass-parser/lib/src/expression/function.test.ts
nex3: Only sort of! It's actually technically its own syntactic construct, distinct from a standard function in that it only ever evaluates one of the `true` or `false` branches. Honestly at this point we could just make it check that at runtime like we do for `calc()` _et al_, but we haven't made that change yet.

PR #2535, opened 2025-03-06
Add support for the supports() function

nex3:


pkg/sass-parser/lib/src/expression/string.test.ts
```diff
@@ -284,5 +290,24 @@ describe('a string expression', () => {
...
+      expect(utils.parseExpression('f#{o}o')).toMatchSnapshot());
+
+    it('supports()', () =>
+      expect(
+        (
+          (
+            scss.parse('@import "foo" supports(width: 1 + 1)')
+              .nodes[0] as ImportRule
+          ).imports.nodes[0] as StaticImport
+        ).modifiers!.nodes[1],
```
Goodwine, line 310: I'm confused here. What is the value of `nodes[0]`, is it `supports(`?
And if so, is there a `nodes[2]` with the value of `)`?

nex3: The first `nodes[0]` is the whole `@import` statement. The second one is `"foo" supports(width: 1 + 1)` (because `@import` in Sass can have multiple comma-separated clauses). Then `nodes[1]` is actually `supports(width: 1 + 1)`.

Goodwine (approved):

PR #2536, opened 2025-03-07
Drop support for bogus combinators

nex3:
This also improves the way SassExceptions thrown from within custom
functions are handled. If the wrapped exception already has a span,
it's turned into a MultiSpanSassException with an additional span
pointing to the function invocation.

Closes https://github.com/sass/dart-sass/issues/1727

[see sass/sass#4040]
[see sass/sass-spec#2049]

### DIFF
--- a/lib/src/deprecation.dart
+++ b/lib/src/deprecation.dart
@@ -15,7 +15,7 @@ enum Deprecation {
   // DO NOT EDIT. This section was generated from the language repo.
   // See tool/grind/generate_deprecations.dart for details.
   //
-  // Checksum: 8f593fa3dcf8de884b25a82e4625a8bc9effef50
+  // Checksum: b715850a96d4d346e09fecb54444ccec6dc69610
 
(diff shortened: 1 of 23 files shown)


lib/src/deprecation.dart
```diff
@@ -15,7 +15,7 @@ enum Deprecation {
   // DO NOT EDIT. This section was generated from the language repo.
   // See tool/grind/generate_deprecations.dart for details.
   //
-  // Checksum: 47c97f7824eb25d7f1e64e3230938b88330d40b4
+  // Checksum: b715850a96d4d346e09fecb54444ccec6dc69610
```
jathak, line 18: Not sure what's going on here. The checksum is correct for the YAML file the marks `bogus-combinators` as obsolete, but the actual code is only updated `mozDocument`, not `bogusCombinators`

jathak: Oh wait, now I figured it out. The `obsolete` value was added to the YAML, but the status wasn't changed from `active`

PR #2538, opened 2025-03-07
Add support for parsing unary operations

nex3:


pkg/sass-parser/lib/src/expression/unary-operation.test.ts
```diff
@@ -0,0 +1,268 @@
...
+      it("removes the old operand's parent", () => {
+        const oldOperand = node.operand;
+        node.operand = {text: 'zip'};
+        expect(oldOperand.parent).toBeUndefined();
+      });
+
+      it('assigns operand explicitly', () => {
+        const operand = new StringExpression({text: 'zip'});
+        node.operand = operand;
+        expect(node.operand).toBe(operand);
```
Goodwine, line 63: should this check for `operand.parent` to be `node`?

Goodwine (approved):

Issue #2539, opened 2025-03-07
Support new attr() function syntax


nex3:
See https://github.com/sass/sass/issues/4030

- [ ] Deprecation
- [ ] Final changes

PR #2540, opened 2025-03-08
Add support for `%` as a stand-alone expression

nex3:
This also deprecates user-defined functions named "type".

See #2539

[see sass/sass#4045]
[see sass/sass-spec#2051]


lib/src/deprecation.dart
```diff
@@ -166,6 +166,13 @@ enum Deprecation {
         'Global built-in functions that are available in sass: modules.',
   ),
 
+  /// Deprecation for functions named "type".
+  typeFunction(
+    'type-function',
+    deprecatedIn: '1.86.0',
+    description: 'Functions named "type".',
+  ),
+
```
Goodwine, line 175: off-topic: for whatever reason when I generate the deprecations I end up getting a different formatting, like:

```dart
  typeFunction('type-function', deprecatedIn: '1.86.0',
    description: 'Functions named "type".'),
```

do you happen to know what that would be? this is when running `dart run grinder`... I guess I may have some config file somewhere that tells dart to behave this way 🤔

Goodwine (approved):

lib/src/deprecation.dart
nex3: I saw something similar and I couldn't figure out exactly what it was. The Dart formatter is in the process of rolling out new changes so it's probably something related to that, but I don't know what exactly. I suspect making the generator add a trailing comma would force it to behave the same way for both the old and new formatter logic, though.

lib/src/parse/stylesheet.dart
```diff
@@ -2089,9 +2097,16 @@ abstract class StylesheetParser extends Parser {
           length: operator.operator.length,
         );
       }
-      operands.add(singleExpression);
       whitespace(consumeNewlines: true);
-      singleExpression_ = _singleExpression();
+
+      if (operator == BinaryOperator.modulo && !_lookingAtExpression()) {
+        addSingleExpression(StringExpression.plain(
+            '%', scanner.spanFromPosition(scanner.position - 1)));
```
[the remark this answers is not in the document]
nex3: Yeah, that's true. Want to send a fix?

Issue #2549, opened 2025-03-22
Provide CHECKSUMS file on releases


otaxhu:
I was trying to download Sass in Releases page, but then realized that it does not provide a CHECKSUMS file, I think it would be nice to add it and have another layer of security for this project.

Related PR #2550

## What this issue solves:
- Not require the user to login with its Github account just to check the released files integrities (by using the Github Attestation API). Also by doing it, we are decentralizing the distribution of the software, making it more available to everyone.

ntkme:
Checksum file only helps checking if download is corrupted or not, and if the archive is corrupted the unarchive step usually just straight out fail. So having checksum does not really improve anything in this case.

From a security perspective if you are worried about supply chain attacks where the whole archive is replaced, keep in mind that the checksum is also mutable if it’s uploaded to release artifacts, so that also won’t prevent such attack unless you upload the checksum in an immutable storage.

I suggest take a look at artifact attestation instead: https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds

ntkme:
> For downloading any dependency in Meson, it requires the checksum to be specified.

In this case, it's more of making sure the external dependency is immutable for security purpose. You can just compute the checksum of the downloaded file and checkin it the checksum yourself, just like how package manager's lock file works. 

> Artifact attestation does generate a checksum for every file?, if so, how do I download it?

It signs the checksum for every file and upload the signature in an immutable store, which you can download and verify with github cli or api. There is no precomputed checksum file that you can download and copy-paste, but this is much more secure as you don't have to worry about the authenticity of the checksum file.

otaxhu:
And how would I know that the file is not corrupted if I don't have a reference to compare it to?

otaxhu:
But meson doesn't have that kind of integration with Github.

It's still nice to have a copy-paste version of the hashes. I opened #2550 PR

otaxhu:
This also decentralizes the distribution of Sass, allowing different distributors to distribute this awesome software, instead of spamming the Github API, they will just need to download the CHECKSUMS file once.

ntkme:
> It's still nice to have a copy-paste version of the hashes.

There are two properties here:

- Immutability
- Authenticity

There are different levels of security:

1. A checksum stored together with the release artifact have neither of the properties.
[...]

PR #2550, opened 2025-03-22
Generating CHECKSUMS file on release.yml workflow

otaxhu:
Solves #2549


nex3:
See discussion in #2549.

PR #2552, opened 2025-03-28
Improve case canonicalization performance

### LINKED ISSUE #2548: Sass hangs when a file is imported from a location with a large file tree
### Summary

I have a scss file that contains an `@import` or `@use` statement that references a file in another location. That other location is provided using the `--load-path` CLI option. When I try to run sass on that single file, it hangs for what seems like an infinite amount of time (more than 10 minutes).

### Details

- Operating System: Windows
- Runtime environment: NodeJS

### Root cause

I dug in your source code and understood that the root cause is the algorithm that walks up the source file location and caches the directories names with their "real" case ([io.dart#_realCasePath](https://github.com/sass/dart-sass/blob/main/lib/src/io.dart#L36)). In my case, while walking up, it falls into a directory with more than 10000 subfolders and gets stuck there because the current implementation of [io/js.dart#dirExists](https://github.com/sass/dart-sass/blob/main/lib/src/io/js.dart#L163) based on sequential synchronous calls to `fs.existsSync` and `fs.statsSync` is extremely costly (in my case, I measured ~1min for 1000 folders).

Based on the following comment in the code, you already know that the current implementation performs poorly:
> TODO(nweiz): Use an SDK function for this when dart-lang/sdk#35370 and/or nodejs/node#24942 are fixed, or at least use FFI functions.

I looked at the mentioned issues and it seems that none of them get a lot of attention and they may never get implemented (especially on NodeJS side).

### Proposal

Based on the discussions in the aforementioned issues, it seems important to avoid symbolic links resolution but I don't fully grasp why and I'm no expert on those, so the proposal below may not be valid...

I think you could leverage the NodeJS API `fs.realpath.native` to check whether or not the path contains a symbolic link, and if doesn't, you don't have to explore the entire file tree to get the proper case since `realpath.native` already returns it. If there is a symbolic link, then you can fall back to your existing mechanism.

I don't know Dart but here is what it could look like in JS:

```js
const realPath = fs.realpathSync.native(path)
if (realPath.toUpperCase() === path.toUpperCase()) {
  // No symbolic link
  let reconstructed
  realPath.split(path.sep).forEach((part) => {
    reconstructed = reconstructed ? join(reconstructed, part) : part
    _realCasePath.putIfAbsent(reconstructed)
  })
  return realPath
}
return helper(path)
```

It's not perfect because it does not address the case of paths with symbolic links but it would noticeably speed up the path computation and caching for all other cases (especially if you get rid of the "walk up" mechanism as you already know the entire path has the correct case).

nex3:
Closes #2548

### DIFF
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -36,7 +36,7 @@ String _realCasePath(String path) {
   // TODO(nweiz): Use an SDK function for this when dart-lang/sdk#35370 and/or
   // nodejs/node#24942 are fixed, or at least use FFI functions.
 
-  //if (!_couldBeCaseInsensitive) return path;
+  if (!_couldBeCaseInsensitive) return path;
 
   if (isWindows) {
     // Drive names are *always* case-insensitive, so convert them to uppercase.
--- a/lib/src/io.dart
+++ b/lib/src/io.dart
@@ -30,14 +30,13 @@ String canonicalize(String path) => _couldBeCaseInsensitive
 
 /// Returns `path` with the case updated to match the path's case on disk.
 ///
-/// This only updates `path`'s basename. It always returns `path` as-is on
-/// operating systems other than Windows or Mac OS, since they almost never use
-/// case-insensitive filesystems.
+/// This always returns `path` as-is on operating systems other than Windows or
+/// Mac OS, since they almost never use case-insensitive filesystems.
 String _realCasePath(String path) {
   // TODO(nweiz): Use an SDK function for this when dart-lang/sdk#35370 and/or
   // nodejs/node#24942 are fixed, or at least use FFI functions.
 
-  if (!_couldBeCaseInsensitive) return path;
+  //if (!_couldBeCaseInsensitive) return path;
 
   if (isWindows) {
     // Drive names are *always* case-insensitive, so convert them to uppercase.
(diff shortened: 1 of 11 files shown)


lib/src/io/js.dart
```diff
@@ -181,6 +181,21 @@ bool dirExists(String path) {
   });
 }
 
+bool linkExists(String path) {
+  if (!isNodeJs) {
+    throw UnsupportedError("linkExists() is only supported on Node.js");
+  }
+  return _systemErrorToFileSystemException(() {
+    try {
+      return fs.statSync(path).isSymbolicLink();
```
ntkme, line 190:
```suggestion
      return fs.lstatSync(path).isSymbolicLink();
```

`stat` checks resolved symlink thus it always return `false` for `isSymbolicLink`. I think you want `lstat` which checks the symlink itself.

ntkme (changes requested):

lib/src/io.dart
```diff
@@ -30,14 +30,13 @@ String canonicalize(String path) => _couldBeCaseInsensitive
...
-/// operating systems other than Windows or Mac OS, since they almost never use
-/// case-insensitive filesystems.
+/// This always returns `path` as-is on operating systems other than Windows or
+/// Mac OS, since they almost never use case-insensitive filesystems.
 String _realCasePath(String path) {
   // TODO(nweiz): Use an SDK function for this when dart-lang/sdk#35370 and/or
   // nodejs/node#24942 are fixed, or at least use FFI functions.
 
-  if (!_couldBeCaseInsensitive) return path;
+  //if (!_couldBeCaseInsensitive) return path;
```
jathak, line 39: Is this supposed to be removed instead of commented out?

jathak (approved):

lib/src/io.dart
nex3: No, this was supposed to just be a local test! Good catch.

PR #2557, opened 2025-04-01
Fix the sass_api release

nex3:
Pub.dev just launched an update that requires GitHub publish actions
to be run from a tag that matches the version number of the published
package.

### DIFF
--- a/.github/workflows/release-sass-api.yml
+++ b/.github/workflows/release-sass-api.yml
@@ -1,0 +1,29 @@
+# This has to be a separate workflow to satisfy pub.dev's somewhat draconian
+# requirements for when and how GitHub actions are allowed publish packages.
+# Specifically, it requires that each publish action be triggered by a tag that
+# contains the version number for that action, so we can't just publish sass_api
+# as part of the standard sass release flow because that was triggered by a tag
+# with the sass version number.
+name: Release sass-api
+
+on:
+  push:
+    tags: ['sass-api-[0-9]+.[0-9]+.*']
+
+jobs:
+  deploy_sass_api:
+    if: github.event.repository.fork == false
+    name: Deploy sass_api
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/util/initialize
+        with: {github-token: "${{ github.token }}"}
+
+      - name: Deploy
+        run: dart run grinder deploy-sass-api
+        env:
+          PUB_CREDENTIALS: "${{ secrets.PUB_CREDENTIALS }}"
+          GH_TOKEN: "${{ secrets.GH_TOKEN }}"
+          GH_USER: sassbot
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -95,6 +95,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+        with:
+          # We have to use this rather than the implicit GitHub token so that
+          # pushing a new tag triggers another action.
+          token: ${{ secrets.GH_TOKEN }}
       - uses: ./.github/util/initialize
         with: {github-token: "${{ github.token }}"}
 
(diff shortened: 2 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -102,22 +102,14 @@ jobs:
...
-          GH_TOKEN: "${{ secrets.GH_TOKEN }}"
-          GH_USER: sassbot
+      - name: Get Sass API version
+        id: sass-api-version
+        run: |
+          echo "version=$(cat pkg/sass_api/pubspec.yaml | sed -nE 's/version: (.*)/\1/p')" | tee --append "$GITHUB_OUTPUT"
+      # This should be /-separated rather than hyphenated, but pub.dev doesn't
+      # currently allow that (dart-lang/pub-dev#8690).
+      - run: git tag sass-api-${{ steps.sass-api-version.outputs.version }}
+      - run: git push --tag
```
ntkme, line 112: If you want to push with the implicit `${{ github.token }}` that comes from github actions, you need to add permissions to the job definition:
```
permissions:
  content: write
```

ntkme: Also, another problem I remember is that pushing with implicit token won't trigger another workflow. So you probably should use `${{ secrets.GH_TOKEN }}` in `checkout` action, that it will trigger the workflow for new tag push.

nex3: Isn't that inherited from the invocation in `ci.yml`? `deploy_sass_parser` is pushing a tag without explicit local permissions below.

nex3: > Also, another problem I remember is that pushing with implicit token won't trigger another workflow. So you probably should use `${{ secrets.GH_TOKEN }}` in `checkout` action, that it will trigger the workflow for new tag push.

Done.

.github/workflows/release-sass-api.yml
```diff
@@ -0,0 +1,33 @@
...
+# with the sass version number.
+name: Release sass-api
+
+on:
+  push:
+    tags: ['sass-api-[0-9]+.[0-9]+.*']
+
+jobs:
+  deploy_sass_api:
+    if: "github.event.repository.fork == false"
```
Goodwine, line 15: Are these quotes necessary? IIRC quotes are only needed when the text contains a `*`

same below

Goodwine (approved):

.github/workflows/release.yml
```diff
@@ -95,29 +95,25 @@ jobs:
...
-          PUB_CREDENTIALS: "${{ secrets.PUB_CREDENTIALS }}"
-          GH_TOKEN: "${{ secrets.GH_TOKEN }}"
-          GH_USER: sassbot
+      - name: Get Sass API version
+        id: sass-api-version
+        run: |
+          echo "version=$(cat pkg/sass_api/pubspec.yaml | sed -nE 's/version: (.*)/\1/p')" | tee --append "$GITHUB_OUTPUT"
+      # This should be /-separated rather than hyphenated, but pub.dev doesn't
+      # currently allow that (dart-lang/pub-dev#8690).
+      - run: git tag sass-api-${{ steps.sass-api-version.outputs.version }}
```
ntkme, line 115: Not entirely sure, but another way this might be broken is that `git` might complain that "user" is not setup. Maybe we need to add these before creating the tag:
```
git config user.email 
git config user.name 
```

.github/workflows/release-sass-api.yml
nex3: There are a number of active YAML characters. You're right that this doesn't need them, but I prefer to use them for anything with `${{ }}` interpolation just to avoid needing readers to remember when `{` is or isn't an active YAML character.

.github/workflows/release.yml
nex3: I believe lightweight tags don't include any author information. (Note that the git user isn't set when tagging `sass-parser` below.)

.github/workflows/release.yml
ntkme: > Isn't that inherited

Permission inheritance in reusable workflow is a bit complicated. The permission declared in parent workflow defines the maximum permission child workflows can ask for, but if a child workflow did not explicitly ask for the permission, it won't get inherit from parent automatically.

https://docs.github.com/en/actions/sharing-automations/reusing-workflows#access-and-permissions

Anyways, better to use a PAT so that it guarantee the push event triggers a new workflow.

.github/workflows/release-sass-api.yml
```diff
@@ -0,0 +1,33 @@
...
+  push:
+    tags: ['sass-api-[0-9]+.[0-9]+.*']
+
+jobs:
+  deploy_sass_api:
+    if: github.event.repository.fork == false
+    permissions:
+      attestations: write
+      contents: write
+      id-token: write
```
ntkme, line 19: `contents` allows create commit/tag/release, `id-token` and `attestations` are needed for artifact attestations. I don't think you need any of these here.

Goodwine (approved):

.github/workflows/release.yml
nex3: Well, it's now using the same setup as `deploy_sass_parser` including the token it uses for checkout so it _should_ work.

PR #2563, opened 2025-04-15
Switch to OSS windows-11-arm runner

ntkme:
https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/


nex3:
Thanks!

PR #2570, opened 2025-05-03
Remove useless `_currentCallable` property in the evaluate visitor

stof:
Since https://github.com/sass/dart-sass/pull/2474, the `_currentCallable` property is not used anymore (except for restoring its state)

### DIFF
--- a/lib/src/deprecation.dart
+++ b/lib/src/deprecation.dart
@@ -18,167 +18,118 @@ enum Deprecation {
   // Checksum: c57ab2eb07ab1df48581b8484ef9bdbad0ddceaa
 
   /// Deprecation for passing a string directly to meta.call().
-  callString(
-    'call-string',
-    deprecatedIn: '0.0.0',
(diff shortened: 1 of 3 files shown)


lib/src/deprecation.dart
```diff
@@ -18,167 +18,118 @@ enum Deprecation {
   // Checksum: c57ab2eb07ab1df48581b8484ef9bdbad0ddceaa
 
   /// Deprecation for passing a string directly to meta.call().
-  callString(
-    'call-string',
-    deprecatedIn: '0.0.0',
-    description: 'Passing a string directly to meta.call().',
-  ),
+  callString('call-string',
```
stof, line 21: changes in this file are the result of running `dart run grinder`

Goodwine: ok so it wasn't just me then!

Goodwine (approved):

PR #2600, opened 2025-07-03
Fix a bug where a double-configuration error wouldn't be emitted

### LINKED ISSUE #2598: [Bug] This module was already loaded false positive
There appears to be an error-reporting bug in Sass 1.89.2 (Sass embedded on WSL-Linux, Dart Sass on Windows)

This bug is nearly identical to Issue #1180 but that one appears to have been resolved.

Please refer to -- or download -- the files in my repo: https://github.com/arikorn/sass-bug-demo

(Note: The example files are taken from CoreUI, so I would not normally have control over the file contents/loading sequence.
I have stripped out about as much as I could to keep it simple: This fileset represents the minimum number of files I needed to recreate the issue.)

The top-level file is ***App.scss***. It loads:
- scss/coreui.scss which loads
    - _variables.scss  // *also loaded in functions/_contrast-ratio.scss*
        - functions/color-contrast-variables
    - _functions.scss
        - "functions/_color-contrast"
            - functions/_contrast-ratio"
            - ../_variables
        - "functions/_color-contrast-variables" // *also loaded in _variables.scss*
        - "functions/_contrast-ratio"; // *also loaded in color-contrast.scss*

(Hopefully that's it! Notice that there are at two or three "loops" in the loading sequence.)

There is an (intentional) typo in the variable name, so it is not found in the imported library. (The correct name is $white.)
This transpiles correctly without the typo, but with it:

Run: `sass App.scss`

Sass reports:
~~~
Error: This module was already loaded, so it can't be configured using "with".
   ┌──> scss/_functions.scss
3  │   @forward "functions/contrast-ratio"; // also loaded in color-contrast
   │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ new load
   ╵
   ┌──> scss/functions/_color-contrast-variables.scss
1  │   @use "contrast-ratio" as *;
   │   ━━━━━━━━━━━━━━━━━━━━━━━━━━ original load
   ╵
   ┌──> App.scss
8  │ ┌ @use 'scss/coreui' with (
9  │ │    $whte: #fff
10 │ │ );
   │ └─^ configuration
   ╵
  scss/_functions.scss 3:1  @forward
  scss/coreui.scss 2:1      @use
  App.scss 8:1              root stylesheet
~~~

instead of:
~~~
Error: This variable was not declared with !default in the @used module.
  ╷
9 │    $whte: #fff
  │    ^^^^^^^^^^^
  ╵
  App.scss 9:4  root stylesheet
~~~

nex3:
Closes #2598

See https://github.com/sass/sass-spec/pull/2061

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.90.0
+
+* **Potentially breaking bug fix:** Fix a situation where Sass wasn't emitting
+  an error when loading a `@forward`ed module with a configuration when that
+  module had already been loaded with a different configuration *and* all
+  configured variables in the new configuration had already been used.
+
 ## 1.89.2
 
 ### Embedded Host
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -50,7 +50,7 @@ final class Configuration {
   /// will be considered to have the same original config if they were created
   /// as a copy from the same base configuration.
   bool sameOriginal(Configuration that) =>
-      _originalConfiguration == that._originalConfiguration;
+      identical(_originalConfiguration, that._originalConfiguration);
 
   /// The empty configuration, which indicates that the module has not been
   /// configured.
@@ -70,7 +70,7 @@ final class Configuration {
 
   /// Creates a new configuration from this one based on a `@forward` rule.
   Configuration throughForward(ForwardRule forward) {
-    if (isEmpty) return const Configuration.empty();
+    if (isEmpty) return this;
     var newValues = _values;
 
     // Only allow variables that are visible through the `@forward` to be
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -30,8 +30,7 @@ abstract interface class Module<T extends AsyncCallable> {
   /// [AstNode.span] if the span isn't required, since some nodes need to do
   /// real work to manufacture a source span.
   ///
-  /// Implementations must ensure that this has the same keys as [variables] if
-  /// it's not `null`.
+  /// Implementations must ensure that this has the same keys as [variables].
   Map<String, AstNode> get variableNodes;
 
   /// The module's functions.
--- a/lib/src/visitor/async_evaluate.dart
+++ b/lib/src/visitor/async_evaluate.dart
@@ -854,8 +854,8 @@ final class _EvaluateVisitor
   }) async {
     var url = stylesheet.span.sourceUrl;
 
+    var currentConfiguration = configuration ?? _configuration;
     if (_modules[url] case var alreadyLoaded?) {
-      var currentConfiguration = configuration ?? _configuration;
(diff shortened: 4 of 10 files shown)


arikorn:
I am concerned that this PR would break existing codebases (see @nex3's justification for this PR in #2598). It is possible that my PR #2602 would mitigate these effects.

jathak (approved):

arikorn:
Thank you for the update @nex3: this is great! It appears to resolve the examples we've discussed and to accomplish what I intended in PR #2602.

There is a minor issue if the variable is in an `@use` file instead of `@forward` . For example:
~~~sass
// App.scss (top-level Sass file)

@use 'scss/coreui2' with (
     $white: #aaa
) ;
~~~

~~~sass
// scss/coreui2.scss
@use "variables";
@forward "functions";

~~~

with the remaining files are as described in Issue #2598, and in https://github.com/arikorn/sass-bug-demo

The result is:
~~~
Error: This variable was not declared with !default in the @used module.
  ╷
9 │       $white: #aaa
  │       ^^^^^^^^^^^^
  ╵
  ..\dart-tests\App2.scss 9:7  root stylesheet

~~~

The error message is wrong ($white is declared with `!default` and it's in *_variables.css*) and could be probably be avoided if `@use` was treated similarly to `@forward`, i.e. don't cause an error if there are no remaining config variables.

Just to be clear: the only change here from the previous examples is that *coreui2.scss* loads *variables.scss* with `@use` instead of `@forward`

Note, though, that this is not a breaking change, since the same input causes an error when using the main branch as well.

I have other specific comments that I'll attach to the code review, but nothing bigger than this.

arikorn:
Oh drat!  It also fails the test you mention [here](https://github.com/sass/dart-sass/issues/2601#issuecomment-3046528111), which can be rephrased as:
~~~sass
 // App.scss (top-level Sass file)

@use 'scss/coreui2' as _;
@use 'scss/coreui2' with (
     $white: #aaa
) ;
~~~
(Giving a "variable not declared with !default" error instead of "module was already loaded" error.)

Perhaps the answer is that this is a work-in-progress, and those unused functions will fix it?

nex3:
> The error message is wrong ($white is declared with `!default` and it's in __variables.css_) and could be probably be avoided if `@use` was treated similarly to `@forward`, i.e. don't cause an error if there are no remaining config variables.

This error message is correct. `$white` is not declared with `!default` _in `scss/coreui2`_, which is the module that is being `@use`d in the context of that error. Only `@forward` adds additional variables to a module; an inner `@use` is a black box as far as downstream consumers are concerned.

> (Giving a "variable not declared with !default" error instead of "module was already loaded" error.)

Good catch. Looks like we're not handling "could have been configured" logic correctly for forwards.

jathak (approved):

PR #2612, opened 2025-07-29
chore: Add VS Code's config folder to .gitignore

arikorn:
The title says it all. Allows VS Code users to manage their configs w/o interfering with the repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ dist/
 
 # Generated protocol buffer files.
 *.pb*.dart
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ dist/
 
 # Generated protocol buffer files.
 *.pb*.dart
-.vscode
+.vscode/


ntkme:
Files created by IDE are more suitable to go into a global gitignore file, rather than per repository gitignore files.

See https://docs.github.com/en/get-started/git-basics/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

arikorn:
This is good advice @ntkme and yet there are reasons to include it in this repo (esp. the third point here)
1. It's such a small effort to provide this little courtesy.
2. In fact if you look at the most popular code repositories on Github, they exclude .vscode either explicitly or implicitly in *.gitignore* (implicitly = by excluding dot files). See for example, React, Flutter, Angular, CoreUI, Vue v2, Tensorflow, etc. Even Linux excludes .vscode implicitly.
3. Some repos such as Bootstrap, vuejs v3 and Angular actually include .vscode in their repo, so excluding .vscode globally may actually be problematic.

ntkme:
> In fact if you look at the most popular code repositories on Github, they exclude .vscode either explicitly or implicitly in .gitignore (implicitly = by excluding dot files). See for example, React, Flutter, Angular, CoreUI, Vue v2, Tensorflow, etc. Even Linux excludes .vscode implicitly.

Dot files are a different story. I'd say most projects exclude them due to security concerns: hidden malicious hooks, config files with secrets, .DS_Store or cache files leaking private info, etc.

Almost nobody add `.vscode` explicitly, because if you add vscode, someone else may ask for intellij, vim, emacs, etc. Choice of editor is a never ending culture war.

> Some repos such as Bootstrap, vuejs v3 and Angular actually include .vscode in their repo, so excluding .vscode globally may actually be problematic.

This is to enforce certain shared config preset by the development team (e.g. code indentation, etc). You can still add a global ignore and it would still help in the case that your own modified config won't get checked in accidentally. You still  do `git add -f` to add ignored files if you actually intend to propose an improvement to the shared config preset.

arikorn:
> Almost nobody add .vscode explicitly, because if you add vscode, someone else may ask for intellij, vim, emacs, etc. Choice of editor is a never ending culture war.

@ntkme I gather you didn't actually look at those repos. They explicitly exclude .vscode. These are the most popular code repos on GitHub, as ranked by GitHub (number of "stars" ranking in the hundreds of thousands each). I would not call these repos "nobodies", nor do they seem to share or experience your concerns about culture wars. Here's the link for the React .gitignore, to get you started: https://github.com/facebook/react/blob/88b40f6e41004773f19d37827980c66a8cf1865b/.gitignore#L23C1-L23C8 

> Dot files are a different story. I'd say most projects exclude them due to security concerns: hidden malicious hooks, config files with secrets, .DS_Store or cache files leaking private info, etc.

If you prefer excluding all dot files, I have no objection to changing this PR in that direction. FWIW, the evidence I presented does not support your claim the "most projects exclude them" -- in fact even this repo doesn't!

ntkme:
> They explicitly exclude .vscode.

That's their choices because those projects have too many noob users and they don't have bandwidth to deal with every single one of them who doesn't know how to configure global gitignore or .git/info/exclude.

> FWIW, the evidence I presented does not support your claim the "most projects exclude them" -- in fact even this repo doesn't!

English is not my first language, and what I meant by "I'd say most projects exclude them due to ..." is actually "I'd say projects exclude them mostly due to ...". - But I think the majority of native English speakers would understand my intention without problem.

Anyways, I have never said any of your idea is wrong, and I'm just giving my honest opinions. It's your freedom to have confirmation bias and misinterpret others in order to reinforce your narratives.

nex3:
I agree with @ntkme, ignores that are specific to the editors or other tools that aren't used by the project itself belong in personal gitignores, not project gitignores.

PR #2615, opened 2025-08-07
Print non-integer number at full precision in inspect mode

ntkme:
- Closes https://github.com/sass/sass/issues/3953

https://github.com/sass/sass-spec/pull/2067

### DIFF
--- a/pkg/sass-parser/package.json
+++ b/pkg/sass-parser/package.json
@@ -37,7 +37,7 @@
   "devDependencies": {
     "@types/jest": "^30.0.0",
     "copyfiles": "^2.4.1",
-    "expect": "^29.7.0",
+    "expect": "^30.0.5",
     "gts": "^6.0.2",
     "jest": "^30.0.4",
     "jest-extended": "^6.0.0",


ntkme:
I think this has absolutely no impact on normal css outputs, because it only changes how numbers are written in inspect mode. However, this currently breaks a few language tests where the result contains scss strings, which can only be serialized in inspect mode:

- The deprecation migration suggestions are using inspect mode output to produce suggested scss code, and now they are printed at higher than needed precision.
- There is one test case that use meta.inspect to output a scss string directly into css.

ntkme:
I did a [GitHub search](https://github.com/search?q=language%3ASCSS+%22meta.inspect%22&type=code) to find out where people are using or abusing `meta.inspect`. It turns out most of them are just abuses where it can be replaced with interpolation, list append, etc. In those cases this shouldn't really affect the behavior of the output.

It seems that the "legit" real world usage where this proposed change may have a small impact is: https://github.com/oddbird/true/blob/cfcd111acf0c168fd2f02cfa0c2a939dd199c189/sass/data/_details.scss#L127

@mirisuzanne @jgerigmeyer I think this change would actually make the "unpredictable" slightly better, but it would require change to the logic linked above. If you need to support both sass before and after this change, it can be implemented with a feature detection with `meta.inspect(1.00000000001) == meta.inspect(1)` to see if it supports high precision output for numbers in inspect mode or not.

@nex3 Any thoughts?

nex3:
I'd like to hear what Miriam and Jonny have to say, but in principle I'm supportive of this. `meta.inspect()` output is generally expected to be more flexible across versions than other API surfaces; we've even got a warning to that effect [in the documentation](https://sass-lang.com/documentation/modules/meta/#inspect).

jgerigmeyer:
Yeah, this logic makes sense to me. @mirisuzanne What do you think?

mirisuzanne:
We're not using this for core functionality, and the specific level of precision is incidental to our purpose. But at some point it was common for people to be surprised when e.g. `math.div(1,3)` was not equal to a copy/pasted output of that equation, something like `0.3333333333`. This code was just meant to warn that the Sass _output_ of two values is equal, even though Sass doesn't consider the two values equal internally. So if you're expecting equality here, you might need to round the values. 

But maybe string interpolation or some other method is better for comparing what Sass will actually output?

nex3:
Yeah, string interpolation or `"" + $value` will both return the Sass serialization, where `meta.inspect()` will generally skew towards providing more information than can necessarily be represented by CSS (for example, it will also include complex units).

PR #2619, opened 2025-08-17
Use official dart riscv64 docker image

ntkme:
Dart 3.9 now have official trixie based riscv64 docker image.


ntkme:
@nex3 Can you please take a look at this? I hope we can clean this up before next release.

nex3:
Thanks!

PR #2623, opened 2025-08-23
Make misplaced rest arguments a syntax error

nex3:
See https://github.com/sass/sass/issues/4120
See sass/sass-spec#2072
See sass/sass#4121
See https://github.com/sass/dart-sass/issues/2624


nex3:
Hmm... the fact that this was used apparently on purpose in Compass makes me more nervous about landing it without a deprecation. Maybe we should delay the breakage until Dart Sass 2.

Goodwine (approved):

PR #2629, opened 2025-08-28
Implement the new mixed declaration/comment/at-rule behavior

nex3:
See sass/sass#4123
See sass/sass-spec#2075

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## 1.92.0-dev
 
-* Emit declarations, childless at-rules, and comments in the order they appear
-  in the source even when they're interleaved with nested rules. This obsoletes
-  the `mixed-decls` deprecation.
+* **Breaking change:** Emit declarations, childless at-rules, and comments in
+  the order they appear in the source even when they're interleaved with nested
+  rules. This obsoletes the `mixed-decls` deprecation.
 
 * Fix a bug where `@extend` rules loaded through a mixture of `@import` and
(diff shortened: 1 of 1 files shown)


pkg/sass_api/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 15.10.0-dev
```
Goodwine, line 1: Shouldn't this CHANGELOG file update the entry below instead of adding a new one?

Goodwine (approved):

BogoBottom:
Does this pull request mean we can finally upgrade dart-sass all the way from 1.77.6 to 1.92.0 and later version effortlessly? So we can finally ignore the [mixed declaration deprecation](https://sass-lang.com/documentation/breaking-changes/mixed-decls/) safely?

nex3:
That depends what you mean by "safely". If you were confident that none of the CSS specificity changes would matter, you could always have [silenced the deprecation](https://sass-lang.com/documentation/breaking-changes/mixed-decls/#silencing-deprecations-in-dependencies) and continued on as before. If you think the reordering might cause problems, then if you upgrade to a version that reorders the declarations... well, it might cause problems. That's why we had the deprecation period in the first place: to indicate places where the semantics of the output were changing.

BogoBottom:
Yeah, I guess as long as there is no nested rules having same css properties declared inside and outside overwriting one after another, this change is fine.

PR #2639, opened 2025-09-03
Publish a `@sass/types` package with the types of the shared JS API

stof:
Closes https://github.com/sass/sass/issues/4037

The `@sass/types` package will be published with version numbers in sync with the `sass` and `sass-embedded` packages.

This `@sass/types` package is meant to be used by packages wanting to support _any_ implementation of the JS API when writing their type definition.
Both `sass` and `sass-embedded` still ship with type declarations included so that they can be supported natively by typescript (and IDEs).

### DIFF
--- a/.github/util/initialize/action.yml
+++ b/.github/util/initialize/action.yml
@@ -13,7 +13,7 @@ runs:
         sdk: "${{ inputs.dart-sdk }}"
         architecture: "${{ inputs.architecture }}"
 
-    - uses: actions/setup-node@v4
+    - uses: actions/setup-node@v5
       with:
         node-version: "${{ inputs.node-version }}"
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -141,7 +141,7 @@ jobs:
         with:
           token: ${{ secrets.GH_TOKEN }}
       # Set up .npmrc file to publish to npm
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v5
         with:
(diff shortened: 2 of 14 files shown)


pkg/sass-types/README.md
```diff
@@ -0,0 +1,6 @@
+# sass-types
+
+This package is a dual-publish of the [Sass JS API](https://sass-lang.com/documentation/js-api/)
+types. The `sass` and `sass-embedded` packages **still contain types**. This package is for
+users who _only_ need the Sass JS types (such as for `@types/gulp-sass`). It is published
```
stof, line 5: `@types/gulp-sass` does not use this _yet_, but I will use it once published to implement the type declarations for gulp-sass 6.x which uses the modern JS API (types for gulp-sass 5.x are implemented with a dependency on `@types/node-sass`, which was not supporting all the recent options even for the legacy API, but was at least a type-only package not forcing to install a different sass implementation than the one intended to be used)

pkg/sass-types/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "sass-types",
```
stof, line 2: I'll let you decide whether you want it as `sass-types`, `sass-js-api-types`, `@sass/types` or `@sass/js-api-types` (or another name).
For now, I went for a `sass-` prefix rather than the `@sass` scope because none of the other sass packages are scoped on npm.

stof:
@nex3 I updated the PR to use `@sass/types` as package name.

PR #2658, opened 2025-10-09
Fix bug with implicit configuration for nested imports

### LINKED ISSUE #2641: Variables from an imported, forwarded module are not passed through as configuration to another imported, forwarded module
Given code like:
```scss
<==> entrypoint.scss
@import 'def_wrapper';
@import 'ref_wrapper';

<==> _def_wrapper.scss
@forward 'def';

<==> _def.scss
$variable: blue;

<==> _ref_wrapper.scss
@forward 'ref';

<==> _ref.scss
$variable: red !default;

a {
  b: $variable;
}
```
You would expect the `$variable: blue` definition to be passed through the `_ref.scss` as [implicit configuration](https://github.com/sass/sass/blob/main/accepted/import-configuration.md), but that doesn't currently seem to be the case (the above code emits `a { b: red; }`).

jathak:
Actually fixes #2641.

See https://github.com/sass/sass-spec/pull/2089

### DIFF
--- a/lib/src/deprecation.dart
+++ b/lib/src/deprecation.dart
@@ -15,7 +15,7 @@ enum Deprecation {
   // DO NOT EDIT. This section was generated from the language repo.
   // See tool/grind/generate_deprecations.dart for details.
   //
-  // Checksum: 81dd840fe9e83020c1fd88a91e46f49b16190c76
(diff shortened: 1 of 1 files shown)


jathak:
Whoops. Looks like this one never got merged. Updated the versions and will do so once tests come back clean

nex3 (approved):

Issue #2647, opened 2025-09-12
Erroneously seeing the warning: "... deprecation is obsolete, so does not need to be made fatal."


u01jmg3:
Previously, @nex3 was kind enough to [provide the following code](https://github.com/sass/dart-sass/issues/633#issuecomment-2443664927) to exclude certain deprecations.

```js
const sass = require('sass');

const deprecationIds = Object.keys(sass.deprecations);
const excludedDeprecations = ['import'];
const includedDeprecations = deprecationIds.filter(id => !(new Set(excludedDeprecations)).has(id));

sass.compileString(
  '@import "other"', 
  { silenceDeprecations: excludedDeprecations, fatalDeprecations: includedDeprecations }
);
```

This worked without issue. However, I am now seeing the following unexpected warnings:

```
Warning: mixed-decls deprecation is obsolete, so does not need to be made fatal.

Warning: type-function deprecation is obsolete, so does not need to be made fatal.
```

Although it shouldn't be necessary, adding `mixed-decls` and `type-functions` to `silenceDeprecations` removes the previous warnings, but now causes the following:

```
Warning: mixed-decls deprecation is obsolete. If you were previously silencing it, your code may now behave in unexpected ways.

Warning: type-function deprecation is obsolete. If you were previously silencing it, your code may now behave in unexpected ways.
```

Ideally, I would expect deprecations that are obsolete not to be included by Sass.

nex3:
If obsolete deprecations weren't included at all, that would mean that any time we made a deprecation obsolete it would be a breaking change in the JS API because that field would vanish on `sass.deprecations`. We want to make sure it's possible to pass the same options across Sass versions (and potentially implementation) without lots of breakages, and that means keeping around old deprecations. Bear in mind that what you're doing here is already a bit off the beaten path.

You can look at [`Deprecation.status`](https://sass-lang.com/documentation/js-api/interfaces/deprecation/#status) to omit obsolete deprecations from `includedDeprecations`, which should fix your immediate issue.

PR #2671, opened 2025-11-05
--fatal-deprecation excludes obsolete Deprecations

halo:
Using `--fatal-deprecation 1.93.3` causes unrelated warnings:

```
Warning: mixed-decls deprecation is obsolete, so does not need to be made fatal.
Warning: type-function deprecation is obsolete, so does not need to be made fatal.
```

I think it is because `Depdecation.forVersion` does not filter out deprecations that have been marked as `obsoleteIn`.

This is my attempt to solve it. I hope I don't have to sign the Google Individual Contributor License Agreement, because I don't use a Google Account (I see no way to sign the agreement without an account, though I'm fine with agreeing to its contents, although my code is just a single conditional and the obvious related test). Also, the newly added test passes, and the grinder linted my code (as far as I can tell), but I could not run the -x node tests for some reason.

If this is unacceptable feel free to close this PR. It was a nice exercise learning something about Dart! Thank you for your time. 

PS: This project is so well maintained and I'm grateful to @nex3 who kept in in shape for years and put so much effort in documentation and figuring out a path to go forward with the SASS syntax as CSS evolves more and more. Thank you!

Related: https://github.com/sass/dart-sass/issues/2647

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 ## 1.94.2
 
-* Using `--fatal-deprecation 1.92.0` no longer emits warnings about
+* Using `--fatal-deprecation <version>` no longer emits warnings about
   deprecations that are obsolete.
+
+### Dart API
+
+* `Deprecation.forVersion` now excludes obsolete deprecations from the set it
+  returns.
 
 ## 1.94.1
 
--- a/pkg/sass-parser/package.json
+++ b/pkg/sass-parser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sass-parser",
-  "version": "0.4.34",
+  "version": "0.4.35",
   "description": "A PostCSS-compatible wrapper of the official Sass parser",
   "repository": "sass/dart-sass",
   "author": "Google Inc.",


jathak (changes requested):
Good catch with this! `--fatal-deprecation <version>` should definitely not make deprecations that were obsolete as of `<version>` fatal.

However, I actually think it makes more sense to exclude _all_ deprecations that are obsolete in the currently running version from `--fatal-deprecation`, even if they weren't yet obsolete in the `<version>` passed to the flag.

If for instance, `--fatal-deprecation 1.91.0` was added to a user's config when 1.91.0 was the current version, it makes sense for the flag to continue to work without warnings in 1.92.0 when `mixed-decls` and `type-function` are obsoleted without the user having to bump the fatal version to 1.92.0, which would additionally make `with-private` fatal.

lib/src/deprecation.dart
```diff
@@ -220,11 +220,23 @@ enum Deprecation {
       );
 
   /// Returns the set of all deprecations done in or before [version].
-  static Set<Deprecation> forVersion(Version version) {
-    var range = VersionRange(max: version, includeMax: true);
+  static Set<Deprecation> forVersion(Version version, [Version? current]) {
```
jathak: A comparison to the current version is not necessary here. This should just exclude any deprecation where `obsoleteIn` is non-null

jathak (changes requested):

lib/src/deprecation.dart
halo: xzMs8O1KQh17i

halo:
I admit I had to read your last paragraph 5 times before it finally sank in 😆 You are perfectly right.

The tests are somewhat lacking, because I don't know how stub the obsoletion data or sass version numbers.

(Edit: Sorry for all the `--force`. But I think I should have it now. Again 😄)

jathak (commented):
Can you bump the minor versions of the packages and add changelog entries so we can release this fix? (see [here](https://github.com/sass/dart-sass/blob/main/CONTRIBUTING.md#release-process) for instructions)

pkg/sass_api/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 16.0.2
+
+* No user-visible changes.
```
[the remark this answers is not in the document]
jathak: Yeah, `Deprecation.forVersion` is only exposed in the Dart API. I've added a changelog entry for this to the main Dart Sass changelog

PR #2691, opened 2025-12-01
Revert "Bump chokidar from 4.0.3 to 5.0.0 in /package"

jathak:
Reverts sass/dart-sass#2690


ntkme:
@nex3 Release 1.94.3 is broken on Node `<= 18`, which I think is still under the support policy.

nex3:
Oh, oops, didn't realize this was assigned to me!

PR #2709, opened 2025-12-19
Work around a null pointer exception in the Windows bots

nex3:


ntkme:
@nex3 I found a slightly different solution in #2710. I will let you decide which one to go with.

PR #2716, opened 2026-01-01
Emit deprecation warning for case-insensitive function names

kantnero:
Emits warning for case-insensitive function names:

- Deprecation entry: https://github.com/sass/sass/pull/4175
- Tests: https://github.com/sass/sass-spec/pull/2100


kantnero:
Emits warning for case-insensitive function names:

- Deprecation entry: https://github.com/sass/sass/pull/4175
- Tests: https://github.com/sass/sass-spec/pull/2100

PR #2717, opened 2026-01-05
Drop support for default exports

### LINKED ISSUE #2008: Add a default export for Node.js backwards compatibility

nex3:
Closes #2008

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,12 @@
   Leading combinators are still allowed for these functions because they may
   appear in a plain CSS nesting context.
 
+### JavaScript API
+
+* **Breaking change:** The JavaScript API no longer provides access to its
+  members through a default import in ESM. Users must `import * as sass from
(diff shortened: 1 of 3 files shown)


Goodwine (approved):
ah ok, this is for 2.0, I was confused for a second haha

PR #2721, opened 2026-01-07
Expose `Deprecation`s through `Logger.warn()`

nex3:


lib/src/embedded/logger.dart
pamelalozano16: Maybe `import '../deprecation.dart;'` here to make sure `Deprecation` is in scope in each translation unit that also imports `EmbeddedLogger`?

CHANGELOG.md
```diff
@@ -69,6 +69,10 @@
 
   [Dart class modifiers]: https://dart.dev/language/class-modifiers
 
+* **Breaking change:** The `deprecation` argument for `Logger.warn()` is now
+  `Deprecation?` rather than `bool`. This is `null` when the warning isn't for a
+  deprecation, and a `Deprecation` object when it is.
```
pamelalozano16: An example of before vs after (using the Deprecation object) would be useful here.

PR #2729, opened 2026-01-26
Use gamma 2.40 for display-referred rec2020

ntkme:
- https://github.com/w3c/csswg-drafts/issues/12574
- https://github.com/color-js/color.js/pull/672
- https://github.com/sass/sass-spec/pull/2107
- https://github.com/sass/embedded-host-node/pull/418

### DIFF
--- a/lib/src/value/color/space/rec2020.dart
+++ b/lib/src/value/color/space/rec2020.dart
@@ -11,12 +11,6 @@ import '../conversions.dart';
 import '../space.dart';
 import 'utils.dart';
 
-/// A constant used in the rec2020 gamma encoding/decoding functions.
-const _alpha = 1.09929682680944;
-
-/// A constant used in the rec2020 gamma encoding/decoding functions.
-const _beta = 0.018053968510807;
-
 /// The rec2020 color space.
 ///
 /// https://www.w3.org/TR/css-color-4/#predefined-rec2020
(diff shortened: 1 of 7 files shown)


lib/src/value/color/space/rec2020.dart
```diff
@@ -32,21 +26,19 @@ final class Rec2020ColorSpace extends ColorSpace {
   @override
   @protected
   double toLinear(double channel) {
-    // Algorithm from https://www.w3.org/TR/css-color-4/#color-conversion-code
+    // Non-linear transfer function from Rec. ITU-R BT.2020-2 table 4
+    //  Reference electro-optical transfer function from Rec. ITU-R BT.1886 Annex 1
+    //  with b (black lift) = 0 and a (user gain) = 1
+    //  defined over the extended range, not clamped
```
nex3: Let's keep the old reference to the CSS color conversion code. It's easier to directly verify that we match that sample code than to verify that it's doing the right thing per to rec2020 spec.

nex3 (approved):
Please sync these to the latest mains, and I'll bump the pubspec and add a changelog entry.

lib/src/value/color/space/rec2020.dart
ntkme: It was removed at the time when that page was outdated. Now I see it's updated with the new algorithm for new curve, I have reverted the comment back.

PR #2735, opened 2026-02-09
Fix simplifying SassCalculation with unsimplifiable CalculationOperation

ntkme:
This PR fixes the following error when a SassCalculation with unsimplifiable CalculationOperation is returned from custom function:

```
sass.Exception [Error]: TypeError: Instance of 'CalculationOperation0': type 'CalculationOperation0' is not a subtype of type 'Value1'
```

Unsimplifiable CalculationOperation means the lhs and rhs of operation has units that cannot be merged during compile time, for example: `calc(42px + 42%)`.

https://github.com/sass/sass-spec/pull/2111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 1.97.4-dev
+
+### JavaScript API
+
+* Fix a crash when manually constructing a `SassCalculation` for `'calc'` with
+  an argument that can't be simplified.
+
 ## 1.97.3
 
 * Fix a bug where nesting an at-rule within multiple style rules in plain CSS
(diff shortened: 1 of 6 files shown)


lib/src/js/compile.dart
```diff
@@ -388,7 +388,8 @@ Value _simplifyValue(Value value) => switch (value) {
               .map(_simplifyCalcArg)
               .toList(),
         )) {
-          ('calc', [var first]) => first as Value,
+          ('calc', [var first]) =>
+            first is Value ? first : SassCalculation.calc(first),
```
nex3: Can you make this a separate pattern match? As in `('calc', [Value first])`.

nex3:
Thanks for fixing this!

PR #2758, opened 2026-03-30
Support the top-level parent selector

nex3:
See https://github.com/sass/sass/pull/4206
See https://github.com/sass/sass-spec/pull/2129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,10 @@
-## 1.100.0
+## 1.99.0
 
 * Add support for parent selectors (`&`) at the root of the document. These are
   emitted as-is in the CSS output, where they're interpreted as [the scoping
   root].
 
   [the scoping root]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector#using_outside_nested_rule
-
-## 1.99.0
 
 * User-defined functions named `calc` or `clamp` are no longer forbidden. If
   such a function exists without a namespace in the current module, it will be
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 1.100.0
+
+* Add support for parent selectors (`&`) at the root of the document. These are
+  emitted as-is in the CSS output, where they're interpreted as [the scoping
+  root].
+
+  [the scoping root]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector#using_outside_nested_rule
+
 ## 1.99.0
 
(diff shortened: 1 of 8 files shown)


lib/src/ast/selector/list.dart
```diff
@@ -120,12 +120,18 @@ final class SelectorList extends Selector {
...
-        'Top-level selectors may not contain the parent selector "&".',
-        parentSelector.span,
-      );
+      if (accept(const _ParentSelectorVisitor())
+          case ParentSelector(
+            suffix: var _?,
+            :var span,
+          )) {
+        throw SassException(
+          'A top-level selector may not contain a parent selector with a '
```
Goodwine, line 129: Since we had `"&"` before, I'm curious why it was removed. Wouldn't it be easier for a user to understand `&` is the parent selector? Or is it implied that users would know that?

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: sass
-version: 1.99.0
+version: 1.100.0
```
Goodwine: 🥳

Goodwine (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.100.0
+
+* Add support for parent selectors (`&`) at the root of the document. These are
+  emitted as-is in the CSS output, where they're interpreted as [the scoping
+  root].
+
+  [the scoping root]: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Nesting_selector#using_outside_nested_rule
```
Goodwine, line 7: How stable are MDN links? I'm wondering if it's better to link the spec

Goodwine (approved):

pubspec.yaml
ntkme: 1.99 has not been released yet

lib/src/ast/selector/list.dart
nex3: All Sass exceptions include source span information, and we're careful to ensure that's included in the rendering of the exception everywhere. So this will include a little graphic pointing to the `&` selector, which means repeating it in prose is less useful than describing it.

pubspec.yaml
nex3: Oh, that's probably my mistake. I suppose we might as well roll this into that, then.

CHANGELOG.md
nex3: MDN links are pretty stable, and crucially they're *much* more friendly to people who are just looking to understand CSS. The spec is written for an audience of browser implementers, so it tends to be much denser.

PR #2769, opened 2026-05-07
Bump node engine requirement to >=20.19.0 and chokidar requirement to ^5.0.0 

### LINKED ISSUE #2768: Allow chokidar 5.0.0
In the chokidar releases 4.0.2 and 4.0.3 (newest `^4.0.0 version) are [missing the provenance](https://github.com/paulmillr/chokidar/issues/1440#issuecomment-3610663676). This leads to pnpm failing to install if `trustPolicy: "no-downgrade"` is configured.

The [release 5.0.0](https://github.com/paulmillr/chokidar/releases/tag/5.0.0) contains the provenance again. The only major change I can see is that the minimum node.js version was increased.

Could it be an option to extend the range for `chokidar` to allow 4.* and 5.* versions?

ntkme:
- Closes #2768

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -17,7 +17,7 @@
     "node": ">=20.19.0"
   },
   "dependencies": {
-    "chokidar": "^4.0.0",
+    "chokidar": "^5.0.0",
     "immutable": "^5.1.5",
     "source-map-js": ">=0.6.2 <2.0.0"
   },
--- a/package/package.json
+++ b/package/package.json
@@ -14,7 +14,7 @@
     "url": "https://github.com/nex3"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=20.19.0"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package/package.json
```diff
@@ -14,10 +14,10 @@
     "url": "https://github.com/nex3"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=20.19.0"
```
Goodwine, line 17: If we're picking the latest LTS, should this be `20.20.2`?

Goodwine (commented):
I think this looks good, I'll just confirm if it's okay to remove `lts-3`

package/package.json
ntkme: 20.19.0 is the minimum version supporting `require(esm)` - this used to be a 22.x feature, but it got back ported to 20.19.0.

Goodwine (approved):

PR #2772, opened 2026-05-23
Load import-only files through package.json `exports`

nex3:
See sass/sass#4224
See https://github.com/sass/sass-spec/pull/2139


Goodwine (approved):

ntkme:
@nex3 Looks like 1.101.0 release failed with a flaky test.

nex3:
Thanks! I've started it re-running.

PR #2791, opened 2026-06-30
Fix `--watch` not recompiling files modified via atomic save

matthewbjones:
When a watched stylesheet is edited via an atomic save (writing to a temporary file and then renaming it over the original, often called a "safe write"), `sass --watch` does not recompile it. `sed -i` does this, for example. This shows up with the npm/JavaScript sass package when watching without `--poll`, which uses `@parcel/watcher`.

I was able to consistently reproduce the bug both on my macOS machine as well as via a Docker Linux container.

I looked around as best as I could at Issues and Pull Requests and wasn't able to find this exact problem mentioned, however I feel like I was able to gain a sense of a rough timeline:

* The `add` vs `modify` file event handling was added to `--watch` in `v1.6.0` back in 2018, but the bug was latent because `chokidar` reported atomic saves of existing files as a coalesced change/modify event, so the `add` branch was not taken.
* Then `@parcel/watcher` was swapped in `v1.79.5` in October 2024. Unlike `chokidar`, it does not coalesce the rename into a modify; it reports a `create`. This is documented in parcel-bundler/watcher#98, which includes raw event logs comparing the two.
* PR #2444 reworked `--watch` and added mtime safety nets (`reloadAllModified()`) for some cases, but it preserves the `add` vs `modify` file event handling and does not cover parcel's `add`-for-existing-files event.

The existing `_handleModify` already had a branch to handle new files and calls `_handleAdd`, so I believe the minimal change to address this issue is to hand both `MODIFY` and `ADD` filesystem events to `_handleModify` (renamed to `_handleModifyOrAdd`) so it handles it first, where it can tell whether it's an existing tracked file or a newly discovered file.

I added regression tests for both an entrypoint and a dependency modified via an atomic save. On the Node (`@parcel/watcher`) backend where the bug manifests they fail without this change and pass with it.

A note on the rename: previously `_handleAdd` was called in two places: (1) directly in the filesystem event handling, and (2) in an `else` branch of the `MODIFY` event handling. After this change it's only called from `_handleModifyOrAdd`, so there is some argument that `_handleAdd` should be deleted and inlined into the `else` branch. I opted not to do that in this PR, to convey how small this change is.

### DIFF
--- a/lib/src/executable/watch.dart
+++ b/lib/src/executable/watch.dart
@@ -105,7 +105,6 @@ final class _Watcher {
         }
 
         switch (event.type) {
-          // A file may be updated via ADD, when using an atomic-write approach (temp file + rename)
           case ChangeType.MODIFY || ChangeType.ADD:
             _handleModifyOrAdd(event.path);
 
@@ -126,7 +125,12 @@ final class _Watcher {
     }
   }
 
-  /// Handles a modify/add event for the stylesheet at [path].
+  /// Handles a modifyor add event for the stylesheet at [path].
+  ///
+  /// @parcel/watcher reports atomic modifications (where a tool writes a temp
+  /// file then renames it over the existing file) as adds rather than
+  /// modifications, so we treat both event types as identical and disambiguate
+  /// based on our model of the filesystem rather than the reported event.
   ///
   /// Returns whether all necessary recompilations succeeded.
   void _handleModifyOrAdd(String path) {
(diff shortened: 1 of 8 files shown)


matthewbjones:
CLA just submitted

nex3 (approved):
Thanks for the fix!

lib/src/executable/watch.dart
```diff
@@ -128,10 +125,15 @@ final class _Watcher {
     }
   }
 
-  /// Handles a modify event for the stylesheet at [path].
+  /// Handles a modifyor add event for the stylesheet at [path].
```
matthewbjones: @nex3 just wanted to call to your attention a typo of `modifyor` here before this gets merged

PR #2798, opened 2026-07-10
Fix some GitHub actions issues introduced by the Zizmor refactor

nex3:


.github/workflows/build-linux.yml
```diff
@@ -97,6 +98,7 @@ jobs:
                      --platform "$PLATFORM" \
                      --volume "$PWD:$PWD" \
                      --workdir "$PWD" \
+                     --env "TARGET=$TARGET" \
```
ntkme, line 102:
```suggestion
                     --env "LIB=$LIB" \
                     --env "TARGET=$TARGET" \
```

jathak (approved):

Issue #2805, opened 2026-07-21
No release assets for 1.101.3


cocorossello:
In the latest release 1.101.3  there are no release assets. Is this an intentional change or a issue?

Thanks

nex3:
Definitely an issue! We did a big refactor of the actions recently with the introduction of Zizmor, and unfortunately the only way to debug them fully is to run a bunch of releases.

nackjicholson:
@nex3 Hello, will the 1.101.3 release be backfilled with assets or do you need to do a new release?

nex3:
I'll need to do a new release. I just kicked one off for 1.101.4.

ntkme:
1.101.4 failed again.

@nex3 Not sure about other issues in other child jobs, the problem with artifact publishing is that the repo variable should be:

``` diff
-       run: gh release upload "$GITHUB_REF_NAME" build-*/* --repo "$GITHUB_ACTION_REPOSITORY"
+       run: gh release upload "$GITHUB_REF_NAME" build-*/* --repo "$GITHUB_REPOSITORY"
```


GITHUB_ACTION_REPOSITORY:
> For a step executing an action, this is the owner and repository name of the action. For example, actions/checkout.


GITHUB_REPOSITORY:
> The owner and repository name. For example, octocat/Hello-World.

Because we're not running any "action" in this "step," `GITHUB_ACTION_REPOSITORY` is empty, so that `gh` tries to fallback to local git repo to detect the github repo address, which failed.

PR #2809, opened 2026-07-22
Fix more issues downstream from Zizmor

nex3:


jathak (approved):

.github/workflows/release.yml
```diff
@@ -69,6 +69,10 @@ jobs:
       contents: write # Uploads release
 
     steps:
+      # "gh release" requires itself to be run in a Git repository even when
+      # another repository is specified via `--repo`.
+      - run: git init
```
ntkme: This is not the correct fix, see my comment here: https://github.com/sass/dart-sass/issues/2805#issuecomment-5052226843

nex3: Ah, good catch.

PR #2816, opened 2026-07-27
Bump PostCSS in sass-parser

nex3:
This changes a test to align with PostCSS's new semantics (adding one container to another moves nodes rather than copying them).

Closes #2814
Closes #2804


Goodwine (approved):
yay

PR #2835, opened 2026-08-14
Upgrade to Dart 3.13.0

nex3:
This includes reformatting with the new Dart style logic and using a number of new language features.


lib/src/ast/css/modifiable/at_rule.dart
```diff
@@ -10,22 +10,14 @@ import '../value.dart';
...
-
-  @override
-  final FileSpan span;
-
-  new(this.name, this.span, {bool childless = false, this.value})
-    : isChildless = childless;
+final class ModifiableCssAtRule(
+  @override final CssValue<String> name,
+  @override final FileSpan span, {
+  bool childless = false,
```
jathak, line 16: Is there a specific reason we use `childless` for the parameter but `isChildless` for the field here?

lib/src/ast/css/modifiable/declaration.dart
```diff
@@ -11,34 +11,21 @@ import '../value.dart';
...
 
   /// Returns a new CSS declaration with the given properties.
-  new(
-    this.name,
-    this.value,
-    this.span, {
-    required this.parsedAsSassScript,
-    FileSpan? valueSpanForMap,
-  }) : valueSpanForMap = valueSpanForMap ?? value.span {
+  this {
```
jathak, line 28: I think it might make sense to move constructor bodies to the top of the class when present to avoid the parameter list being separated from the body.

lib/src/ast/css/modifiable/import.dart
```diff
@@ -10,21 +10,11 @@ import '../value.dart';
...
-  @override
-  final CssValue<String>? modifiers;
-
-  @override
-  final FileSpan span;
-
-  new(this.url, this.span, {this.modifiers});
-
+final class ModifiableCssImport(
+  @override final CssValue<String> url,
```
jathak, line 14: The doc comment for this was removed

lib/src/ast/sass/expression/legacy_if.dart
```diff
@@ -15,20 +15,17 @@ import '../../../visitor/interface/expression.dart';
 /// evaluated.
 ///
 /// {@category AST}
-final class LegacyIfExpression extends Expression
-    implements CallableInvocation {
+final class LegacyIfExpression(
+  /// The arguments passed to `if()`.
+  @override final ArgumentList arguments,
+
+  @override final FileSpan span,
```
jathak, line 21: For primary constructor parameters without doc comments, sometimes there's a blank line before them and sometimes there's not. I'd lean towards no blank line in these cases, but we should be consistent one way or the other.

lib/src/ast/sass/expression.dart
```diff
@@ -20,12 +20,10 @@ import '../sass.dart';
 /// {@category AST}
 /// {@category Parsing}
 @sealed
-abstract class Expression implements SassNode {
+abstract class Expression() implements SassNode {
```
jathak: I believe you'd said you would leave off the primary constructors for classes with implicit zero-argument constructors. Is that not the case here?

lib/src/ast/node.dart
```diff
@@ -14,7 +14,7 @@ import 'package:source_span/source_span.dart';
 ///
 /// {@category AST}
 @sealed
-abstract interface class AstNode {
+abstract interface class AstNode() {
```
jathak: Same here

lib/src/parse/selector.dart
jathak, line 54: Not sure if it matters in this case since these docs aren't rendered for `sass` or `sass_api`, but when using private named parameters, should the reference be to the version with the underscore (as it appears here) or to the name without an underscore (as they would appear when calling this)?

jathak (commented):
Looking through this, converting everything to primary constructors where possible seems like the right move even if it's sometimes a bit awkward.

lib/src/ast/css/modifiable/at_rule.dart
nex3: The style guide says that [boolean properties should be verb phrases](https://dart.dev/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable) (like "is childless"), but suggests that [as parameters the verb should be omitted](https://dart.dev/effective-dart/design#consider-omitting-the-verb-for-a-named-boolean-parameter).

lib/src/ast/css/modifiable/declaration.dart
nex3: I understand the impulse, but what used to be "a constructor" is now intrinsically split across multiple locations—the parameters are in the class header, additional field initializers are at the field locations, and the constructor body is at the `this` call. I think it makes sense to have the constructor body after field initializers, since it's run after they're initialized.

Given that constructor bodies should come after fields, I think it makes sense to put them after properties as well, for the same reason we've historically put constructors after properties: the properties (like fields) provide a view into the "state" of the object from a user's perspective, even if they're computable from other state internally.

lib/src/ast/css/modifiable/import.dart
nex3: That was intentional, since it's an `@override` where the parent class's doc comment still applies.

lib/src/ast/sass/expression/legacy_if.dart
nex3: I agree there should be no blank line. Let me know if you notice other instances of this.

lib/src/ast/sass/expression.dart
nex3: Nope, this is just a mistake 😄.

lib/src/parse/selector.dart
nex3: Dartdoc complains if it's not the version with the underscore, apparently.

lib/src/ast/sass/expression.dart
nex3: The failing tests reminded me why this was necessary: because this class has a factory constructor, it _doesn't_ have an implicit zero-argument constructor. One needs to be declared explicitly.

jathak (approved):

PR #2847, opened 2026-09-01
Prevent `--watch` from compiling generated CSS files to themselves

### LINKED ISSUE #2846: `--watch` causes infinite loop in many-to-many mode
In many-to-many mode, the CLI compiler does not ignore generated files on subsequent runs:
```bash
$ touch style.scss
$ sass .:dist
$ sass .:dist
$ tree dist
dist
├── dist
│   ├── style.css
│   └── style.css.map
├── style.css
└── style.css.map
```
This causes an infinite loop with `--watch` when the destination is a subdirectory of the source:
```bash
$ sass --watch .:dist
[2026-09-01 06:56:44] Compiled dist/dist/style.css to dist/dist/dist/style.css.
Sass is watching for changes. Press Ctrl-C to stop.

[2026-09-01 06:56:44] Compiled dist/dist/dist/style.css to dist/dist/dist/dist/style.css.
[2026-09-01 06:56:44] Compiled dist/dist/dist/dist/style.css to dist/dist/dist/dist/dist/style.css.
[2026-09-01 06:56:44] Compiled dist/dist/dist/dist/dist/style.css to dist/dist/dist/dist/dist/dist/style.css.
# ...
$ du -sh dist
1.7M    dist
```

Observed on v1.103.1 from the Arch `dart-sass` package.

stygian-ash:
Fix an edge case where
```bash
touch style.scss
sass -w .:out
```
causes in an infinite loop.

* Loop occurred because Sass tries to compile `style.scss` → `out/style.css` → `out/out/style.css` → …
* The extant fix for [#853] misses this since `destination == "out/out/style.css" != "out/style.css" == source`
* Fixed by ignoring all source files that are inside the destination directory, except when the source and destination directories are the same
* Added regression tests

Closes #2846

[#853]: <https://github.com/sass/dart-sass/issues/853>

### DIFF
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -464,8 +464,10 @@ final class ExecutableOptions._(
     return {
       for (var path in listDir(source, recursive: true))
         if (_isEntrypoint(path) &&
-            // Don't compile a CSS file that is already in destination.
-            !(p.extension(path) == '.css' && p.isWithin(destination, path)))
+            // Don't compile a CSS file to its own location.
+            !(source == destination && p.extension(path) == '.css') &&
+            // Ignore files already in the destination if it is inside the source.
+            !(p.isWithin(source, destination) && p.isWithin(destination, path)))
           path: p.join(
             destination,
             p.setExtension(p.relative(path, from: source), '.css'),
--- a/lib/src/executable/options.dart
+++ b/lib/src/executable/options.dart
@@ -464,8 +464,8 @@ final class ExecutableOptions._(
     return {
       for (var path in listDir(source, recursive: true))
         if (_isEntrypoint(path) &&
-            // Don't compile a CSS file to its own location.
-            !(source == destination && p.extension(path) == '.css'))
+            // Don't compile a CSS file that is already in destination.
+            !(p.extension(path) == '.css' && p.isWithin(destination, path)))
           path: p.join(
             destination,
             p.setExtension(p.relative(path, from: source), '.css'),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.104.1-dev
+
+### Command Line Interface
+
+* Many-to-many compilations no longer compile any files that appear in the
+  output directory if the output directory is also within the source directory.
+  This fixes a bug where `--watch` mode could enter an infinite loop recompiling
+  the same CSS file over and over.
+
 ## 1.104.0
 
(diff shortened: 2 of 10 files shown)


ntkme:
A bigger question is that whether we should just ignore everything inside destination dir as source files, if destination dir is inside source dir.

stygian-ash:
> A bigger question is that whether we should just ignore everything inside
> destination dir as source files, if destination dir is inside source dir.

I would lean towards _yes_, but I can't really think of a reasonable use case
where you would have Sass files in the output folder in the first place. Maybe
something like this?
```
.
├── out
│   ├── examples
│   │   └── basic-syntax.sass
│   └── index.html
└── style.scss
```
Here I would find the appearance of `out/examples/out/basic-syntax.css`
surprising and likely unwanted, and if I _did_ desire it, I could simply run
`sass .:out/ out/examples/:out/examples/out/` instead.

Conversely, in a situation like this
```
.
├── dist
│   ├── example-00.sass
│   ├── ...
│   ├── example-10.sass
│   └── index.html
└── style.scss
```
running `sass -w . dist/ dist/:dist/dist/` would result in the generated
`dist/style.css` being copied to `dist/dist`, which is probably bad.

I mean this is a pretty strange thing to do, and in either of these contrived
examples you could just slightly restructure your project, so I'm not sure how
much it matters what we choose.

nex3:
I agree that we should just ignore everything in the output directory.

nex3 (approved):

lib/src/executable/options.dart
```diff
@@ -465,7 +465,9 @@ final class ExecutableOptions._(
       for (var path in listDir(source, recursive: true))
         if (_isEntrypoint(path) &&
             // Don't compile a CSS file to its own location.
-            !(source == destination && p.extension(path) == '.css'))
+            !(source == destination && p.extension(path) == '.css') &&
```
ntkme, line 468: I see the other variant in `watch.dart` does not check for css extension, and I think technically we don't need to check. Shall we remove it to make the logic consistent?

nex3: I don't think it matters too much, but sure.

nex3: Actually, upon examination, I think this is the easiest way to express it. The crucial difference is that the `options.dart` variant only operates on the directories and the original path, while the `watch.dart` variant has access to the destination path. We could change that, but it would make the whole thing less elegant.

PR #2850, opened 2026-09-02
Fix `--watch` not recompiling sources modified during a compilation

### LINKED ISSUE #2849: --watch silently skips files modified while a compilation is in progress
When a source file is modified while `sass --watch` is in the middle of a compilation, that modification is never applied. The CSS on disk stays one generation behind, and nothing indicates it: no error, no warning, and no `Compiled` line for the modification that was dropped.

Reproduction. The only requirement is a stylesheet slow enough to compile that a write can land while it's running; this one takes about 3 seconds:

    mkdir sass-watch-repro && cd sass-watch-repro

    cat > _slow.scss <<'SCSS'
    @use "sass:math";
    @for $i from 1 through 60 {
      .a-#{$i} { %ph-#{$i} { color: red; } }
      @for $j from 1 through 60 {
        .b-#{$i}-#{$j} { @extend %ph-#{$i}; width: math.div($i * 100%, $j); }
      }
    }
    SCSS

    echo '.p {color: p0}' > _p.scss
    echo '.q {color: q0}' > _q.scss
    echo '@use "slow"; @use "p"; @use "q";' > test.scss

    sass --no-source-map --watch test.scss:out.css &
    sleep 6
    echo '.p {color: p1}' > _p.scss   # starts a ~3s compilation
    sleep 1
    echo '.q {color: q1}' > _q.scss   # lands while that compilation is running
    sleep 12

Expected: out.css contains both `p1` and `q1`.

Actual: only the first modification is applied, and the watcher logs a single compilation for the two writes.

    $ grep -E 'color: [pq][01]' out.css
      color: p1;
      color: q0;

    $ cat watch.log
    [2026-09-01 17:56:40] Compiled test.scss to out.css.
    Sass is watching for changes. Press Ctrl-C to stop.

    [2026-09-01 17:56:45] Compiled test.scss to out.css.

Restarting doesn't recover it. `--watch` implies `--update`, and the stale CSS was written after the sources, so a fresh run decides there is nothing to do and exits 0 without compiling:

    $ sass --no-source-map --update test.scss:out.css; echo "exit $?"
    exit 0
    $ grep -E 'color: q[01]' out.css
      color: q0;

Modifying any stylesheet afterwards does recover it, because the next event recompiles the entrypoint from whatever is on disk at that point.

`--poll` behaves identically, so this doesn't appear to be about the watch backend. The modification times point at the cause:

    $ stat -c '%y  %n' _p.scss _q.scss out.css
    2026-09-01 17:56:43.515256175 +0200  _p.scss
    2026-09-01 17:56:44.516290054 +0200  _q.scss
    2026-09-01 17:56:45.852335239 +0200  out.css

`_q.scss` was modified during the compilation that wrote `out.css`, so it is older than the output it isn't part of. The `ifModified` check in `compileStylesheet` compares the transitive source modification times from `StylesheetGraph.modifiedSince` against the destination's modification time, concludes the source is up to date, and skips the recompilation the event had queued. Setting `ifModified` to `false` in the watcher's compile call makes the reproduction pass, which suggests the event itself is delivered and handled correctly and only the recompilation is discarded.

Where this bites in practice: any tool that writes stylesheets in bursts over a window longer than one compilation. A `git rebase` or branch switch across several commits that touch stylesheets does exactly that - the first burst kicks off a compilation and everything written during it is lost, leaving a build that's silently stale until someone touches a file by hand.

Environment: Dart Sass 1.103.1 on Linux, reproduced both with and without `--poll`, and also on current `main`.

One more thing I noticed while narrowing this down, possibly a separate issue: modifications that land during the very first compilation, before the watcher starts iterating its event stream, appear to be dropped outright rather than skipped. The symptom on disk is identical.

corebonts:
Fixes #2849.

`--watch` decides whether a stylesheet needs recompiling by comparing the modification time of the source and its dependencies against that of the CSS file it produced (`compileStylesheet`'s `ifModified`). When a source is modified while a compilation is already in progress that comparison is wrong: the CSS file is written when the compilation finishes, so it is newer than the modification that arrived in the meantime. The source looks up to date, no recompilation happens, and the CSS on disk stays one generation behind, silently. It also survives a restart, since `--update` makes the same comparison.

This threads the time at which the previous compilation started through to `compileStylesheet` and uses it in place of the destination's modification time when it is earlier. A source that changed during a compilation is newer than that timestamp, so it is recompiled. The bound is always the earlier of the two, so this can only ever cause more compilations than before, never fewer, and only for sources whose modification time falls between the start of the last compilation and the output it wrote. Startup, one-shot compilations and `--update` are unaffected, since they pass no timestamp.

The regression test is in the existing `recompiles a watched file` group. It needs a stylesheet that takes several seconds to compile, since the bug only exists while a compilation is in flight; the delay between the two writes is longer than the polling interval so that the two modifications aren't batched into a single recompilation under `--poll`. It fails on `main` in both the `--poll` and non-`--poll` variants and passes with this change; the rest of `test/cli/dart/watch_test.dart` and `update_test.dart` (114 tests) pass with no retries.

One related problem this doesn't address: modifications that land during the initial compilation are dropped outright rather than skipped, because `MultiDirWatcher`'s `StreamGroup` has no subscriber until `watch()` starts iterating it, which happens after that compilation finishes.
I wanted to provide a minimal fix to the more common issue we face, and fixing that means attaching the subscription earlier, which changes startup behaviour, so it seemed better as a separate change. But if you want I can include that one as well.

### DIFF
--- a/test/cli/shared/watch.dart
+++ b/test/cli/shared/watch.dart
@@ -208,6 +208,41 @@ void sharedTests(
               .validate();
         });
 
+        // Regression test for #2849.
+        test("when a dependency is modified during a compilation", () async {
+          await d.file("_slow.scss", _slowSource).create();
+          await d.file("_other.scss", "a {b: c}").create();
+          await d.file("test.scss", "@use 'slow'; @use 'other'").create();
+
+          var sass = await watch(["test.scss:out.css"]);
+          await expectLater(
+            sass.stdout,
+            emits(endsWith('Compiled test.scss to out.css.')),
+          );
+          await expectLater(sass.stdout, _watchingForChanges);
+          await tickIfPoll();
+
+          // Modifying the slow stylesheet starts a compilation that runs for
+          // several seconds. The delay is longer than the polling interval so
+          // that the modification below is seen as a separate change, but short
+          // enough that it lands while that compilation is still in progress.
+          await d.file("_slow.scss", "$_slowSource\n// modified").create();
+          await Future<void>.delayed(Duration(seconds: 3));
+          await d.file("_other.scss", "x {y: z}").create();
+
+          await expectLater(
+            sass.stdout,
+            emits(endsWith('Compiled test.scss to out.css.')),
+          );
+          await expectLater(
+            sass.stdout,
+            emits(endsWith('Compiled test.scss to out.css.')),
+          );
+          await sass.kill();
+
+          await d.file("out.css", contains("y: z")).validate();
+        }, timeout: Timeout.factor(4));
+
         test("when it's modified when watched from a directory", () async {
           await d.dir("dir", [d.file("test.scss", "a {b: c}")]).create();
 
@@ -1215,6 +1250,18 @@ void sharedTests(
     });
   }
 }
+
+/// A stylesheet that takes multiple seconds to compile, so that another file
+/// can be modified while a compilation that includes it is in progress.
+const _slowSource = r"""
+@use "sass:math";
+@for $i from 1 through 80 {
+  .a-#{$i} { %ph-#{$i} { color: red; } }
+  @for $j from 1 through 80 {
+    .b-#{$i}-#{$j} { @extend %ph-#{$i}; width: math.div($i * 100%, $j); }
+  }
+}
+""";
 
 /// Writes [contents] to [path] atomically.
 ///
--- a/test/cli/shared/watch.dart
+++ b/test/cli/shared/watch.dart
@@ -1254,13 +1254,19 @@ void sharedTests(
 /// A stylesheet that takes multiple seconds to compile, so that another file
 /// can be modified while a compilation that includes it is in progress.
 const _slowSource = r"""
-@use "sass:math";
-@for $i from 1 through 80 {
-  .a-#{$i} { %ph-#{$i} { color: red; } }
-  @for $j from 1 through 80 {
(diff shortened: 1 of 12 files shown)


test/cli/shared/watch.dart
```diff
@@ -208,6 +208,41 @@ void sharedTests(
               .validate();
         });
 
+        // Regression test for #2849.
+        test("when a dependency is modified during a compilation", () async {
```
nex3, line 212: I think it's worth having a parallel test for `--update` as well.

nex3 (commented):
Thanks for the submission!

Consider instead setting the modification time of the generated file to the time the compilation began. That way, checking the mtime of the source versus the destination will work even across separate runs of the executable, such as when using `--update`.

test/cli/shared/watch.dart
corebonts: I also thought about modifying the target's date, but I thought that this way it is a smaller change. Implementing the suggestion and adding the additional test.

nex3:
https://github.com/google/dart_cli_pkg/pull/251 and https://github.com/sass/dart-sass/pull/2859 should fix the CI errors here.

nex3 (approved):

nex3:
Thanks for the contribution!

PR #2861, opened 2026-09-14
[meta.load] Add support for `meta.load()` and related features

nex3:
See https://github.com/sass/sass/issues/739
See https://github.com/sass/sass/pull/4247
See https://github.com/sass/sass-spec/pull/2170

[skip sass-embedded]


lib/src/value/module.dart
```diff
@@ -0,0 +1,101 @@
...
+  @override
+  @internal
+  T accept<T>(ValueVisitor<T> visitor) => visitor.visitModule(this);
+
+  @override
+  SassModule assertModule([String? name]) => this;
+
+  /// Asserts that this SassModule belongs to [compileContext] and returns it.
+  ///
+  /// It's checked before interacting with a `SassModule` to prevent sharingx of
```
jathak:
```suggestion
  /// It's checked before interacting with a `SassModule` to prevent sharing of
```

jathak (approved):

